PR #56, opened 2022-04-30
[WIP] Adding the requests module

bartbroere:
Locally, I have been doing some work with getting the Python requests module to work.

This is still a work-in-progress because I couldn't find a way to include Python modules easily yet. For now, I just point it to my own Pyodide build. Any ideas on how to include our custom Python modules?

image

### DIFF
--- a/src/core/computations/python/loadPython.ts
+++ b/src/core/computations/python/loadPython.ts
@@ -13,7 +13,7 @@ declare global {
 
 export async function loadPython() {
   window.pyodide = await window.loadPyodide({
-    indexURL: "https://python.rijksoverhemd.nl",
+    indexURL: "https://cdn.jsdelivr.net/pyodide/v0.19.0/full/",
     stdout: (l: string) => {
       console.log("[WASMPython]", l);
     },
--- a/src/core/computations/python/loadPython.ts
+++ b/src/core/computations/python/loadPython.ts
@@ -13,17 +13,17 @@ declare global {
 
 export async function loadPython() {
   window.pyodide = await window.loadPyodide({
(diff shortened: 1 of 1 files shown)


davidkircos:
Bart - I can think of two ways:

1) Fork Pyodide on quadratichq/pyodide, adding the Requests shim to the packages in this fork. And then I can host this fork through AWS and we can point at it.

2) Copy a Pyodide release into quadratic/public/**pyodide**  and we can add additional packages there. This way it will be hosted directly alongside the app itself, even when run locally.

I think number 2 is a better solution, but may be more difficult to update and maintain over time.

davidkircos:
@bartbroere I am thinking a combination of both approaches is the correct method. 

I have created a fork of pyodide/pyodide at https://github.com/quadratichq/pyodide

I'd like to make a pull request to merge your shim into this fork. And then I will do a build of Pyodide and ship the result in quadratic in `public/pyodide` source folder.

Could you create a PR of your shim to quadratichq/pyodide ?

bartbroere:
Sounds good! I have created a pull request on your pyodide fork. Have you considered adding the pyodide fork as a git submodule of quadratic?

src/core/computations/python/loadPython.ts
```diff
@@ -13,17 +13,17 @@ declare global {
 
 export async function loadPython() {
   window.pyodide = await window.loadPyodide({
-    indexURL: "https://cdn.jsdelivr.net/pyodide/v0.19.0/full/",
+    indexURL: "https://python.rijksoverhemd.nl",
```
bartbroere, line 16: Revert this line to how we found it

```suggestion
    indexURL: "https://cdn.jsdelivr.net/pyodide/v0.19.0/full/",
```

PR #59, opened 2022-05-10
Support Python Requests 

davidkircos:
This branch self hosts a version of pyodide `Pyodide NFI version 0.17.0.2021.11.24` which contains a modified version of python requests to run in the browser.


davidkircos:
Hey @bartbroere I could not get Pyodide to build from `quadratichq/pyodide:main` so I downloaded your build of Pyodide from release `NFI version 0.17.0.2021.11.24` and put it into its branch of Quadratic.

I am realizing that making any sort of API request without the proper CORS headers is going to fail. This will make it difficult for users to make any API requests directly out of Quadratic. 

I am considering what it would take to make a proxied version of requests, where all requests from inside pyodide in quadratic are piped through a proxy server. This way it would not be limited to the features available in the browser and could make requests directly to external APIs. I will have to make a proxy anyhow for database connections, so it seems like that will be the most logical long term direction.

bartbroere:
> I downloaded your build of Pyodide from release `NFI version 0.17.0.2021.11.24` and put it into this branch of Quadratic.

Cool! The newer (pre-)releases are perhaps more suitable, I just didn't mark them as the latest stable release yet, but we have deployed them without any real issues. I always use the Docker image `pyodide/pyodide-env` combined with the Docker tag that I find in the `.circleci` configuration. That seems to do the trick when I try to build Pyodide with extra packages myself.

For the most recent release these were my build steps.

```bash
docker run -v /home/bart/Repositories/pyodide/:/src/pyodide -it pyodide/pyodide-env:20220504-py310-chrome101-firefox100 bash
cd pyodide && source pyodide_env.sh
PYODIDE_PACKAGES="*" make
```

> I am considering what it would take to make a proxied version of requests, where all requests from inside pyodide in quadratic are piped through a proxy server. This way it would not be limited to the features available in the browser and could make requests directly to external APIs. I will have to make a proxy anyhow for database connections, so it seems like that will be the most logical long term direction.

I have read the [MDN page on CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and came to the conclusion that without such a proxy on the server side, you simply can't get all HTTP requests to work unfortunately. There are projects like [CORS anywhere](https://github.com/Rob--W/cors-anywhere/) that already do a bit of man-in-the-middle CORS header manipulation. Something like this on the server side might be necessary indeed.

But since you are already planning on making an SQL proxy, that sounds like a logical place to put the HTTP proxy indeed.

I'm still following this project with a lot of interest!

davidkircos:
Marking this as closed. For now, Requests are supported via `js.fetch` 

https://docs.quadratic.to/reference/python-cell-reference/making-requests

PR #108, opened 2022-08-26
Features/axis lines

davidfig:
Since I was here already, this has the same logic for axes lines as the grid lines. I'm not sure how strongly you feel about react PIXI verse vanilla PIXI. This ones reuses the drawGridLines instead of updating the component. I can do either way but this is my preference. Let me know if you prefer it as a component and I'll change the grid lines to a component as well.


davidkircos:
I don't have a strong preference between ReactPIXI and Vanilla. 

I've used React Pixi in places where it has helped in passing the application state.

davidkircos:
Generally if Vanilla PIXI is clean to use, we probably should because it will be probably be faster.

davidkircos:
@davidfig I noticed 2 regressions from this PR

1) When zoomed out, axis lines should not disappear with the grid lines.
2) View > Show Axis no longer toggles the visibility of the Axis lines

Screen Shot 2022-08-26 at 10 44 54 AM

PR #113, opened 2022-08-27
Refactor pixi

davidfig:


src/core/gridGL/GridHeaders.tsx
```diff
@@ -0,0 +1,232 @@
...
+        }
+      }
+    };
+
+    const drawCorner = () => {
+      corner.clear();
+      corner.beginFill(colors.headerCornerBackgroundColor);
+      corner.drawRect(bounds.left, bounds.top, rowWidth, cellHeight);
+      corner.endFill();
+    };
```
davidkircos, line 213: In Excel when you click the corner it selects everything. We might want to trigger "Zoom to Fit" as well as Select All when a user clicks the corner. 

I created a future issue for this. https://github.com/quadratichq/quadratic/issues/116

src/core/gridGL/graphics/ViewportComponent.tsx
```diff
@@ -48,74 +48,74 @@ const PixiComponentViewport = PixiComponent('Viewport', {
...
-            );
-            cull.cull(visibleBoundsExtended);
-
-            UpdateGridAlphaOnZoom(globals);
+    let dirty = true;
+
+    viewport.on('zoomed', () => dirty = true);
+    viewport.on('moved', () => dirty = true);
+
+    const graphics = viewport.addChild(new PIXI.Graphics());
```
davidkircos, line 56: variable name isn't descriptive

src/core/gridGL/graphics/hooks/useTicker.ts
```diff
@@ -0,0 +1,16 @@
+import * as PIXI from 'pixi.js';
```
davidkircos, line 1: unused?

davidkircos:
- ZoomMenu doesn't flag dirty properly
- Offest show and hide headings
- Axes line dims when you zoom out
- Code documentation, clean up variable names

src/core/gridGL/QuadraticGrid.tsx
```diff
@@ -49,8 +88,43 @@ export default function QuadraticGrid() {
     useMenuState();
   const [rightClickPoint, setRightClickPoint] = useState({ x: 0, y: 0 });
 
+  const pointerEvents = usePointerEvents({
```
davidkircos, line 91: Is this a hook? Or a function registering events?

davidkircos: Would it make sense to put this in the ViewportComponent.tsx instead of passing from QuadraticGrid?

src/core/gridGL/QuadraticGrid.tsx
```diff
@@ -1,27 +1,32 @@
...
 import { colors } from '../../theme/colors';
 import { useMenuState } from '@szhsin/react-menu';
 import RightClickMenu from '../../ui/menus/RightClickMenu';
-
 import { ViewportEventRegister } from './interaction/ViewportEventRegister';
+import useLocalStorage from '../../hooks/useLocalStorage';
+import { gridHeadingsProps } from './graphics/gridHeadings';
+import { axesLinesProps } from './graphics/axesLines';
+
+export interface Size {
```
davidkircos, line 26: Size of what? If this is a generic type move to a new file.

src/core/gridGL/QuadraticGrid.tsx
```diff
@@ -49,8 +88,43 @@ export default function QuadraticGrid() {
...
+    setEditorInteractionState,
+    viewportRef,
+    headerSize,
+  });
+
+  if (loading || !canvasSize) return null;
+
   return (
     <div
+      className="canvas"
```
davidkircos, line 120: `canvas-container`?

src/core/gridGL/QuadraticGrid.tsx
```diff
@@ -71,64 +145,46 @@ export default function QuadraticGrid() {
...
-            setEditorInteractionState
-          );
-        }}
-        style={{ display: loading ? 'none' : 'inline' }}
         // Disable rendering on each frame
         raf={false}
+
         // Render on each state change
-        renderOnComponentChange={true}
+        // renderOnComponentChange={true}
```
davidkircos, line 159: How does a cell updating trigger a render?

src/core/gridGL/graphics/ViewportComponent.tsx
```diff
@@ -48,74 +48,74 @@ const PixiComponentViewport = PixiComponent('Viewport', {
...
+    const headings = viewport.addChild(new PIXI.Container());
+    const headingsGraphics = headings.addChild(new PIXI.Graphics());
+    const labels = headings.addChild(new PIXI.Container());
+    const corner = headings.addChild(new PIXI.Graphics());
+
+    // Quadratic Render Loop, render when dirty.
+    // Remember when anything changes on the stage to either set viewport.dirty = true
+    // Or use a react component that is a child of viewport (React Pixi will trigger a render)
```
davidkircos, line 64: I don't think these comments are true anymore.

Why are there two dirty flags?

src/core/gridGL/graphics/ViewportComponent.tsx
```diff
@@ -48,74 +48,74 @@ const PixiComponentViewport = PixiComponent('Viewport', {
...
+    // FPS log
+    // ticker.add((time) => {
+    //   console.log(`Current Frame Rate: ${props.app.ticker.FPS}`);
+    // });
+
+    console.log('[QuadraticGL] environment ready');
     return viewport;
   },
+
+  applyProps(viewport: Viewport, oldProps: ViewportProps, newProps: ViewportProps) {
```
davidkircos, line 108: `applyProps: (viewport: Viewport, oldProps: ViewportProps, newProps: ViewportProps) => {`

src/core/gridGL/graphics/axesLines.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface IProps {
+  viewport: Viewport;
+  graphics: PIXI.Graphics;
+}
+
+export const axesLinesProps = { showGridAxes: true };
+
+export function axesLines(props: IProps): void {
+  if (axesLinesProps.showGridAxes) {
```
davidkircos, line 12: two sets of props?

src/core/gridGL/graphics/gridUtils.ts
```diff
@@ -0,0 +1,12 @@
+import { Viewport } from 'pixi-viewport';
+
+export function alphaGridLines(viewport?: Viewport): false | number {
```
davidkircos, line 3: Return 0 instead of false to simplify?

Rename `calculateAlphaForGridLines`

src/core/gridGL/interaction/usePointerEvents.ts
```diff
@@ -0,0 +1,180 @@
...
+
+interface IProps {
+  viewportRef: React.MutableRefObject<Viewport | undefined>;
+  interactionState: GridInteractionState;
+  setInteractionState: React.Dispatch<React.SetStateAction<GridInteractionState>>;
+  setEditorInteractionState: React.Dispatch<React.SetStateAction<EditorInteractionState>>;
+}
+
+const MINIMUM_MOVE_POSITION = 5;
+const DOUBLE_CLICK_TIME = 200;
```
davidkircos, line 17: I think the double click standard is 500ms

src/styles.css
```diff
@@ -9,21 +9,6 @@ html,
...
-}
-
-canvas {
-  position: fixed;
-  top: 0;
-  left: 0;
-  width: 100%;
-  height: 100%;
-
-  outline: none;
```
davidkircos, line 27: I think we can get rid of the class if it is only passing one style

src/styles.css
```diff
@@ -9,21 +9,6 @@ html,
...
-  -webkit-tap-highlight-color: transparent; /* mobile webkit */
+.canvas {
+  -webkit-tap-highlight-color: transparent;
```
davidkircos, line 13: -webkit-tap-highlight-color: transparent;

src/constants/gridConstants.ts
```diff
@@ -4,3 +4,10 @@ export const CELL_TEXT_MARGIN_LEFT = 2;
 export const CELL_TEXT_MARGIN_TOP = -1;
 export const GRID_SIZE = 150;
 export const ZOOM_ANIMATION_TIME_MS = 250;
+export const LABEL_MAXIMUM_WIDTH_PERCENT = 0.7;
+export const LABEL_MAXIMUM_HEIGHT_PERCENT = 0.5;
+export const LABEL_PADDING_ROWS = 2;
+export const GRID_HEADER_FONT_SIZE = 9;
```
davidkircos, line 10: These are only used in `GridHeadings.ts`, they should be defined in that file

davidfig:
> Would it make sense to put this in the ViewportComponent.tsx instead of passing from QuadraticGrid?

Probably, but it requires a bunch of refactoring since usePointerEvents is React code, and ViewportComponent is mostly non-React. We'd have to pass the state functions into ViewportComponent and then rework the usePointerEvents into vanilla javascript for this to be better. I'm going to hold off on doing this.

src/core/gridGL/QuadraticGrid.tsx
davidfig: moved to types/cellReference. Probably not perfect but more generic.

src/core/gridGL/graphics/ViewportComponent.tsx
davidfig: i updated the comments.

viewport.dirty => forces a renderer.render() -- this can be set even when the viewport doesn't change to force a render
let dirty => forces a redraw of gridLines, axesLines, and headers followed by a renderer.render => this is only triggered by a moved or zoomed event from viewport

src/core/gridGL/graphics/axesLines.ts
davidfig: Regrettably. the axesLinesProps is a way for a React to update this component. Since the IProps is used in PixiComponentViewport.create, updating IProps does not change the PIXI.Ticker function. Updating this exported local variable does. It's a hack, but it's what I came up with to work with PixiComponentViewport's configuration.

src/core/gridGL/interaction/usePointerEvents.ts
davidfig: depends on the app. but updated

src/styles.css
davidfig: eslint doesn't like "-webkit-tap-highlight-color" as a style

src/constants/gridConstants.ts
davidfig: you don't want to centralize constants all in one place so they're easier to change?

src/core/gridGL/QuadraticGrid.tsx
davidfig: i uncommented it, but i think it was working w/o it. we can optimize down the road. I'll also add to my list a visual indicator (in HTML) of when pixi renders (behind a debug flag). that'll give us a better idea of what's happening under the hood.

src/constants/gridConstants.ts
davidkircos: I think because these are specific to the Header, I would think to look for them in the header

src/core/gridGL/graphics/axesLines.ts
davidkircos: Ok. Could you add that whole reply as a code comment?

davidkircos:
@davidfig I noticed one additional bug. When dragging to select cells, if you don't MouseUp over the canvas it gets stuck in the moving state.

I like to attach the events for mouseMove and MouseUp to the `window` because then it will always capture releasing the mouse even if it is outside of the window. Example Implementation for resizing the Code Editor https://github.com/quadratichq/quadratic/pull/119/commits/d09a1ea81394496f8c697a4e5f3e53cb195f4f0f#diff-2bc202dc21e9b27b3ad74739a0edda788d6d4da097d558b077a50d6fcc23276eR187

davidfig:
> I like to attach the events for mouseMove and MouseUp to the window because then it will always capture releasing the mouse even if it is outside of the window. Example Implementation for resizing the Code Editor https://github.com/quadratichq/quadratic/commit/d09a1ea81394496f8c697a4e5f3e53cb195f4f0f#diff-2bc202dc21e9b27b3ad74739a0edda788d6d4da097d558b077a50d6fcc23276eR187

Yup, I'll make the change tonight.

davidkircos (approved):

PR #128, opened 2022-09-25
Add numpy-like indexing

boaarmpit:
Enables following syntax:
```
1. grid[row                        , col                        ]
2. grid[row_slice_min:row_slice_max, col                        ]
3. grid[row                        , col_slice_min:col_slice_max]
4. grid[row_slice_min:row_slice_max, col_slice_min:col_slice_max]
```

TODO: Enable `first_row_header` parameter.
Options:
1. Pass parameter in dict. E.g. `grid[1:10, 1:10, {'first_row_header': True}]` ([ref](https://stackoverflow.com/questions/31916617/using-keyword-arguments-in-getitem-method-in-python))
2. Create two grid objects with and without built in `first_row_header` parameter. E.g. `grid` and `gridh`.

### DIFF
--- a/src/core/computations/python/run_python.py
+++ b/src/core/computations/python/run_python.py
@@ -26,11 +26,13 @@ def attempt_fix_await(code):
     code = re.sub(r"([^a-zA-Z0-9]|^)cell\(", r"\1await cell(", code)
     code = re.sub(r"([^a-zA-Z0-9]|^)c\(", r"\1await c(", code)
     code = re.sub(r"([^a-zA-Z0-9]|^)getCells\(", r"\1await getCells(", code)
+    code = re.sub(r"([^a-zA-Z0-9]|^)grid\[", r"\1await grid[", code)
 
     code = code.replace("await await getCell", "await getCell")
     code = code.replace("await await c(", "await c(")
     code = code.replace("await await cell(", "await cell(")
     code = code.replace("await await cells(", "await cells(")
+    code = code.replace("await await grid[", "await grid[")
 
     return code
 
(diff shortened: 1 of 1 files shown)


boaarmpit:
@davidkircos
Migrating your comment from my original PR.

> Does it make sense to put this interface directly on the cells object?
> 
> That way you could call cells() like a function with a row and column, or like a data frame cells[] with numpy like syntax.
> 
> This could be a more intuitive way to access data for people coming from the Pandas/Numpy world.
> 
> Feel free to pull against quadratichq/quadratic:main. We can have a full discussion there.
> 
> Thanks!

Regarding this comment: 

> Does it make sense to put this interface directly on the cells object

I agree that that would be nice, but I'm not sure if there is a way of adding the `__getitem__` method to a function in order to make it subscriptable with '[]'.

src/core/computations/python/run_python.py
```diff
@@ -189,13 +191,60 @@ async def cell(p0_x, p0_y):
...
+    grid = Grid()
+
     globals = {
         "getCells": getCells,
         "getCell": getCell,
         "c": c,
         "result": None,
         "cell": cell,
         "cells": cells,
+        "grid": grid,
```
davidkircos, line 247: can simplify to `"grid": Grid(),`

davidkircos:
@boaarmpit Thinking about this issue a bit more. I like the idea of being able to directly reference the grid via subscripts.

I think this is a net addition, if you rebase from master and mark the PR as ready for review. I think we can merge. I made one simple comment on the code.

Issue #129, opened 2022-09-29
Double Clicking Top Left Corner of Headers Should Move to Cell 0,0


davidkircos:
When you double click on this location (marked with the red arrow in the screenshot below), no matter where you are in the grid you will be navigated back to the upper right cell position at 0,0. 
<img width="1676" alt="Screen Shot 2022-09-29 at 10 27 38 PM" src="https://user-images.githubusercontent.com/3479421/193190425-bb724dc7-c60c-43bc-b379-e4fe8ca22ffe.png">

PR #130, opened 2022-09-30
Clicking top left header highlights cell at 0,0

wolewnicki:
This issue should resolve https://github.com/quadratichq/quadratic/issues/129.

I added a structure the works sort of like a `switch` expression in c# or `match` statement in rust. I called it `match` just like in rust. I believe this structure is useful to quickly handle possible states in the code. I was able to reduce the nesting in this function by converting to a `match` over if statements. I also converted this function to an async function to again reduce another block and hopefully improve readability.

I would love feedback and I apologize if this is too large of a change for this issue. I am hoping to use this utility and contribute further.

### DIFF
--- a/src/core/gridGL/interaction/onDoubleClickCanvas.ts
+++ b/src/core/gridGL/interaction/onDoubleClickCanvas.ts
@@ -1,8 +1,10 @@
 import { GridInteractionState } from '../../../atoms/gridInteractionStateAtom';
 import { EditorInteractionState } from '../../../atoms/editorInteractionStateAtom';
 import { GetCellsDB } from '../../gridDB/Cells/GetCellsDB';
+import { match } from '../../../utils/match';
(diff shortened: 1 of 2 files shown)


davidkircos:
Hi @wolewnicki thank you for making your first contribution. The match statement is pretty cool, I can see how it would be useful.

I believe that we want to write code for humans and that great code is easy to read and understand. It seems like this change makes the logic in `interaction/onDoubleClickCanvas.ts` more difficult to read and understand. Could you help me understand the benefit in this case?

For issue #129 the idea is that when you double click on this location (marked with the red arrow in the screenshot below), no matter where you are in the grid you will be navigated back to the upper right cell position at 0,0. Does that make sense?
Screen Shot 2022-09-29 at 10 27 38 PM

davidkircos:
Issue closed in https://github.com/quadratichq/quadratic/pull/133

PR #133, opened 2022-10-09
Highlight header

davidfig:
- [x] Highlight headings on selection
- [x] Highlight headings on multiple selection
- [x] Show first and last selected column/row (when zoomed out)
- [x] Hide default row/column labels when overlapping selected column/rows headings (when zoomed out)
- [ ] navigate to (0, 0) when double clicking the corner

### DIFF
--- a/src/core/gridGL/QuadraticGrid.tsx
+++ b/src/core/gridGL/QuadraticGrid.tsx
@@ -76,6 +76,13 @@ export default function QuadraticGrid() {
   // Interaction State hook
   const [interactionState, setInteractionState] = useRecoilState(gridInteractionStateAtom);
 
+  useEffect(() => {
+    gridHeadingsGlobals.interactionState = interactionState;
+    if (viewportRef.current) {
+      viewportRef.current.dirty = true;
+    }
+  }, [interactionState]);
+
   // Editor Interaction State hook
   const [editorInteractionState, setEditorInteractionState] = useRecoilState(editorInteractionStateAtom);
 
--- a/src/core/gridGL/graphics/ViewportComponent.tsx
+++ b/src/core/gridGL/graphics/ViewportComponent.tsx
@@ -77,16 +77,17 @@ const PixiComponentViewport = PixiComponent('Viewport', {
           const bounds = viewport.getVisibleBounds();
           cull.cull(bounds);
 
+          gridHeadings({
+            viewport,
+            headings,
+            graphics: headingsGraphics,
+            labels,
+            corner,
+          });
+
           if (dirty) {
             gridLines({ viewport, graphics: gridGraphics });
             axesLines({ viewport, graphics: axesGraphics });
-            gridHeadings({
-              viewport,
-              headings,
-              graphics: headingsGraphics,
-              labels,
-              corner,
-            });
             dirty = false;
           }
 
--- a/src/theme/colors.ts
+++ b/src/theme/colors.ts
@@ -7,6 +7,7 @@ export const colors = {
   highlightYellow: 0xffc800,
   independence: 0x5d576b,
   headerBackgroundColor: 0xffffff,
+  headerSelectedBackgroundColor: 0xeeeeee,
   headerCornerBackgroundColor: 0xffffff,
   lightGray: '#f6f8fa',
   mediumGray: '#cfd7de',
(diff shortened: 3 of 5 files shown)


src/core/gridGL/interaction/usePointerEvents.ts
```diff
@@ -105,9 +165,20 @@ export const usePointerEvents = (
     setPointerMoved(false);
   };
 
+  const changeMouseCursor = (world: PIXI.Point): void => {
+    // todo: this is not ideal -- need a better way of getting canvas
+    const canvas = document.querySelector('canvas');
```
davidkircos, line 170: I've used `document.getElementById('QuadraticCanvasID')` in another function. 

We could export QuadraticCanvas from a helper file.

src/theme/colors.ts
```diff
@@ -7,6 +7,7 @@ export const colors = {
   highlightYellow: 0xffc800,
   independence: 0x5d576b,
   headerBackgroundColor: 0xffffff,
+  headerSelectedBackgroundColor: 0xeeeeee,
```
davidkircos, line 10: what do you think about `0xf2fbff` for `headerSelectedBackgroundColor`?

davidfig: I preferred the light gray. But the selection color works as well.

Screen Shot 2022-10-09 at 6 54 17 PM

src/core/gridGL/interaction/usePointerEvents.ts
davidfig: I can get the canvas from useApp(), but it requires passing a bunch of props to send it to the right place (since it needs to be in the right context to call it). This works for now.

davidkircos: Ok sounds good. Can you change the call to get by ID? This way we can at least find all instances where we pull the canvas this way in the future.

src/core/gridGL/graphics/ViewportComponent.tsx
```diff
@@ -77,16 +77,17 @@ const PixiComponentViewport = PixiComponent('Viewport', {
           const bounds = viewport.getVisibleBounds();
           cull.cull(bounds);
 
+          gridHeadings({
+            viewport,
+            headings,
+            graphics: headingsGraphics,
+            labels,
+            corner,
+          });
```
davidkircos, line 86: Curious why this has to run even if the viewport isn't dirty

src/core/gridGL/interaction/usePointerEvents.ts
davidfig: i tried that originally but it didn't work. i'll try again.

src/core/gridGL/graphics/ViewportComponent.tsx
davidfig: Because I don't know if the cell changed and the heading backgrounds need to be updated. I can do a better job caching the last selection values and then doing a compare to lighten the load.

davidfig: Added.

src/core/gridGL/QuadraticGrid.tsx
```diff
@@ -76,6 +76,13 @@ export default function QuadraticGrid() {
   // Interaction State hook
   const [interactionState, setInteractionState] = useRecoilState(gridInteractionStateAtom);
 
+  useEffect(() => {
+    gridHeadingsGlobals.interactionState = interactionState;
+    if (viewportRef.current) {
+      viewportRef.current.dirty = true;
+    }
+  }, [interactionState]);
```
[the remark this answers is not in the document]
davidfig: interactionState isn't passed to ViewportComponents, which triggers the viewport.dirty.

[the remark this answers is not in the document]
davidkircos: We could `const [interactionState] = useRecoilState(gridInteractionStateAtom)` on the ViewportComponent, and then set dirty when it changes.

The downside of using useEffect on QuadraticGrid, means that all the other stuff in this file is rerun which is a bit expensive. Not sure if it is needed here or not. 

I do notice that when I interact with the grid the QuadraticGrid component runs 5 times.

We could also probably just put this code directly on the component. Because the component will re-run whenever the interaction state changes. No need for the useEffect.
```
    if (viewportRef.current) {
      viewportRef.current.dirty = true;
    }
```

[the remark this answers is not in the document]
davidkircos: You can ignore this. Resolved. There is no issue.

davidkircos (approved):

PR #160, opened 2022-11-19
try firebase

davidkircos:


davidkircos:
Seems like a good amount of work to set up. Probably easier to stick with AWS Amplify for now.

PR #165, opened 2022-12-03
Rust core user actions

davidkircos:


quadratic-core/src/actions/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::codestore::CodeCell;
+use crate::grid::{Cell, Command, GridController};
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum UserActions {
+    RunCell(Pos),
+    SetCellValue(Pos, String),
+    SetCodeValue(Pos, Option<CodeCell>),
+    // PasteCells(Pos),
+    // CopyCells(Rect), // Does not add anything to the undo stack...
```
HactarCE, line 13: We can modify the transaction code to not save anything if no `Command`s are executed

quadratic-core/src/grid/controller.rs
```diff
@@ -347,6 +347,10 @@ impl GridController {
         &self.grid
     }
 
+    pub fn get_graph(&self) -> &DGraphController {
```
HactarCE, line 350: `graph()` is sufficient; no `get_`

PR #183, opened 2023-01-18
feat: command palette

jimniels:
Context: [research & design document](https://www.notion.so/Command-Bar-f220c5f1640740c3995787d915a76cc9).

# To Test

- `CMD + P` shows/hides the command palette
  - "Actions" button in upper right also triggers the command palette
- By default, list opens with all results. Typing filters list to matching results
- Up/down arrow keys highlights active selection in list
- `Enter` or click runs the associated action of the active item in the list
- Commands with shortcuts show keyboard shortcut combination
- Commands with state show a checkbox to represent on/off
- `ESC` or clicking outside the command palette closes it

---

### Development Tasks

- [x] Design tweaks (align icons, fix padding/spacing, etc)
- [x] Architect solution for and wire up actions associated with command
- [x] Fuzzy search
  - Both for _finding_ commands and possibly _sorting_ matches (e.g. "Zo" would prioritize "**Zoo**m" matches over "Borders: Apply hori**zo**ntal"
  - This could be an enhancement later, though other tools have this feature
- [x] Interaction paradigm
  - click outside the command palette closes it
  - can't open the command palette while the code editor is open?
- [x] Inner scrolling of list
- [x] Audit and include all existing (coming soon) commands in Quadratic today
- [x] Centralize access to keyboard shortcut symbols
- [x] Find `@TODO`s in the code and move them into PR description (and fix)
- [x] Make button to access command bar in UI (with hotkey shortcut preview)
- [x] Make sure list order matches between menus and command palette
- [x] Fix `CMD` + `P` when cell has active data entry

### Deliberately excluded from this PR

- Centralized place to register all keyboard shortcuts?
- Support "Recent" functionality where 3-5 most recent commands appear at the top of the list?
- Centralized place/data for all commands, whether in the top menu or the command palette?
- Actions in the Toolbar today that aren't (yet) in the command palette:
  - All "coming soon..." actions
  - File -> Open recent
  - File -> Open sample
  - Format: Clear all (clears border & fill & any other formatting)
  - Format: Clear fill color
  - Format: Fill color
    - The idea of nested commands will come later
  - Border style
  - Show debug menu

### Questions

- [x] Should we include "Coming soon" items in the list? If so, should we indicate them as such with a label, or is just "disabled" enough?
  - **Decision**: Gonna say "no" to this right now. As new actions are supported, we'll add them to the command list. No need to populate it with stuff you can't do. The toolbar menus are a sufficient place for "coming soon" actions now.
- [x] Should we show a list of results by default (e.g. on an empty search)? If so, what's the right prioritization of ordering?
  - **Decision**: Yes, we will show all by default (so you can scroll through and see them). Maybe this changes in the future. For now, we show all based on our own best decisions around ordering.
- [x] Interaction paradigm definition
  - Does command palette show when a cell is focused with data input? or only on when a cell is selected but not actively accepting data input?
    - **Decision**: No. Command palette won't trigger when cell actively has input.
  - Are all actions applicable all of the time? e.g. "Apply borders" would only be applicable to the actively selected cell, but "Undo" would be applicable to both the actively selected cell _AND_ the code editor.
    - Example: if you have the cursor focused in the code editor and you're typing something and hit `CMD + P` does the command palette open? Presumably yes, because you might want to run "undo" for example. But then what about all those text/cell formatting options, what happens if you run those? Does it apply to the selected cell, even though your cursor is over in the code editor? what about undo? does the undo action apply to the code editor or the selected cell?
    -  **Decision**: Will try allowing command palette to open, even with code editor, and see how intuitive it is to implicitly understand actions apply to cells most often.


src/atoms/editorInteractionStateAtom.ts
```diff
@@ -5,13 +5,15 @@ import { CellTypes } from '../core/gridDB/gridTypes';
 export interface EditorInteractionState {
   showCellTypeMenu: boolean;
   showCodeEditor: boolean;
+  showCommandPalette: boolean;
```
jimniels, line 8: Is this the right place for this state?

Felt like I needed something global and this is what I could ascertain we do with global state. Seemed like a logical place to put this particular piece of data...

davidkircos: I think this is a good place to control add it. That way anything that sets the state can make sure only a valid state of menus is shown

src/core/gridGL/interaction/keyboard/keyboardViewport.ts
```diff
@@ -2,26 +2,46 @@ import { Viewport } from 'pixi-viewport';
...
 
   if ((event.metaKey || event.ctrlKey) && event.code === 'Minus') {
     zoomInOut(viewport, viewport.scale.x * 0.5);
-    event.preventDefault();
     return true;
   }
 
   if (event.shiftKey && event.code === 'Digit1') {
     zoomToFit(sheet, viewport);
-    event.preventDefault();
```
jimniels, line 24: These don't need to be here because the function that calls `keboardViewport()` already does the preventDefault

package.json
```diff
@@ -33,6 +33,7 @@
     "color": "^4.2.3",
     "date-fns": "^2.28.0",
     "fontfaceobserver": "^2.3.0",
+    "fuzzysort": "^2.0.4",
```
jimniels, line 36: Initially I tried [fuse.js](https://github.com/farzher/fuzzysort) but honestly it was not working very good, even after trying to calibrate the appropriate options.

Then I found [fuzzysort](https://github.com/farzher/fuzzysort) which provided exactly the kind of UI/X I wanted right out of the box — and it's a small lib, no deps

src/utils/isMac.ts
```diff
@@ -1,4 +1,2 @@
 // from https://stackoverflow.com/a/70519514/1955997
-export function isMac(): boolean {
-  return /(macintosh|macintel|macppc|mac68k|macos)/i.test(navigator.userAgent);
-}
+export const isMac: boolean = /(macintosh|macintel|macppc|mac68k|macos)/i.test(navigator.userAgent);
```
jimniels, line 2: Not sure if there's a reason to execute this everytime `isMac()` gets called as the userAgent shouldn't change...so I made it a var with a stored value rather than a function that recalculates each time

src/ui/menus/CommandPalette/ListItems/Help.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { OpenInNew } from '@mui/icons-material';
+import { DOCUMENTATION_URL, BUG_REPORT_URL } from '../../../../constants/urls';
+import { CommandPaletteListItemDynamicProps } from '../CommandPaletteListItem';
+
+interface Props extends CommandPaletteListItemDynamicProps {
+  sheetController: any;
+  app: any;
+}
+
+export const CPLIHelpViewDocs = (props: any) => {
```
davidkircos, line 11: I'd recommend naming them like this `HelpViewDocsCPLI` that way auto-complete is easer in VSCode :)

public/index.html
```diff
@@ -50,17 +50,10 @@
     .App {
       height: 100%;
     }
+    /* Disable inertial scroll
+       https://stackoverflow.com/questions/28103125/disable-inertia-scroll-for-single-page-webapp */
+    body {
+      overflow: hidden;
+    }
```
jimniels, line 57: This isn't enough. Will have to fix this...

jimniels: Two finger pinch on trackpad breaks if you remove the script.

src/ui/icons/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+// Built on doc guidance from material UI:
+// https://mui.com/material-ui/icons/
+//
+// Icons designed originally on 24 x 24px artboard. Figma file:
+// https://www.figma.com/file/HdhrQ6vUZH4quMl7WFX6Vd/Custom-Icons?node-id=0%3A1&t=t5GeoagkcP0W7hK2-1
+//
+// Contents must be manually copy/pasted from exported .svg files.
+// Probably could make this better in the future by using something like:
+// https://github.com/gregberge/svgr
```
jimniels, line 11: Added these icons to get nice icon representations for all the border commands.

CleanShot 2023-01-24 at 12 08 38@2x

Material icons didn't have a match for these (surprisingly) so I created our own additions


They weren't i

src/ui/menus/CommandPalette/ListItems/Borders.tsx
```diff
@@ -0,0 +1,115 @@
...
+          action={() => {
+            clearBorders();
+          }}
+          icon={<BorderClear />}
+        />
+      );
+    },
+  },
+  ...[
+    { label: 'Borders: Style thin', icon: <BorderThin />, changeBordersArgs: { type: BorderType.line1 } },
```
jimniels, line 80: These don't actually work right now. Just passing an updated border type doesn't change the border. This might be a bug but will likely require a fix in the `changeBorder` function returned from `useBorders`

src/core/gridGL/interaction/keyboard/keyboardViewport.ts
```diff
@@ -1,33 +1,43 @@
...
   if (!viewport) return false;
 
+  if ((event.metaKey || event.ctrlKey) && event.code === 'KeyP') {
+    setEditorInteractionState({
+      ...editorInteractionState,
+      showCommandPalette: !editorInteractionState.showCommandPalette,
+    });
+    return true;
+  }
+
```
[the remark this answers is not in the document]
jimniels: Fixed. `CMD + P` will not trigger anything on cell input

davidkircos (approved):
This is an awesome improvement to the product! Really great work on a complex PR.

Merge `development` with the prettier fix into this branch, then good to merge into `development`!

PR #187, opened 2023-01-23
Improving update cells.

dk8996:
This is small change but it should improve performance when cells to update is large. Still trying to figure out how to write a test for this.


davidkircos:
I am not sure how to get unit tests to run with Github Actions from an outside PR. 

Should be able to test this change locally with `npm run test:unit`

davidkircos:
It looks like unit tests fail in the dev branch, unrelated to this change. But the file `src/core/transaction/tests/cell_evaluation.test.ts` contains tests that test `src/core/actions/updateCellAndDCells.ts` 

Probably pretty easy to add a test to cover this change.

PR #191, opened 2023-01-24
Development clipboard improvement

davidkircos:


davidkircos:
@jimniels could you review this PR for functionality?

Working on cut/copy/paste within Quadratic and from outside of Quadratic (sheets, excel).

FYI Still doesn't work in Firefox. But shouldn't throw an error now.

jimniels:
Functionality looks pretty good. Super excited to see Cut coming!

A couple points of things I saw (not sure how much of this is or is not in scope for this PR):

- Visual indicator when doing CUT - excel/sheets both provide a visual cue (a dashed border in a highlight color) on the cells from which you cut (excel’s is animated). This is super useful as a when you're in that intermediate step of having cut but not yet pasted.
  - Jan-24-2023 16-07-52
- Undo: if you cut/paste and then hit undo, both sheets and excel jump you back to where you were before doing CUT. With Quadratic it takes two undos to get you there (first undo removes the paste, second undo puts it back in the original cell)

These are issues that seem to be present for both cut/paste as well as copy/paste, so they're probably things that should be addressed outside of this PR?

- Formatting: formatting isn't preserved when cutting (or copying for that matter) and pasting. This is true of cutting/copying inside of Quadratic but also across applications (e.g. cut/copy + paste from excel to sheets and sheets tries to preserve formatting, though it's not always perfect). 
  - Example: here’s a cut/paste in quadratic where I want to move that whole top formatted row
  - Jan-24-2023 16-11-59
- Cut/copy + paste the contents of a cell that generates a lot of rows blocks the main thread and makes the UI unresponsive until the cut/copy transaction completes. Not sure how to resolve this but it does feel weird that it's not snappy fast and makes the whole grid freeze and become unresponsive for a few seconds (same is true if you do CMD+Z undo on a large transaction)
  - Here's an example where the use case would be: take the top two rows, which are generated programatically (the second row draws like 1,000 rows) and shift the whole thing down a couple rows. Trying to do a CUT makes the browser become unresponsive (and ultimately, it doesn't cut the whole thing
  - <video src="https://user-images.githubusercontent.com/1316441/214443497-a794ab9f-765b-48cc-a234-e0397926664f.mp4" />

davidkircos:
@jimniels , thanks for the feedback. 

Cut with the visual indicator will have to be a future story. It's a bit more work, and not sure if it's worth it right now.

Copy / Paste formatting should be a future story. I've created a PR for it.

Performance issues are known. We need a general way to bulk update cells. The current way is pretty inefficient. This will also be addressed in a future issue.

- If you didn't notice any other problems, I will go ahead and merge this. It seems like a net improvement!

PR #195, opened 2023-01-26
fix: prettier SVG error & warnings

jimniels:
I keep getting the CI failing on prettier checks because of the logo svg. You can see the command run locally here:

CleanShot 2023-01-25 at 17 07 53@2x

For now, I suggest ignoring SVG as [it's not even officially supported (yet) by prettier](https://github.com/prettier/prettier/issues/5322). 

This adds prettierignore for SVGs and then I ran `--write` to autofix the other issues. This should fix the CI issue.


davidkircos (approved):
:ship:

PR #198, opened 2023-01-27
feat: "Go to" menu

jimniels:
# Overview

"Go to" or "Jump to" command which accepts coordinates for a cell or range and navigates the user to that position on the grid. The idea is that no input is "invalid" so users can key in whatever feels most natural to them as a set of coordinates, e.g. `1 1 5 5` or `1, 1, 5, 5` or `(1, 1) - (5, 5)` are all valid input.

**Implementation notes**:

- "Go to" shortcuts from other apps:
  - Excel: `CMD` + `G`
  - Sheets: `CMD` + `J`
  - VSCode: `CTRL` + `G`

## To Test

a.k.a things to test on this PR

- [x] Toggle the "Go to" menu by:
  - [x] Clicking on "Cursor" in the lower left
  - [x] Clicking on "Selection" in the lower left
  - [x] Shortcut `CMD` + `G` or `CMD` + `J`
  - [x] Menu option in the command palette
- [x] Handle user input
  - [x] Incomplete syntax (we auto-fill range)
  - [x] Edge case: backwards ranges `(1,1) - (0,0)`
  - [x] Edge case: Ranges of 1 `(0,0) - (0,0)`
  - [x] Negative cell references
- [x] Scroll "go to" cell/selection into viewport after choosing a value
- [x] Write unit tests for input -> coordinates function
- [x] Fix typescript errors and ts-ignore
- [x] Fix Go to menu and Command palette so they appear at a fixed position from the top (this way closing one menu and opening another in sequence makes them appear in the same location on screen)

Notes:

- `CMD` + `G`/`J` triggers opens find menu in browser when actively editing a cell
- Passed on putting "Go to" in the top menu of the app somewhere, as it’s accessible from the footer status bar instead.
- Commas are accepted as delimiters between coordinates, NOT as delimiters in larger numbers, e.g. if you typed `10,250` that takes you to `10` and `250`, it is not interpreted as ten thousand, two hundred and fifty. 
- I didn't limit how far out you can go. If you try to jump to cell one million, one million, it will take you there. But if you try to zoom out from there the app basically freezes up.


src/ui/menus/GoTo/GoTo.tsx
```diff
@@ -0,0 +1,137 @@
...
+                primary={`Go to ${value && isValidInput ? '' : 'current'} ${
+                  coordinates.length === 1 ? 'cell' : 'range'
+                }: ${coordinates.map(([x, y]) => `(${x}, ${y})`).join(' – ')}`}
+              />
+            </ListItemButton>
+          </ListItem>
+
+          <Divider />
+          <ListItem disabled>
+            <ListItemText primary="Specify a cell – 0 0 — or a range — 0 0 -5 -5" />
```
davidkircos, line 99: This example is confusing. I suggest:

`Specify a cell "0, 0" or a range "0, 0, 10, 10"`

src/ui/menus/GoTo/GoTo.tsx
```diff
@@ -0,0 +1,137 @@
...
+      showGoToMenu: false,
+    });
+    setValue('');
+  };
+
+  const parsedCoordinates = getCoordinatesFromInput(value);
+  const isValidInput = parsedCoordinates.length > 0;
+  const coordinates = isValidInput
+    ? parsedCoordinates
+    : [[interactionState.cursorPosition.x, interactionState.cursorPosition.y]];
```
davidkircos, line 46: It might make more sense to default to `0, 0` than the current position. 

Adds a quick way to get back to the origin. Open GoTo Menu, and press Enter.

src/core/gridGL/interaction/keyboard/keyboardViewport.ts
```diff
@@ -16,11 +16,21 @@ export function keyboardViewport(options: {
   if ((event.metaKey || event.ctrlKey) && event.code === 'KeyP') {
     setEditorInteractionState({
       ...editorInteractionState,
+      showGoToMenu: false,
       showCommandPalette: !editorInteractionState.showCommandPalette,
     });
     return true;
   }
 
+  if ((event.metaKey || event.ctrlKey) && event.code === 'KeyG') {
```
davidkircos, line 25: Could you also make Command J work? I keep trying it!

src/ui/menus/GoTo/GoTo.tsx
jimniels: I know, it's confusing. I wanted to try and communicate: "you can type as few characters as you want, just spaces between numbers is the minimum you need" but i feel like that almost looks strange as an example `0 0 10 10` and commas is more human friendly for parsing a string of digits. So i'll go with that for now

src/ui/menus/GoTo/GoTo.tsx
jimniels: Love it

src/core/gridGL/interaction/keyboard/keyboardViewport.ts
jimniels: "go to", "jump to" 

you say tomato, i say tomato

👍

jimniels:
Going to a range whose selection is not visible at 100% zoom feels a little incomplete. Right now it just shows everything in the viewport as being selected.

CleanShot 2023-01-31 at 15 47 06@2x

It feels like this needs an additional "zoom to selection" where it would zoom you out to fit the selected range you specificed, e.g.

CleanShot 2023-01-31 at 15 45 40@2x

Perhaps we can get to that later

src/ui/components/dialog-menu.css
```diff
@@ -0,0 +1,5 @@
+/* Any command palette style menu should show up in a fixed position at the top */
+.MuiDialog-container {
+  align-items: flex-start;
+  padding-top: 48px;
+}
```
jimniels, line 5: I don't love this solution. I tried a couple other solutions but couldn't find the right combination of props based on the MUI docs. This works, so I'm going with this for now. We'll probably want to share more of this component code in the future, so we can refactor and refine then.

src/ui/menus/CommandPalette/ListItems/Edit.tsx
```diff
@@ -97,6 +99,27 @@ const ListItems = [
...
+          {...props}
+          action={() => {
+            setEditorInteractionState({
+              ...editorInteractionState,
+              showGoToMenu: true,
+            });
+          }}
+          icon={<East />}
+          shortcut="G"
+          shortcutModifiers={[KeyboardSymbols.Command]}
```
jimniels, line 118: `CMD` + `J` also works, but we're advertising it as `CMD` + `G`

src/ui/menus/CommandPalette/styles.css
```diff
@@ -0,0 +1,5 @@
+/* Any command palette style menu should show up in a fixed position at the top */
+.MuiDialog-container {
+  align-items: flex-start;
+  padding-top: 48px;
+}
```
jimniels, line 5: I don't love this solution (or [this](https://github.com/quadratichq/quadratic/pull/198/files#diff-066419d9cfa1a3ed333d9d926380811b788e18980afe01be5731ba9d5faca404)) but I tried a couple other solutions but couldn't find the right combination of props based on the MUI docs.

This works, so I'm going with this for now. We'll probably want to share more of this component code in the future as we create more command-palette-like menus, so we can refactor, refine, and share code then.

src/ui/menus/CommandPalette/CommandPalette.tsx
```diff
@@ -49,7 +50,6 @@ export const CommandPalette = (props: Props) => {
     const el = document.querySelector(`[data-command-bar-list-item-index='${selectedListItemIndex}']`);
     if (el) {
       el.scrollIntoView({
-        behavior: 'smooth',
```
jimniels, line 52: @davidkircos this takes away the "smooth scroll" when keying up/down through the command palette list.

src/ui/menus/CommandPalette/ListItems/Edit.tsx
davidkircos: What's the logic for G over J?

src/ui/menus/CommandPalette/CommandPalette.tsx
davidkircos: Nice, good find.

davidkircos (approved):

src/core/gridGL/interaction/keyboard/keyboardViewport.ts
```diff
@@ -16,11 +16,21 @@ export function keyboardViewport(options: {
   if ((event.metaKey || event.ctrlKey) && event.code === 'KeyP') {
     setEditorInteractionState({
       ...editorInteractionState,
+      showGoToMenu: false,
       showCommandPalette: !editorInteractionState.showCommandPalette,
     });
     return true;
   }
 
+  if ((event.metaKey || event.ctrlKey) && (event.code === 'KeyG' || event.code === 'KeyJ')) {
```
davidkircos, line 25: we should add a return false if `event.altKey` is pressed before looking at any of these keyboard events

davidkircos: Like right after line `14`

src/ui/menus/CommandPalette/ListItems/Edit.tsx
jimniels: I went with G because the menu's name is "Go To" in both the code and the UI (e.g. we don't call it "Jump to..." in the command palette).

I'm not particularly strong either way whether we call it "Jump to" in the app or "Go to" but whatever we call it, i feel like that's the advertised keyboard shortcut.

src/core/gridGL/interaction/keyboard/keyboardViewport.ts
jimniels: Good catch. [Done](https://github.com/quadratichq/quadratic/pull/198/files#diff-17cd90e926d09d4ec633c7255878e0797a1e29ae77b086e7c3d32f8b2affca32R14)

davidkircos (approved):

Issue #207, opened 2023-02-02
Toggling view options doesn't work without first panning the grid


jimniels:
- Do a fresh load of the page
- Toggle a view preference (from the command palette), like "Show Row and Column Headers"
- Note that the row/column headers still remain on screen. However, as soon as you pan the canvas, the preference kicks in and the row/column headers disappear.

![Feb-02-2023 11-36-43](https://user-images.githubusercontent.com/1316441/216419252-dcade54e-8b31-4b68-9df0-b5585158f4f8.gif)

This seems to be the case for any view-related preference. They only work if you've already panned the canvas at some point in time. But upon initial page load, before panning the canvas, they don't work.

PR #208, opened 2023-02-02
feat: Presentation Mode

jimniels:
- Adds a feature to show/hide application UI (which basically is a shortcut to show/hide all view preferences).
- Adds a feature to show/hide debug menu from the command palette

Feb-02-2023 16-29-20

CleanShot 2023-02-02 at 16 37 58@2x

Notes:

- Figma calls this "Show/Hide UI" and you can trigger it with `CMD` `.` or `CMD` `\`
- VSCode has a "Zen mode" `CMD` `K` `Z`
- **NOTE: `\` is used as "clear formatting" in Sheets**

Todos:

- [x] Add "Hide application UI" to command palette & top menu
- [x] Add "Show debug menu" to command palette
  -  If debug menu is open, it closes when "Hide UI" is toggled (but doesn't re-open when you turn off "Hide UI")
- [x] Hide application UI controls visibility of debug menu
- [x] `CMD` + `\` shortcut for "Hide application UI"
- [x] ~Maybe rename it to "Show/Hide UI" like figma? Need a little more thought...~ Will call it "Hide application UI"


jimniels:
This is kind of hard to test without [this issue being resolved first](https://github.com/quadratichq/quadratic/issues/207).

davidkircos:
I use this in Figma a lot. I like calling it "Show / Hide UI" that is super clear to me. Also Command + / seems like a good short cut.

Love this feature.

jimniels:
@davidkircos I don't love the name "Show/Hide UI" because it feels like it breaks uniformity with how we name all our other view preferences, e.g. we already have other view prefs which are labeled "Show ____"

- Show row and column headings
- Show grid axis
- Show grid lines
- Show cell type outlines

In theory, they're doing the same thing as this, so shouldn't they be called "Show/Hide ____" e.g.

- Show/hide row and column headings
- Show/hide grid axis
- Show/hide grid lines
- Show/hide cell type outlines
- Show/hide UI

But the "shown" or "hidden" state is communicated by checkmarks in our app, so you don't really need the dual label. For example, this is Figma today: can you tell, from only this screenshot, whether it's showing or hiding the UI? i.e. does the checked box  mean it's showing or hiding when the label says both "Show" and "Hide"?

CleanShot 2023-02-02 at 14 49 47@2x

But to your point, perhaps its obvious enough when you use the app.

The options in my head are:

1. Give it a name, like "Zen mode" or "Presentation mode", as it's not just toggling individual preferences like the others but a grouping of them. Then you can easily have a checkbox like `[X] Presentation mode` means it's on (unchecked means its off)
2. Have it match the pattern of the other labels and maybe give it a more precise name, like "Show grid UI"
3. Just roll with "Show/Hide UI"

Typing this out is an exercise in thinking and maybe just "Show/hide UI" is enough and people will just use it without thinking about it too much — but I don't have to love how it breaks consistency in labeling with the other menu items ha

davidkircos:
How about “Show application UI”?

jimniels:
I like that.

In theory we could hide _the entire UI_ with that command

CleanShot 2023-02-02 at 15 01 13@2x

You'd probably want some little temporary floating UI, like "return to app (`CMD` + `/`)"

But I like keeping the chrome for now.

CleanShot 2023-02-02 at 15 01 27@2x

I'll roll with that.

jimniels:
Ok, I like calling it "Hide application UI"

CleanShot 2023-02-02 at 15 06 47@2x

That distinguishes it apart from the others as a kind of toggle for all of them (which are on by default). Same for the "View" menu:

CleanShot 2023-02-02 at 15 42 51@2x

When you toggle any "Show" it exits you from the "hide application mode" which I think feels right.

src/ui/menus/TopBar/SubMenus/useGridSettings.ts
```diff
@@ -24,10 +24,15 @@ interface GridSettingsReturn {
...
   setShowCellTypeOutlines: (value: boolean) => void;
+
+  // Return fn to derive state from other values
+  hideApplicationUI: () => boolean;
+  toggleApplicationUI: () => void;
 }
 
 export const useGridSettings = (): GridSettingsReturn => {
   const [settings, setSettings] = useLocalStorage('gridSettings', defaultGridSettings);
+  const [showDebugMenu, setShowDebugMenu] = useLocalStorage('showDebugMenu', false);
```
jimniels, line 35: I'm unsure what to think about this....

The file is named `useGridSettings` so it's about settings in the grid.

But I think, more generically, it's really about "View" settings because "Hide application UI", at least in my mind, should control showing/hiding the "Debug menu" (which, I think, is not technically a "grid setting" because it lives at the level of the application).

So that's a long way to say: I think this code works, but maybe is starting to break out of its original abstraction. Not sure if that's something we should consider refactoring now, or just leave it be and roll with it. I tried to follow the conventions in this file.

src/ui/menus/TopBar/SubMenus/useGridSettings.ts
```diff
@@ -85,5 +90,41 @@ export const useGridSettings = (): GridSettingsReturn => {
...
-  return { ...settings, setShowGridAxes, setShowHeadings, setShowGridLines, setShowCellTypeOutlines };
+  const hideApplicationUI = useCallback(() => {
+    return !(
+      settings.showHeadings ||
+      settings.showGridAxes ||
+      settings.showGridLines ||
+      settings.showCellTypeOutlines ||
+      showDebugMenu
+    );
+  }, [settings]);
```
jimniels, line 101: This is derived state, so i'm returning it as a function you have to call to get the boolean representation of state vs. an actual boolean.

src/ui/menus/TopBar/SubMenus/useGridSettings.ts
```diff
@@ -85,5 +90,41 @@ export const useGridSettings = (): GridSettingsReturn => {
...
+    });
+
+    // Don't show the debug menu when you unhide the UI.
+    // Only hide it if it's visible when the user toggles "Hide UI"
+    if (!isHidden && showDebugMenu) {
+      setShowDebugMenu(false);
+    }
+
+    emitGridSettingsEvent();
+  }, [setSettings, hideApplicationUI, emitGridSettingsEvent, showDebugMenu, setShowDebugMenu]);
```
jimniels, line 119: I wanted each place that consumes this function to not have to worry about passing a value. They can just say "whatever state it's in, toggle it" then they don't have to get the state and figure it out themselves in each place.

It breaks the convention for the other pieces of state in this file, but maybe that's ok because it's not a piece of stored state anyway (it's derived state).

davidkircos:
@jimniels I rewrote this to work as Show/Hide UI override, so it saves your other settings and just overrides Show/Hide UI

jimniels:
@davidkircos this is great. Definitely better to have "show/hide UI" be an _override_ of your current settings vs. a replacement/reset of them. That's a nice addition.

A couple of thoughts on the current state of this.

## Keyboard shortcuts

We still have a clash with keyboard shortcuts.

Both "Clear formatting" and "Show/hide UI" are `CMD` + `\`.

It looks like you might've changed clear formatting to `CMD` + `/` (??) but it's not updated in all places. For example, the formatting toolbar still shows `CMD` + `\`

CleanShot 2023-02-22 at 20 51 38@2x

Which is what the command palette says for 'show/hide ui'

CleanShot 2023-02-22 at 20 49 14@2x

This is my two cents:

- Google Sheets uses `CMD` + `\` for clear formatting.
- Figma uses `CMD` + `\` for clear formatting _as well as_ `CMD` + `.`
- Sketch uses `CMD` + `.` for clear formatting

Figma probably uses `CMD` + `.` to appease Sketch users (and idk why they use `CMD` + `\`, perhaps a legacy shortcut?) I'm feeling like maybe we should keep the Google Sheets shortcut and use the figma/sketch for show/hide UI. In other words, in Quadratic:

- `CMD` + `\` - Clear formatting
- `CMD` + `.` - Show/hide UI

## Naming is hard

I still don't love the name, "Show/Hide UI". When you represent it as a state, it makes no sense which state you're in. For example look at this:

CleanShot 2023-02-22 at 20 49 38@2x

Which state is that in? Am I showing the UI or hiding it? What am I turning on or off? Showing the UI? Or hiding it? You can't tell from that toggle.

I like calling it "Hide UI" or "Hide interface". This is what Sketch calls it

CleanShot 2023-02-22 at 20 48 13@2x

I like it because it makes it clear when you represent the state, what state you're in:

CleanShot 2023-02-22 at 21 00 33@2x

But also I like that it feels named to match its function, e.g. "Whatever state the UI is in based on my preferences, override that and hide the application interface". The name of the feature goes with the grain of its function. I like that.

CleanShot 2023-02-22 at 21 03 00@2x

(That said, if we do go with "Show/Hide UI" we should keep sentence case consistent, so "Show/hide UI" instead of "Show/Hide UI")

davidkircos:
@jimniels 
Screenshot 2023-02-24 at 11 06 30 AM

Interestingly, Figma does "Show/Hide UI", and all their other settings are sentence case. I like this 🤷 

For shortcuts, Command + / makes sense for Show/Hide UI, and Command + \ makes sense for clear formatting. That remains consistent with Sheets

davidkircos:
@jimniels I updated all language to refer to it as "Presentation Mode" let me know what you think

src/gridGL/interaction/keyboard/keyboardViewport.ts
```diff
@@ -27,12 +29,13 @@ export function keyboardViewport(options: {
...
     setEditorInteractionState,
-    // pointer,
+    presentationMode,
+    setPresentationMode,
   } = options;
 
   if (!viewport || event.altKey) return false;
 
-  if ((event.metaKey || event.ctrlKey) && event.code === 'KeyP') {
+  if ((event.metaKey || event.ctrlKey) && (event.code === 'KeyP' || event.code === 'KeyK' || event.code === 'Slash')) {
```
jimniels, line 38: Niiiice.

jimniels:
Trying an additional enhancement on this, see https://github.com/quadratichq/quadratic/pull/326

davidkircos (approved):

PR #210, opened 2023-02-03
feat: support dragging when space is being pressed

jimniels:
Resolves #203

Issue #135, opened 2022-10-10
Scroll Bars


davidkircos:
"One suggestion is just purely UI, it’s hard to pan… i.e. move around the spreadsheet at least on my browser (google chrome). I’m not sure what solution is best, but a simple scrollbar could work, or using arrow keys, left, right, up down."

Suggestion from user feedback.

davidfig:
Maybe add a spacebar (or a modifier key) + mouse drag to move the spreadsheet similar to Figma works?

davidkircos:
@davidfig Is there an existing control for this in `pixi-viewport`? Something we could very easily just turn on?

davidfig:
drag.options.keyToPress was added by a contributor. I haven't used it yet, but it's worth trying. If it doesn't work, it's pretty easy to implement.

deshumake:
Are there any plans for the right mouse button? Clicking and dragging with it could be nifty

Scrolling works really well with a touchpad btw

davidkircos:
@deshumake right click plan is a context menu. Holding a key + click and drag would be great as discussed above.

deshumake:
I was meaning dragging the right click, or is that going to be the context menu you were talking about? If I remember correctly, I believe that's what unity uses to pan, and I thought it was pretty intuitive

davidkircos:
Ah. Right click for context menu. Right click and drag for movement.

Seems worth trying out!

PR #212, opened 2023-02-03
feat: support dragging when space is being pressed

### LINKED ISSUE #203: Implement spacebar for panning
Based on [research around navigating infinite canvas apps](https://www.notion.so/Navigation-Infinite-Canvas-Tools-23789227181b419e9a72444308343ccf), a common paradigm is to allow the spacebar as a shortcut to the panning tool (spacebar + click/drag to move around).

Expectations:

- Pressing `Space` should short-circuit input to a cell (e.g. space can be entered in a cell, but not as the first character)
- Holding down `Space` changes the CSS cursor to `grab` and allows users to click and drag
- Holding down `Space` AND click/drag changes the CSS cursor to `grabbing` and moves the canvas around
- Letting go of `Space` exits user from the panning tool

Feb-01-2023 16-06-01

jimniels:
Mimic Figma in how you can pan around the canvas

- (Re-applies) changes from #210
- Resolves #203 
- Reference: `keyToPress` for drag options https://github.com/quadratichq/quadratic/issues/135#issuecomment-1274955019
- Standardize usage of `react-device-detect` package. Device-specific flags for things like saving space (e.g. `isMobile`) can be used where necessary. Otherwise, the `isMobile` flag (which tests true for mobile and tablet) is used for the `IS_READONLY_MODE` constant that controls when the app is in "readonly" mode.

Feb-15-2023 17-13-22

How it works:

CleanShot 2023-02-24 at 18 25 14@2x

Todos:

- [x] Holding down spacebar switches cursor to `grab`
- [x] Holding down spacebar and mouse switches cursor to `grabbing` and allow panning
- [x] Letting go of spacebar/mouse gets you out of "panning mode"
- [x] If you're in "panning mode", taking the mouse outside of the grid keeps you in that mode
- [x] Add enums for pan mode


davidkircos:
Do you have to set this flag to `true`? 

https://github.com/quadratichq/quadratic/pull/212/files#diff-e3096530de99986094fddf045d3c58ac41adaf9ac648dae7e41d8b05822867c8R74

jimniels:
Ah yes, duh. That does make dragging work, but now there are other issues ha. For example:

1. When you hold down spacebar and click to drag, the cell you click to drag gets selected (which we would want to prevent).
2. Dragging outside the browser window changes from panning to selecting. I would expect it to continue to be panning.
3. We would want to toggle the cursor when panning, e.g. when spacebar gets pressed down we would want to change the cursor to `grab` and then when we're dragging we would want to change it to `grabbing`.

These seems like controls or aspects of state that would need to come from pixi (??) and I'm not sure how to do them. Might have to leave this one until I can get help on it.

davidkircos:
Yeah, I think this will be a bit tricky.

Probably best to put on the back burner for now.

David

src/gridGL/QuadraticGrid.tsx
```diff
@@ -51,7 +62,14 @@ export default function QuadraticGrid(props: IProps) {
   // Right click menu
   const [showContextMenu, setShowContextMenu] = useState(false);
 
-  const { onKeyDown } = useKeyboard({
+  // Pan mode
+  const { onMouseDown, onMouseUp, onKeyDown, onKeyUp } = usePanMode({
+    panMode: interactionState.panMode,
+    interactionState,
+    setInteractionState,
+  });
```
jimniels, line 70: I'm not sure if this is the best way to write this?

It feels cleaner, but does it mean that the event handlers (which need access to state  in this component) don't get to use a closure for access the values and it causes unnecessary regeneration of the event handlers themselves? 

React hooks trip me up here. But it does seem to work. I like the encapsulated logic but unsure if it's not idiomatic react.

src/gridGL/pixiApp/PixiApp.ts
```diff
@@ -73,7 +72,10 @@ export class PixiApp {
     this.viewport = new Viewport({ interaction: this.renderer.plugins.interaction });
     this.stage.addChild(this.viewport);
     this.viewport
-      .drag({ pressDrag: isMobileOnly }) // enable drag on mobile, no where else
+      .drag({
+        pressDrag: true, // enable drag on mobile, no where else
```
davidkircos, line 76: this comment is no longer valid

davidkircos:
@jimniels drag no longer works on mobile

davidkircos:
@jimniels drag no longer works on mobile

jimniels:
@davidkircos ah good catch. I needed to [conditionally pass in the spacebar key for dragging](https://github.com/quadratichq/quadratic/pull/212/commits/d79db1c3e0b6f331e498a4034697d6114dd49358#diff-07dfae82d5a245c6f7a92c5b147bca8506b4cc9938ffcd2c5e634ae8c73cd4aeR78) Should work now.

davidkircos:
When you are typing into a cell, space still activates drag mode. 

I think we just need to preventDefault on space here? 
https://github.com/quadratichq/quadratic/blob/main/src/gridGL/interaction/CellInput.tsx#L193

jimniels:
@davidkircos good catch on the cell input. Should've caught that. It's fixed now and ready for review once more

PR #214, opened 2023-02-03
feat: context menu

jimniels:
Formatting menu _is_ the context menu.

- Adds "Cut", "Copy", "Paste" to the formatting toolbar
- Adds tooltips (with shortcuts) to the toolbar
- Makes "right click" trigger the formatting toolbar
- Adds support for "clear formatting" keyboard shortcut

CleanShot 2023-02-03 at 16 33 29@2x


src/core/gridGL/interaction/keyboard/keyboardViewport.ts
```diff
@@ -21,6 +22,10 @@ export function keyboardViewport(options: {
     });
     return true;
   }
+  if ((event.metaKey || event.ctrlKey) && event.code === 'Backslash') {
```
jimniels, line 25: The shortcut for clear formatting in google sheets is `CMD` + `\`, which is what I used here.

But we also talked about `CMD` + `\` for "show/hide UI" [which hasn't been implemented yet but we've explored](https://github.com/quadratichq/quadratic/pull/208)

src/core/gridGL/interaction/keyboard/useKeyboard.ts
```diff
@@ -31,6 +33,13 @@ export const useKeyboard = (props: IProps): { onKeyDown: (event: React.KeyboardE
     app,
     sheetController,
   } = props;
+  const { clearFormatting } = useFormatCells(sheetController, app);
+  // @ts-expect-error
+  const { clearBorders } = useBorders(sheetController.sheet, app);
```
jimniels, line 38: Typescript fails on the app argument here because it says it _can be_ undefined, so it won't accept the prop being passed...

jimniels: If i remove [the conditional aspect of `app`](https://github.com/quadratichq/quadratic/pull/214/files#diff-f7aea0d262a79bcb3510bce5a571eaf6223bd82a01ff8065819b88f7e39555ffL19), then I won't get an error here, but i don't know if/why app _might_ be conditional in this case...?

e.g. change this

```
app?: PixiApp;
```

to this:

```
app: PixiApp
```

And everything works? But the parent component `<QuadraticGrid>` that passes down `app` has a lot of `useEffect` calls in it where it conditionally checks for the presence of `props.app?.thing` so I'm unsure of the mechanics here... cc @davidkircos

davidkircos:
Font Color, Bold, and Italic are missing tooltips.

davidkircos (approved):

src/core/gridGL/interaction/keyboard/useKeyboard.ts
davidkircos: If you made it non optional and typescript didn't complain that it would sometime be `undefined` it is probably not possible for it to be undefined. So should be good.

PR #223, opened 2023-02-07
move saveLastLocal to end of transaction, fix bug with formatting not…

davidkircos:
… redoing bold, italic, etc

### DIFF
--- a/src/core/transaction/runners/setCellFormatRunner.ts
+++ b/src/core/transaction/runners/setCellFormatRunner.ts
@@ -2,15 +2,10 @@ import { Sheet } from '../../gridDB/Sheet';
 import { Statement } from '../statement';
 import { CellFormat } from '../../gridDB/gridTypes';
 import { PixiApp } from '../../gridGL/pixiApp/PixiApp';
-import { localFiles } from '../../gridDB/localFiles';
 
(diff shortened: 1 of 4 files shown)


src/core/transaction/runners/setCellFormatRunner.ts
```diff
@@ -2,15 +2,10 @@ import { Sheet } from '../../gridDB/Sheet';
...
-import { localFiles } from '../../gridDB/localFiles';
 
 const CopyCellFormat = (format: CellFormat | undefined): CellFormat | undefined => {
   if (format === undefined) return undefined;
-  return {
-    x: format.x,
-    y: format.y,
-    fillColor: format.fillColor,
-  };
+  return { ...format };
```
davidkircos, line 8: Simple bug, we need to copy the whole object.

src/core/transaction/sheetController.ts
```diff
@@ -73,6 +74,7 @@ export class SheetController {
 
     // TODO: This is a good place to do things like mark Quadrants as dirty, save the file, etc.
     // TODO: The transaction should keep track of everything that becomes dirty while executing and then just sets the correct flags on app.
+    localFiles.saveLastLocal(this.sheet.export_file());
```
davidkircos, line 77: only call saveLastLocal once at the end of the transaction

Issue #220, opened 2023-02-06
Indicate active formatting state


jimniels:
Similar to Notion, it would be a nice-to-have to indicate the current state in the formatting toolbar:

<img width="656" alt="CleanShot 2023-02-06 at 13 54 48@2x" src="https://user-images.githubusercontent.com/1316441/217083471-ce2af904-d83a-4be2-bd09-a8ff43739ac5.png">

Example:

<img width="1049" alt="CleanShot 2023-02-06 at 13 58 23@2x" src="https://user-images.githubusercontent.com/1316441/217084103-091af782-78ae-4e38-9dbb-65f315aca6cd.png">

But we also need to holistically consider all the places where we allow formatting and whether we want to communicate state (on/off/indeterminate) there too.

There's the formatting in the toolbar:

<img width="266" alt="CleanShot 2023-02-07 at 10 49 20@2x" src="https://user-images.githubusercontent.com/1316441/217324741-9813c0ff-b604-4691-adb8-c1f0cd5f8e23.png">

As well as in the command palette:

<img width="555" alt="CleanShot 2023-02-07 at 10 49 37@2x" src="https://user-images.githubusercontent.com/1316441/217324782-b727b160-6fb3-4a08-9b31-e70da273de1c.png">

davidkircos:
very low priority

PR #229, opened 2023-02-07
fix: menus

jimniels:
- Add clear all formatting to the command palette (with shortcut)

CleanShot 2023-02-07 at 15 46 35@2x

- Removes stateful nature of bold/italic menus (we can bring this back one day, see #220)
- Removes (currently disabled) text alignment and text wrapping from formatting menu
- Fixes menu alignment/padding

CleanShot 2023-02-07 at 15 48 27@2x

- Fix so there's parity in ordering and grouping between format toolbar menu and format context menu

CleanShot 2023-02-07 at 16 04 33@2x

- Tries to resolve weirdness of top-level tooltips (tooltips could be interacted with, so it would show up even when your cursor went to the menu)
  - Tried to fix this the best I could with the fact that we're trying to make two different UI libraries (`mui` and `react-menu`) work together. It could be better, but I don't think it's worth spending the time on right now.

Feb-07-2023 13-58-35


davidkircos (approved):
:ship:

PR #245, opened 2023-02-10
fix: update floating formatting bar

jimniels:
- Re-style the dividers in the formatting menu to contrast with the vertical column lines (so they don't blend in with each other) by 1) not letting them run to the edges of the widget, and 2) run lines perpendicular to column line
- Fix the 'text color' dropdown to not have box shadows (must've lost this change in a merge somewhere?)

image

### DIFF
--- a/src/ui/menus/ContextMenu/FloatingContextMenu.tsx
+++ b/src/ui/menus/ContextMenu/FloatingContextMenu.tsx
@@ -371,9 +371,7 @@ function MenuDivider() {
       flexItem
       style={{
         // Style these to stand in contrast against with the vertical column lines
-        margin: '10px 6px',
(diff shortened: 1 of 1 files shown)


src/ui/menus/ContextMenu/FloatingContextMenu.tsx
```diff
@@ -252,6 +252,7 @@ export const FloatingContextMenu = (props: Props) => {
           </IconButton>
         </TooltipHint>
         <Menu
+          className="color-picker-submenu"
```
jimniels, line 255: This must've gotten lost somewhere, but this removes that box shadow from appearing

Screenshot 2023-02-10 at 12 30 54 PM

PR #248, opened 2023-02-10
fix double undo bug for clearing all formatting

davidkircos:
Bug:
1. Add some borders and formatting
2. Clear formatting
3. Press Undo

Previously it would be two Undos to get back, one for borders one for formatting. now it is one.


jimniels:
BOOM 💥

LGTM 👍

PR #251, opened 2023-02-10
Command Palette: additional icons / css fix 

waveminded:
adding more icons to the command palette for consistency, and CSS fix to remove bottom scrollbar empty space on Command Palette


davidkircos:
@jimniels could you review this PR?

src/ui/menus/CommandPalette/ListItems/File.tsx
```diff
@@ -2,13 +2,15 @@ import { CommandPaletteListItemSharedProps } from '../CommandPaletteListItem';
 import { CommandPaletteListItem } from '../CommandPaletteListItem';
 import { newGridFile, openGridFile } from '../../../../core/actions/gridFile/OpenGridFile';
 import { SaveGridFile } from '../../../../core/actions/gridFile/SaveGridFile';
+import { UnarchiveOutlined, NoteAddOutlined, SaveAlt } from '@mui/icons-material';
```
jimniels, line 5: I'm thinking about these icons a little more...

image

Here's my thoughts, tell me what you think:

I don't love how different these icons are from each other. They're all conceptually about "file" but their icons are all so different. Even the arrows on the icons are differently-shaped arrows

CleanShot 2023-02-13 at 11 01 59@2x

I feel like using the outline of a "file" for each icon would be the most intuitive. It doesn't look like the default material UI has three different "file" icons representing new, save, and open. If we go with just what they have, I would think [FileOpenOutlined](https://mui.com/material-ui/material-icons/?query=file&theme=Outlined&selected=FileOpenOutlined) might be best for open. As for save, maybe the traditional floppy disk [SaveOutlined](https://mui.com/material-ui/material-icons/?query=save&theme=Outlined&selected=SaveOutlined) would be best?

image

Even that, I still don't love. Again, I feel like the notion of a "document" icon that is repeated for each would be the most useful. Maybe we could create our own "save" icon that is the same as the existing "FileOpenOutlined" but we reverse the direction of the arrow, e.g.

image

Part of me feels like a blank document might be even better for "new", e.g.

image

That last one is the one I like the best. What do you think?

jimniels: Happy to show you how to create that file too...it's pretty straightforward.

jimniels (commented):
I like the iterative direction towards improvements here. Have a couple thoughts. Let me know what you think...

src/ui/menus/CommandPalette/ListItems/File.tsx
waveminded: I dig the last iteration for now. I know the MUI icons aren't the greatest. Which file are you referring? Just manipulating the svg of the icon?

jimniels: Yeah, the "new file" would just be the same icon. "open file" would be the [UploadFile](https://mui.com/material-ui/material-icons/?query=file&selected=UploadFile) icon, and the "save file" would be a custom one we create from the UploadFile icon where we just invert the direction of the arrow.

jimniels (approved):
LGTM! 👍

Issue #236, opened 2023-02-08
Link to Documentation from Python + Formulas Code Editor


davidkircos:
<img width="927" alt="Screenshot 2023-02-08 at 1 07 08 PM" src="https://user-images.githubusercontent.com/3479421/217639748-c1c36e40-e9a7-4e02-8aae-7dfaadbb2673.png">

Link to Python Getting Started, and to be created Formulas documentation.

PR #263, opened 2023-02-14
feat: code editor console improvements

jimniels:
Generally enhances and re-styles the "console" in the editor.  Resolves part of #236 (still need to create formulas docs)

- Errors and `print()` log to the "Output" tab (errors are elevated)

CleanShot 2023-02-13 at 23 11 13@2x

- Language-specific context lives in the "About" tab

CleanShot 2023-02-13 at 23 12 59@2x

- Refactor language in the in the command palette to be shorter

CleanShot 2023-02-13 at 14 29 02@2x

- Console can be resized vertically

Feb-13-2023 23-16-31


src/constants/urls.ts
```diff
@@ -1,3 +1,5 @@
 export const API_URL = 'http://localhost:8000';
 export const DOCUMENTATION_URL = 'https://docs.quadratichq.com';
+export const DOCUMENTATION_PYTHON_URL = `${DOCUMENTATION_URL}/reference/python-cell-reference`;
+export const DOCUMENTATION_FORMULAS_URL = `${DOCUMENTATION_URL}`; // TODO
```
jimniels, line 4: Right now this just points to the docs home page as we don't have a formulas-specific page

davidkircos:
All these changes look great to me. My one comment is changing the OUTPUT to be an input field so the user can select all inside of it and have a cursor be the default value, like VSCode.

Screenshot 2023-02-14 at 8 23 42 AM

jimniels:
@davidkircos this is actually what it's doing now (maybe it wasn't for you?)

I didn't use an `input` or `textarea` because those don't allow you to style the text inside of them (and I think styling the error text is a nice touch)

CleanShot 2023-02-14 at 10 19 57@2x

It makes it feel more like VSCode which does the same:

CleanShot 2023-02-14 at 10 20 56@2x

So to achieve the same effect, [I use `contenteditable` and prevent editing the text](https://github.com/quadratichq/quadratic/pull/263/files#diff-2bc202dc21e9b27b3ad74739a0edda788d6d4da097d558b077a50d6fcc23276eR352-R363) this allows us to style the text of the output while still allowing the user to do `CMD` + `A` to select all text inside

Feb-14-2023 10-23-22

Also: this kind of implementation allows for a cursor to show up in the output box _when the user focuses it_

Feb-14-2023 10-24-38

This is similar to vscode, which won't show the cursor in the "output" tab unless the tab comes into focus by the user

Feb-14-2023 10-26-12

jimniels:
@davidkircos maybe take one more look, as I refined the styles on the resize controls.

First, the resize control was a 1px line aligned to _one side_ of the target box.

CleanShot 2023-02-14 at 12 02 46@2x

Now it is a 1px line centered within with the target area

CleanShot 2023-02-14 at 11 59 47@2x

This makes the hover interactions feel more natural when you try to resize, as the target area is symmetrically balanced vs. asymmetrically balanced:

Feb-14-2023 11-46-13

Also: the resize target area now lights up when you hover over the area _with intent_ (e.g. slight pause over the area). VSCode has this when resizing:

Feb-14-2023 11-46-39

Which is a nice effect and now quadratic has it:

Feb-14-2023 11-46-34

All the above required CSS to achieve the effect, which is why you'll see it implemented as a CSS solution instead of inline styles via JS.

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -196,164 +200,287 @@ export const CodeEditor = (props: CodeEditorProps) => {
...
+              }
+            }}
+            onCut={preventDefault}
+            onPaste={preventDefault}
+            style={{ outline: 'none' }}
+          >
+            {hasOutput && (
+              <>
+                {std_err && (
+                  <span style={{ display: 'flex', alignItems: 'center', gap: '8px', color: red[700] }}>
```
davidkircos, line 369: use `colors.error`

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -196,164 +200,287 @@ export const CodeEditor = (props: CodeEditorProps) => {
...
+          <Console evalResult={evalResult} editorMode={editorMode} />
         )}
       </div>
-    );
-  return <></>;
+    </div>
+  );
 };
 
+function Console({ evalResult, editorMode }: { evalResult: cellEvaluationReturnType | undefined; editorMode: string }) {
```
davidkircos, line 327: move to a separate file

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -196,164 +200,287 @@ export const CodeEditor = (props: CodeEditorProps) => {
...
+              </p>
+            </>
+          )}
+        </TabPanel>
+      </div>
+    </>
+  );
+}
+
+function LinkNewTab({ href, children }: { href: string; children: React.ReactNode }) {
```
davidkircos, line 408: Move to a new file. Can share a file with TabPanel

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -196,164 +200,287 @@ export const CodeEditor = (props: CodeEditorProps) => {
...
+          }}
+        >
+          {children}
+        </pre>
+      )}
+    </div>
+  );
+}
+
+function ResizeControl({ setState, position }: { setState: Function; position: 'TOP' | 'LEFT' }) {
```
davidkircos, line 444: Move to new file

davidkircos (changes requested):

src/ui/menus/CodeEditor/CodeEditor.tsx
jimniels: Actually moved this to a new file and shared it with one very similar already used in the command palette, so now it works in both cases.

davidkircos (approved):

PR #276, opened 2023-02-16
Fix copy paste xml

davidkircos:


jimniels (approved):
Copy/paste with html entities looks like it's working 👍 

Also great fix on the keyboard navigation copy/paste

PR #277, opened 2023-02-16
Improve Cell Type Outlines

davidkircos:
- [x] Remove cell-type outlines from plain text
- [x] Remove cell-type outlines from computed cell
- [x] Improve array output outlines to match type

Screenshot 2023-02-16 at 11 06 39 AM


jimniels (approved):
This is all i've ever wanted. 🎅  came early this year!

PR #279, opened 2023-02-16
feat: style & rename cell type outlines toggle

jimniels:
Style the toggle in the upper right to make it hint more at its function (rather than a generic, label-less toggle).

Also, once #277 gets merged, the function of this toggle is less about "cell type" outlines and more about the outlines behind a specific kind of cell, e.g. code/formulas, so my suggestion is we actually rename it to be "Show code cell outlines".

- [x] New toggle in app bar with tooltip
- [x] New name in tooltip, "Show code cell outlines"
- [x] New name in header menu bar
- [x] New name in command palette 

Mar-02-2023 11-04-21

Labels updated to "Show code cell outlines" in respective parts of app:

CleanShot 2023-02-28 at 15 18 32@2x

CleanShot 2023-02-28 at 15 18 19@2x

---

### History

First iteration tried an icon that toggled its appearance:

Feb-16-2023 13-08-39

Design notes:

I tried a couple different design ideas for this as a toggle

CleanShot 2023-02-16 at 13 17 31@2x

But ultimately felt that an icon in the same size/shape as the others in the toolbar was most effective.

For the on/off mode, rather than a fill or color change, I went with a dashed line. I tried a version where we color the "on" state, but it didn't quite feel appropriate and in line with how we use that color elsewhere in the app (which is a primary color and not necessarily a stateful color — though it is that way on the switches). I'm on the fence on this, but the current code keeps the icon the same color regardless of state (and only changes the shape).

Feb-16-2023 13-12-36


davidkircos:
@jimniels My thought for this one is lets go ahead with the rename and skip the Icon change. Keep the switch (for now).

src/ui/menus/TopBar/TopBar.tsx
```diff
@@ -24,6 +25,64 @@ interface IProps {
   sheetController: SheetController;
 }
 
+const CustomSwitch = styled(Switch)(({ theme }) => ({
```
davidkircos, line 28: Move this component to a different file.

src/ui/menus/TopBar/TopBar.tsx
```diff
@@ -24,6 +25,64 @@ interface IProps {
...
+      content: '""',
+      position: 'absolute',
+      top: '50%',
+      transform: 'translateY(-50%)',
+      width: 18,
+      height: 18,
+      backgroundSize: '18px',
+    },
+    '&:before': {
+      backgroundImage: `url('data:image/svg+xml;utf8,<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill="white" d="M6.2998 13.5L1.7998 9L6.2998 4.5L7.25605 5.45625L3.7123 9L7.25605 12.5438L6.2998 13.5ZM11.6998 13.5L10.7436 12.5438L14.2873 9L10.7436 5.45625L11.6998 4.5L16.1998 9L11.6998 13.5Z" /></svg>')`,
```
davidkircos, line 46: It seems like the icons could be a little bit smaller.

Also with the primary blue the switch really draws attention. Could you maybe do an outlined button with the icon?

src/ui/menus/TopBar/TopBar.tsx
davidkircos: Generally, I think the switch design is excellent! Nice job with the icon.

davidkircos (changes requested):

src/ui/menus/TopBar/TopBar.tsx
jimniels: Just pushed up a change that inverts the color so it's not so attention-grabbing. Here's what it looks like, what do you think?

Mar-01-2023 12-24-29

jimniels: Latest:
Mar-02-2023 11-04-21

davidkircos (approved):

PR #280, opened 2023-02-17
Update Example Files

davidkircos:


jimniels (changes requested):
Examples look good and are a great iteration over what we previously had 👍 

Only thing I noticed: I would suggest we use consistent casing across the files. Right now it's kind of mixed. We should stick to "Sentece case" (first capitalized, all subsequent lower unless proper noun, e.g. "Use Python examples from PIP" 

For example, here's the `python.grid` file

CleanShot 2023-02-22 at 06 49 09@2x

And here's `default.grid`

CleanShot 2023-02-22 at 06 53 02@2x

Other than that, these look good to me 👍

davidkircos:
@jimniels I went ahead and made those changes. Ready for review again

jimniels (approved):
Sorry, I didn't include screenshots of the other grids that had casing inconsistencies. I just went ahead and fixed those and pushed. We should be good now 👍

PR #299, opened 2023-02-20
upgrade sentry, add replays

davidkircos:


davidkircos:
This appears to be setup correctly. However, Sentry doesn't recognize the replay session. Needs another look.

PR #307, opened 2023-02-22
Fix Array Output Resize

### LINKED ISSUE #281: bug: Python Doesn't Clear Previous Cells Correctly with DataFrame
DataFrame results are not correctly resized

Feb-16-2023 17-26-21

davidkircos:
Fixes #281


jimniels (commented):
This looks good to me.

One thing, though i'm not sure if it's related to these changes (or perhaps [the updated example files PR](https://github.com/quadratichq/quadratic/pull/280))?

When I go "File > Open sample > Expenses" and go to change a value, it doesn't redraw at all:

Feb-22-2023 07-01-16

However, if i go in and edit the code that draws that table, save and run it, then change the filter value, now it works as expected.

Feb-23-2023 11-32-07

I still don't totally understand how cell dependencies work, but seems like the dependency array in the sample files have to be updated for this change to take effect?

davidkircos:
@jimniels I think that is an example files issue. I'll merge the examples PR first and check if it fixes it in this branch.

jimniels (approved):
Yup, that seemed to fix it 👍

PR #308, opened 2023-02-23
feat: File Browser Update (urls, examples, etc)

### LINKED ISSUE #323: feat: working in multiple tabs/windows
**Problem**

Today in Quadratic, if you have multiple tabs open, the experience can be a bit befuddling and there's a possibility for lost data.

There are two specific use cases I want to call out:

1. There is no notion of "multiple files" open at the same time across different tabs. If I go to `app.quadratichq.com` and open `python.grid`, then I open a new tab of `app.quadratichq.com` it'll open _the last saved state_ of `python.grid`. In that tab, if I open a new file, say `opan_ai.grid` and then go back to the first tab (which has `python.grid`) and I hit refresh, that new tab doesn't reload the `python.grid` file I had open when I hit refresh. It opens _whatever file was last opened on that origin_, in this case it will open `open_ai.grid` even though I was on the `python.grid` file when I hit refresh.
2. Under the current paradigm (point no. 1 above) changes fall out of sync with each other. If you edit a cell in `python.grid` in tab 1, then switch to tab 2 that also has `python.grid` open, that change you just made is not there (unless you refresh the page). If you then make a different change and go back to tab 1 and refresh, the initial change you made in tab 1 is actually lost and the change you made in tab 2 is now there.

https://user-images.githubusercontent.com/1316441/221667920-4266c76c-4039-4b20-9ac1-0aa02953f714.mp4

**Solution**

1. It seems like each tab should have it's own notion of an "open file" so you can work on multiple files at the same time and, if you hit refresh, you don't lose your changes. In tab 1 I can have `python.grid`, in tab 2 I can have `open_ai.grid` and if I hate "refresh" in either of those tabs, they stay on the file I was working on.
2. For tabs with the same file open (on the same origin), changes would propagage across tabs/windows using the [Broadcast Channel API](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) (I haven't used it but supposedly this is what it does). In this way, you don't run into issues where the same file is edited across multiple tabs (changes sync across open tabs).

Point 2 might need more conversation. We'd have to have a good way to identify files as being "the same file" and that could get tricky. Somebody might be opening `python.grid` in a new tab because they want to branch off it and have it _not_ be the same file that's open in tab 1. So perhaps that one is worth punting on until a "cloud" version of Quadratic.

But I do think point 1 deserves further consideration for improvement now. It seems odd to me that you can't have multiple tabs of quadratic open to work on multiple, different files, at one time without losing your work if you refresh. _One `.grid` file per tab_.

davidkircos:
Resolves #323 

Functionality / things to test:

- Routing:
  - `/` loads the file menu _unless_ it's the first visit, then it loads the default file
  - `/?local={valid-file-id}` loads a file from memory
  - `/?local={invalid-text}` loads the file menu with an error
    - e.g. `/?local=foo`
  - `/?file={valid-url}` loads the `.grid` file over the network
    - Can test using example files from the `v0.3.0-beta` pulled from jsdeliver:
      - [/?file=https://cdn.jsdelivr.net/gh/quadratichq/quadratic@v0.3.0-beta/public/examples/python.grid](https://pr-308.de5w5iglj13on.amplifyapp.com/?file=https://cdn.jsdelivr.net/gh/quadratichq/quadratic@v0.3.0-beta/public/examples/python.grid)
      - [/?file=https://cdn.jsdelivr.net/gh/quadratichq/quadratic@v0.3.0-beta/public/examples/startup_portfolio.grid](https://pr-308.de5w5iglj13on.amplifyapp.com/?file=https://cdn.jsdelivr.net/gh/quadratichq/quadratic@v0.3.0-beta/public/examples/startup_portfolio.grid)
      - [/?file=https://cdn.jsdelivr.net/gh/quadratichq/quadratic@v0.3.0-beta/public/examples/expenses.grid](https://pr-308.de5w5iglj13on.amplifyapp.com/?file=https://cdn.jsdelivr.net/gh/quadratichq/quadratic@v0.3.0-beta/public/examples/expenses.grid)
  - `/?file={invalid-url}` loads the file menu with an error
    - e.g. `/?file=https://google.com`
- File menu:
  - Opens with: `CMD` + `O`, Toolbar "File -> Open...", or Command Palette "File: Open..."
  - Close with:  `ESC` or `X` in UI
  - Create a new file (names it "Untitled", goes to origin, zooms to 100%)
  - Open an example file (names it file name, goes to origin, zooms to 100%)
  - Open a file from local device (names it based on file on disk sans extension, goes to origin, zooms to 100%)
    - **Make sure to test this with an  old file** (something from the current version of production)
  - Open a file from a URL (names based on file path in, goes to origin, zooms to 100%)
    - Test with a valid `.grid` URL: https://cdn.jim-nielsen.com/files/sample-file.grid
    - Test with a valid `.grid` URL **from an old file**
    - Test with an invalid URL, e.g. https://google.com
  - Delete an in-memory file (deletes it from UI and indexDB)
  - Download a local copy of an in-memory file
- "File -> Save local copy" (from toolbar or command palette) downloads file with current name + `.grid` extension
- "File -> New" (from toolbar or command palette) starts a new file, zoomed to 100% at (0,0)
- In-app link to "Files" page in docs works
- Can rename a file from the top bar (empty filenames or filenames with only spaces `   ` are not allowed)
- "Open file" menu accessible from tablet/mobile
- Files you currently have in memory are migrated to our new storage mechanism
  - You can test this locally by running `main` on `localhost:3000` and create a couple different files in memory, then checkout this branch and load the app and the files you created should be accessible in the file menu

Todos:
- [x] Current files in memory should survive this new change when it merges
- [x] Write docs for how files/file formats work in Quadratic (code currently points at [docs.quadratichq.com/files](https://docs.quadratichq.com/files))
- [x] Fix contextual TODOs in code
- [x] Firing twice on each save (appears to be something with pixiapp, not new code)
- [x] Support old files
- [ ] Fix/update sample files
  - [ ] Update with version "1.1" meta info?
  - [x] Some aren't loading right now?
  - [x] Updates from `main`?
- [x] Add filename to document.title `{File Name} - Quadratic`
- [x] File issue for after merge: open separate PR to use context on `app` and `sheetConroller`

Nice to haves:
- [x] Example files that, when you clone them, appear with the same name as in the example file list
- [x] Weird flicker when sheet loads first, then open menu renders
  - Coming from how the app loads. `useLocalFiles` hook has an async action to load the file from memory. Would require re-architecting how this works...

Leaving out:

- Ability to deep link to specific tabs of the "Open file" menu, e.g. from the header click "File -> New from URL" which would open this new screen with the "Remote" URL tab selected
- Contexts for `sheetController` and `app` https://github.com/quadratichq/quadratic/issues/387


package.json
```diff
@@ -50,7 +51,9 @@
     "recoil": "^0.6.0",
     "sass": "^1.56.2",
     "typescript": "^4.4.4",
-    "web-vitals": "^1.0.1"
+    "uuid": "^9.0.0",
```
jimniels, line 54: I wonder if we need this dependency....

[It looks like](https://stackoverflow.com/a/2117523/1339693) you can create IDs with [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) which is a native browser feature. Support appears to be [good enough](https://caniuse.com/mdn-api_crypto_randomuuid) (better than probably other things in our app)

So I'm thinking: maybe we just strip out this dependency and use `crypto.randomUUID()` instead. Objections?

cc @davidfig @davidkircos

src/grid/actions/updateCellAndDCells.ts
```diff
@@ -120,12 +121,12 @@ export const updateCellAndDCells = async (args: ArgsType) => {
...
         // if array output
         if (result.array_output !== undefined && result.array_output.length > 0) {
-          if (result.array_output[0][0] !== undefined && typeof result.array_output[0] !== 'string') {
+          if (Array.isArray(result.array_output[0])) {
             // 2d array
             let y_offset = 0;
             for (const row of result.array_output) {
               let x_offset = 0;
-              for (const cell of row) {
+              for (const cell of row as ArrayOutput) {
```
jimniels, line 129: [This change in the `array_output` schema](https://github.com/quadratichq/quadratic/pull/308/files#diff-786b8eb1d8f122847835f3faa2451f8d2e9ca3401a219c1f6dfca5c02d49db5aR11) necessitated how we check for this type because Typescript didn't like it.

CleanShot 2023-03-29 at 13 26 14@2x

src/grid/computations/types.ts
```diff
@@ -1,10 +1,17 @@
...
+
+const ArrayOutputSchema = z.array(z.union([z.string(), z.number(), z.boolean()]));
+
+export const CellEvaluationResultSchema = z.object({
+  success: z.boolean(),
+  std_out: z.string().optional(),
+  std_err: z.string().optional(),
+  output_value: z.string().or(z.null()).or(z.undefined()),
+  cells_accessed: z.tuple([z.number(), z.number()]).array(),
+  array_output: z.union([ArrayOutputSchema, z.array(ArrayOutputSchema)]).optional(), // 1 or 2d array
```
jimniels, line 11: As you can see in the diff, previously this was a 2d array:

`(string|boolean|number)[][]`

But now it is:

`(string|boolean|number)[][]`
`(string|boolean|number)[]`
`undefined`

Without this change, runtime validation of our previous example files were failing because `array_output` in many cases was a 2d array OR a 1d OR undefined.

src/grid/sheet/Sheet.ts
```diff
@@ -52,17 +52,16 @@ export class Sheet {
...
     const { cells, formats } = this.grid.getArrays();
     return {
       columns: this.gridOffsets.getColumnsArray(),
       rows: this.gridOffsets.getRowsArray(),
       cells,
       formats,
       borders: this.borders.getArray(),
-      render_dependency: this.render_dependency.save(),
       cell_dependency: this.cell_dependency.exportToString(),
-      version: '1.0',
```
jimniels, line 65: `version` gets removed from here because it's file meta info and that is extracted in [the function that downloads the file](https://github.com/quadratichq/quadratic/pull/308/files#diff-ff625f246fa2b9b5d0cb19ad7fdcf3d7cc4c44ce1f92fe11d60b7809f60d6309R191)

davidkircos:
When processing old files to new the order is reversed: (not a big deal)
CleanShot 2023-03-31 at 20 00 04@2x
screencapture-localhost-3000-2023-03-31-20_06_02

Have you seen this "SCROLLABLE"? thing in the console before?
CleanShot 2023-03-31 at 20 11 06@2x

What do you think about chaing the Quadratic menu to have a "back to files" option at the top the way Figma does?
CleanShot 2023-03-31 at 20 11 59@2x
In Sheets the top left icon takes to back to your files list
CleanShot 2023-03-31 at 20 13 44@2x

Still testing file actions, but haven't noticed any problems so far. @jimniels

davidkircos:
Visual overflow on mobile image

HactarCE:
Ctrl + O on Windows opens the browser popup. I'm not sure if it's possible for us to override that.

HactarCE:
Also, `?local=...` on Firefox always opens the file menu with no error.

Somehow by refreshing with the right timing using `?local=<invalid>` on Chrome, I managed to completely erase all the local files (!!!) but I can't reproduce this. Not sure how that happened.

jimniels:
@HactarCE 

> Ctrl + O on Windows opens the browser popup. I'm not sure if it's possible for us to override that.

It should be possible, we're overriding it in the other browsers. Not sure why [this code](https://github.com/quadratichq/quadratic/pull/308/files#diff-6adf56e5497f99c762b3b7574a7aa4fca83f6907cb947ea22e878a2fc1a47e91R42) isn't working on windows but it is on Mac.

However, that said, `CMD` + `O` was not implemented as a toggle. It was one-way, e.g. only for opening the "File menu" not closing it. I've tweaked that. Now `CMD` + `O` should toggle the file menu's visibility, e.g. opens it if it's not visible and closes it if it's visible _except_ in the case where you don't have an actively open file in the app in which case it does nothing (because there's nothing to close and go to).

> Also, ?local=... on Firefox always opens the file menu with no error.

Hm...I can't reproduce this on Firefox for Mac.

Loading a file using `?local=<valid-id>` loads the file:

CleanShot 2023-04-03 at 12 13 37@2x

Whereas loading a file using `?local=<invalid-id>` (e.g. `/?local=abcd`) can't load the file and opens the file menu with an error.

CleanShot 2023-04-03 at 12 14 06@2x


> Somehow by refreshing with the right timing using ?local=<invalid> on Chrome, I managed to completely erase all the local files (!!!) but I can't reproduce this. Not sure how that happened.

This one makes me nervous, but I haven't been able to reproduce it. 

I did add [a check in the code](https://github.com/quadratichq/quadratic/pull/308/commits/242a019de9ba778a1ce5ec45689512aa276a6fa3#diff-ff625f246fa2b9b5d0cb19ad7fdcf3d7cc4c44ce1f92fe11d60b7809f60d6309R359-R362) that will only delete stuff when updating old files if it successfully imports it.

Again, i'm not sure how this was triggered. Delete actions are only accessible in the code _when triggered by a user action **and** confirmation_ via the UI. The other deletes happen only on migration files. So i'm not sure how this even happened...?

> It would be nice if dragging a .grid file in loaded it instead of trying to parse a CSV

Agree. I was actually wanting to do that while I was working on this branch. But figured we'd push that for a later feature. I [opened an issue for it](https://github.com/quadratichq/quadratic/issues/397).

jimniels:
@davidkircos [I added a check to the code](https://github.com/quadratichq/quadratic/pull/308/commits/242a019de9ba778a1ce5ec45689512aa276a6fa3#diff-ff625f246fa2b9b5d0cb19ad7fdcf3d7cc4c44ce1f92fe11d60b7809f60d6309R359-R362) that makes sure we only delete a file in memory when we are sure it's been imported/migrated to this new way of saving/storing files.

However, it is worth noting that if a file in memory fails to migrate/import, that fails silently to the user (there's an error logged to the console). So, right now, that file just stays there in localstorage but is inaccessible to the user via the application UI. We don't tell them one of their files failed to import and there's no UI for trying to retrieve it somehow. This is less than ideal but i'm not really sure how much it happens? Maybe we should log this as a specific event in sentry just to see if it happens a lot?

jimniels (approved):
Gonna go ahead and approve this PR

CleanShot 2023-04-03 at 16 45 33@2x

PR #310, opened 2023-02-23
Drag and Drop CSV Support

davidkircos:
Allows you to drag and drop a CSV file into Quadratic. It will populate where you drop it.

- [x] Shows what cell you'll drop a file into
- [x] Populates with CSV
- [x] Show error if file can't be imported
- [x] Help menu showing how to upload CSV
- [x] Add to documentation


davidkircos:
Alternate idea:

What if CSV import worked like a cell ouput? 

Press = one of the options was CSV, and you can upload one in the side panel and it would expand.
This way, you could replace the source file in the cell easily using the side editor.

Alternatively, you could drag and drop a file, and it would create a cell type of CSV. so we could keep this easy import functionality.

davidkircos:
Switch to https://www.npmjs.com/package/xlsx ?

Also, support all other types of file imports and raw export data?

src/ui/components/FileUploadWrapper.tsx
```diff
@@ -0,0 +1,117 @@
...
+  // triggers when file is dropped
+  const handleDrop = async (e: DragEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
+    setDragActive(false);
+
+    if (e.dataTransfer.files && e.dataTransfer.files[0]) {
+      const file = e.dataTransfer.files[0];
+      console.log('file.type: ', file.type);
+      if (file.type === 'text/csv' || file.type === 'text/tab-separated-values') {
```
davidkircos, line 65: not 100% sure this is the only file meme types we can support.

davidkircos:
New Import data menu and how it works for uploading a CSV.

Mar-03-2023 11-48-35

src/ui/menus/TopBar/SubMenus/DataMenu.tsx
```diff
@@ -2,31 +2,61 @@ import Button from '@mui/material/Button';
...
+          </Tooltip>
+        }
+      >
+        <MenuHeader>Import File</MenuHeader>
+        <MenuItem
+          onClick={() => {
+            setShowCSVImportHelpMessage(true);
+          }}
+        >
+          <MenuLineItem primary="CSV" Icon={UploadFile} />
```
jimniels, line 40: Did you decide not to put like "Import: CSV" in the command palette?

CleanShot 2023-03-03 at 19 22 34@2x

If we do, we'd probably want to rethink this icon, as it clashes with "Upload local file"

CleanShot 2023-03-03 at 19 21 29@2x

CleanShot 2023-03-03 at 19 23 55@2x



Maybe we'd even just want to remove "Upload local..." for now, and we can have something new when we have the new "Open..." file menu

jimniels, line 40: Did you decide not to put like "Import: CSV" in the command palette?

CleanShot 2023-03-03 at 19 22 34@2x

If we do, we'd probably want to rethink this icon, as it clashes with "Upload local file"

CleanShot 2023-03-03 at 19 21 29@2x

CleanShot 2023-03-03 at 19 23 55@2x



Maybe we'd even just want to remove "Upload local..." for now, and we can have something new when we have the new "Open..." file menu

src/grid/actions/insertData/insertCSV.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+    complete: function () {
+      props.sheetController.end_transaction();
+    },
+    step: function (row) {
+      const cellsToInsert: Cell[] = [];
+
+      // results.data.forEach((row, rowIndex) => {
+      //@ts-expect-error
+      row.data.forEach((text, cellIndex) => {
```
jimniels, line 31: Why is typescript yelling at you here? because you don't have defined types? (you also don't on lines 26 and 20, so wouldn't it be yelling at you there too? ha

src/grid/actions/insertData/insertCSV.ts
```diff
@@ -0,0 +1,50 @@
...
+    error: function (error, File) {
+      props.reportError(error.name + ': ' + error.message);
+    },
+    complete: function () {
+      props.sheetController.end_transaction();
+    },
+    step: function (row) {
+      const cellsToInsert: Cell[] = [];
+
+      // results.data.forEach((row, rowIndex) => {
```
jimniels, line 29: Seems like you should just delete this line (and 41) if you're not going to use it...

src/ui/components/FileUploadWrapper.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  // triggers when file is dropped
+  const handleDrop = async (e: DragEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
+    setDragActive(false);
+
+    if (e.dataTransfer.files && e.dataTransfer.files[0]) {
+      const file = e.dataTransfer.files[0];
+      console.log('file.type: ', file.type);
```
jimniels, line 64: Remove this console.log?

src/ui/components/FileUploadWrapper.tsx
```diff
@@ -0,0 +1,117 @@
...
+          onDrop={handleDrop}
+          style={{
+            position: 'absolute',
+            width: '100%',
+            height: '100%',
+            top: '0px',
+            right: '0px',
+            bottom: '0px',
+            left: '0px',
+            backgroundColor: 'rgba(0,0,0,0.5)',
```
jimniels, line 103: &lt;my-two-cents>

I kinda feel like i've been avoiding the "dark overlays" for modals/dialogs (at least the few i've done).

I really like the approach you're taking of "just drag something onto the grid" when you want to bring in a file, like it's no big deal, no big import process. Very much like Figma, where you just drag n drop something, boom, done. CMD+Z if you don't like it.

So this big black overlay feels heavy-handed in the experience of the app.

I have two thoughts:

1. Just don't have an overlay at all. Just have it be like Figma: you drag and drop but the UI doesn't change when you're doing it. 
2. Maybe a white overlay instead of a black one, e.g.

CleanShot 2023-03-03 at 19 33 50@2x

The white overlay just feels like it fits more with the overarching "light" aesthetic of Quadratic.

But part of me wonders if maybe we could getaway with having no UI at all (no backdrop) for the import?

I'm honestly fine whatever you decide to go with.

&lt;/my-two-cents>

src/ui/components/PresentationModeHint.tsx
```diff
@@ -14,13 +14,11 @@ export default function PresentationModeHint() {
   }, [presentationMode]);
 
   return (
-    <Snackbar
-      anchorOrigin={{ vertical: 'bottom', horizontal: 'center' }}
+    <QuadraticSnackBar
```
jimniels, line 17: Ah this is smart, just re-use what we have.

We'll probably want to refactor this at some point to have a single-use messaging system that accounts for the idea of successive messages. If you do these two things quickly (drag n drop an unsupported file, then quickly go into presentation mode), they'll overlay each other, which isn't a big deal right now as i doubt it'll happen to folks, but maybe we should file an issue?

CleanShot 2023-03-03 at 19 37 57@2x

src/ui/menus/TopBar/SubMenus/FormatMenu/FormatMenu.tsx
```diff
@@ -56,7 +56,7 @@ export const FormatMenu = (props: IProps) => {
     <Menu
       onMenuChange={onMenuChange}
       menuButton={
-        <Tooltip title="Format" arrow disableInteractive enterDelay={500} enterNextDelay={500}>
+        <Tooltip title="Cell format" arrow disableInteractive enterDelay={500} enterNextDelay={500}>
```
jimniels, line 59: I like this renaming, but minor consistency thing.

With this change, the menus are now:

- Main menu
- Import data
- Cell format
- Number format

I feel like we should either stick to nouns across the board, which would mean changing "Import data" which sounds more verb-ish to "Data import", e.g.

- Main menu
- Data import
- Cell format
- Number format

src/ui/menus/TopBar/SubMenus/DataMenu.tsx
```diff
@@ -2,31 +2,61 @@ import Button from '@mui/material/Button';
...
+        menuButton={
+          <Tooltip title="Import data" arrow disableInteractive enterDelay={500} enterNextDelay={500}>
+            <Button style={{ color: 'inherit' }}>
+              <DataObjectOutlined fontSize="small"></DataObjectOutlined>
+              <KeyboardArrowDown fontSize="small"></KeyboardArrowDown>
+            </Button>
+          </Tooltip>
+        }
+      >
+        <MenuHeader>Import File</MenuHeader>
```
jimniels, line 34: I like this change.

What do you think about more terse language on those headers and a divider between them?

From this:

CleanShot 2023-03-03 at 19 45 25@2x

To this:

CleanShot 2023-03-03 at 19 49 12@2x

jimniels (approved):
Great feature! This is going to be a huge boon to getting Quadratic into a useful state _you need your own data_ and csv imports is a great first step!

I love the way you ended up tackling it with a single interface for bringing in data: drag n drop! Feels very natural, especially for an "infinite canvas" app (vs. like a 3-step "import wizard" ha)

I left a few comments/suggestions, none of them blocking. Feel free to read them and make changes as you see fit.

Otherwise 👍

src/ui/components/FileUploadWrapper.tsx
davidkircos: I got rid of the overlay, this feels better to me!

src/grid/actions/insertData/insertCSV.ts
davidkircos: Yeah, `row.data` isn't a defined type. I added some better type annotations, and TS still isn't super happy.

src/ui/menus/TopBar/SubMenus/DataMenu.tsx
davidkircos: Agreed! Changed.

src/ui/menus/TopBar/SubMenus/FormatMenu/FormatMenu.tsx
davidkircos: Agreed, changed!

src/ui/components/PresentationModeHint.tsx
davidkircos: Yeah, this isn't a common thing to happen. MUI supports a auto stacking thrid party library https://github.com/iamhosseindhv/notistack

PR #311, opened 2023-02-24
Significantly speed up deletion when deleting a large range of cells

davidkircos:
Steps to reproduce:

1. Open app.QuadraticHQ.com and load the monte carlo example file.
2. Select all and press backspace. 
3. Time how long it takes. (for me it freeze the browser)

Try this same procedure using this branch, it should delete and be responsive again within a second.


src/grid/actions/DeleteCells.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const DeleteCells = async (args: DeleteCellsArgs) => {
+  const { x0, y0, x1, y1, sheetController, pyodide, app, create_transaction } = args;
+
+  if (create_transaction ?? true) sheetController.start_transaction();
+
+  // delete cells row by row
+  for (var current_row = y0; current_row <= y1; current_row++) {
+    const cells_to_delete = sheetController.sheet.grid.getNakedCells(x0, current_row, x1, current_row);
```
davidkircos, line 23: This used to do the whole range at once, how it does it row by row

jimniels (approved):
This looks good to me. It definitely froze up the tab before and now it takes ~1 second. Nit pick is: it could be faster ;)

I did notice one small issue unrelated to this PR while testing (https://github.com/quadratichq/quadratic/issues/313) but otherwise this LGTM 👍

Issue #22, opened 2022-02-18
feat: Data Types


davidkircos:
- [ ] Have grid cells format text for common data types: currency, date, number, etc
- [ ] In Python take cells input and interpret them natively in their Quadratic Cell type. Ex if a user inputs a date, when the cell data arrives in python, it should be a DateTime type.

jimniels:
Example where I had to use `int` to get the result i wanted:

I was trying to take a range of cells and multiply each value by 10, but as you can see, i wasn't expecting `11*10` to be `1111111111`

![image](https://github.com/quadratichq/quadratic/assets/1316441/a1fd0038-bcee-48f8-b2fe-5e9a642d0ce6)

Once i wrapped the cell value in int then it worked as expected:

![image](https://github.com/quadratichq/quadratic/assets/1316441/669ecda9-5323-446d-bcd3-a71b37705efa)

jimniels:
Here's an example where I wanted to work with the value of a cell as a string, but had to convert it using `str()` first

<img width="929" alt="CleanShot 2023-05-11 at 16 44 50@2x" src="https://github.com/quadratichq/quadratic/assets/1316441/8fa65c25-94ff-458c-bb00-9db87dc31d6d">

Once I wrap what `cell` returns, e.g. `str(cell(x, y))` then I can run `.split()`

<img width="933" alt="CleanShot 2023-05-11 at 16 45 04@2x" src="https://github.com/quadratichq/quadratic/assets/1316441/83670b1a-870a-4dc8-ada6-68050bdf5ad4">

jimniels:
[From the discord](https://discord.com/channels/1021488427977416714/1109582323575439400/1109943455238586388):

> Another small developer experience suggestion, right now cell(x, y) return an object, which forces me to wrap each cell in a str call. Maybe there's a way to add an alternative method that returns it as a str or int directly?

davidkircos:
will be supported starting in `next`

PR #318, opened 2023-02-25
Alignment

### LINKED ISSUE #290: feat: Text Alignment
- [x] Right Alignment
- [x] Center Alignment
- [x] Auto Right for Numbers
- [x] Auto Left for Text
- [x] Check Overflow properly when left, right, and center.

davidfig:
Fixes #290


davidkircos (changes requested):
Overflow bug when aligning right.

Screenshot 2023-02-26 at 10 07 23 AM

davidkircos:
@davidfig 

Numbers can not be aligned left (we should default to 'auto' then the user could force one to align left)
Strings that start with a number, but are not a number are also aligned right by default. Should only align right when the cell contains only a valid number.

Feb-26-2023 10-36-14

src/ui/menus/CommandPalette/ListItems/Format.tsx
```diff
@@ -80,6 +89,29 @@ const ListItems = [
       return <CommandPaletteListItem {...props} icon={<DecimalDecrease />} action={textFormatDecreaseDecimalPlaces} />;
     },
   },
+  {
+    label: 'Format: Align Left',
```
jimniels, line 93: We should change these to match the labels we're using for bold/italitc which are prefixed with the word 'Text' and use sentence case, 

CleanShot 2023-02-27 at 13 33 45@2x

In other words:

- Text: Align left
- Text: Align right
- Text: Align center

(Also these live in a different file, see `src/ui/menus/CommandPalette/ListItems/Text.tsx`

davidkircos (changes requested):
Partial review:
- Cells starting with a number that are not a number should be left aligned by default like text. "50 hello" should align to the left.
Screenshot 2023-03-04 at 11 24 59 AM

davidfig:
Turns out isNaN for strings is way more complicated than I thought. Here's the best solution based on my research:

```
// from https://stackoverflow.com/a/58550111
function isStringANumber(s: string): boolean {
  return s.trim() !== '' && !isNaN(s as any as number);
}
````

**Comment by davidkircos** [comment]
@davidfig, that makes sense. For consistency, I also replaced the function that decides if a `cell.value` is a number in the `CellTextFormatter` with the same function. See this commit. https://github.com/quadratichq/quadratic/pull/318/commits/a7b374743c634f57c4e4debeca0de483e99959b1

However, there is still an issue. We need to check if the label is a number _before_ applying the CellTextFormatting, not after. Otherwise, adding something like Currency formatting changes the default alignment. See gif.
Mar-04-2023 20-59-12

**Comment by davidkircos** [comment]
@davidfig seeing some odd behavior with formatting jumping around when loading files. It also appears to be a bit of a performance hit when true rendering. When loading the Monte Carlo sim doc it seems to be rendering it a few times, and formatting jumps around (see gif)
Mar-05-2023 09-45-18

File: .env [hunk 2]
```diff
@@ -1 +1,2 @@
-REACT_APP_VERSION=$npm_package_version
\ No newline at end of file
+REACT_APP_VERSION=$npm_package_version
+REACT_APP_DEBUG=1
```
davidkircos, line 2: this will put debug mode on in prod. use `.env.local` to override which is in the gitignore

davidkircos (changes requested):
Some text overlaps 

CleanShot 2023-03-20 at 14 45 38@2x

Numbers too big to fit into a cell should not remain left-aligned. They should either right-align or default to exponents. What do you think?

jimniels:
Ok, I update this PR with the latest from main and resolved merge conflicts. After playing around with it sum, these are my thoughts:

## Big takeaway

I personally feel like we should start this feature super conservatively.

What do I mean by that? We provide the controls for alignment in the UI to the user but don't auto-align anything. Everything continues to be left-aligned by default. If the user wants to align something, they now have the controls but otherwise we're not taking a stance and implementing any features around auto-alignment based on cell data types.

Separately, we can file an issue where we track the kinds of things that feel like they should be aligned by default, e.g. cells that are only numbers, cells where the user starts typing `$` followed by numbers, etc., which all feels like maybe it relates to the concept of #22 too.

In short, I personally don't think we're quite ready to go any further in the product with alignment other than simply: allow the user to align things how they want.

## Small item

Editing a cell doesn't preserve alignment (feels like it should? you can do that with an `<input>` can you not)? If nothing else, we should make sure we do this in #454

## Conclusion

If we do the two items above, IMO it's ready to be merged.

jimniels (approved):
Once we remove the auto-alignment (per [this comment](https://github.com/quadratichq/quadratic/pull/318#issuecomment-1552134373)) I'm good with these changes 👍

PR #326, opened 2023-02-27
feat: make presentation mode go full screen

jimniels:
Make the presentation mode go full screen by hiding header/footer and provide a little UI hint on how to get out of that mode that disappears after ~3 seconds

Feb-27-2023 14-45-58

The code editor would stay open if it's already open

Feb-27-2023 14-46-57

---

Figma does this nicely when you're in a prototype and you show/hide the figma UI, it tells you how to exit and hides the message after `n ` seconds

CleanShot 2023-02-27 at 14 35 45@2x

### DIFF
--- a/src/gridGL/pixiApp/PixiApp.ts
+++ b/src/gridGL/pixiApp/PixiApp.ts
@@ -169,7 +169,7 @@ export class PixiApp {
     this.destroyed = true;
   }
 
-  private resize = (): void => {
+  resize(): void {
     if (!this.parent || this.destroyed) return;
     const width = this.parent.offsetWidth;
     const height = this.parent.offsetHeight;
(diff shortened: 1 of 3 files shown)


davidkircos:
I like this change. However it stretches the canvas vertically when hiding the topbar and bottombar. @davidfig do you know how to tell pixi to rescale the canvas?

davidfig:
PixiApp.resize should do it.

jimniels:
`resize` was private on `app` so [i just made it public](https://github.com/quadratichq/quadratic/pull/326/files#diff-07dfae82d5a245c6f7a92c5b147bca8506b4cc9938ffcd2c5e634ae8c73cd4aeR172) (any problem with that?)

and [i added an effect](https://github.com/quadratichq/quadratic/pull/326/files#diff-020d66b82eb8cfed3573be863da049dd39242a7856be2aba1a38ad8dd63866dfR35-R39) so anytime `presentationMode` changes, it runs a resize. Seems to work great!

PR #332, opened 2023-02-28
fix: bind `this` to the correct context

jimniels:
I broke grid resizing [here](https://github.com/quadratichq/quadratic/pull/326/commits/44b1dccf2bd1944e35b974c71aa9666f58557389#diff-07dfae82d5a245c6f7a92c5b147bca8506b4cc9938ffcd2c5e634ae8c73cd4aeL172) so right now when you resize the browser, the grid doesn't redraw (it stretches like a bitmap image).

But this fixes it by making sure `this` is bound to the pixi app (not the window)

### DIFF
--- a/src/gridGL/pixiApp/PixiApp.ts
+++ b/src/gridGL/pixiApp/PixiApp.ts
@@ -175,7 +175,7 @@ export class PixiApp {
     this.destroyed = true;
   }
 
-  resize(): void {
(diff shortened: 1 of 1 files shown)


davidkircos (approved):

davidfig:
The class function = design pattern ensures this is set right w/o having to manually bind it :) It's a useful pattern.

PR #336, opened 2023-03-03
Add array literals in formulas

HactarCE:
Uses same syntax as Excel: e.g., `{1, 2, 3; 4, 5, 6}`

Empty arrays and nonrectangular arrays are not allowed

Also added link to new [formula docs](https://docs.quadratichq.com/formulas)


davidkircos (changes requested):
Increment `quadratic-core` version and update the link to docs for formulas in CodeEditor.

davidkircos (approved):

PR #339, opened 2023-03-04
Small fixes

davidkircos:
- [x] Press Escape to leave Presentation Mode
- [x] Presentation Mode set to false on page reload

### DIFF
--- a/src/ui/components/PresentationModeHint.tsx
+++ b/src/ui/components/PresentationModeHint.tsx
@@ -1,7 +1,6 @@
 import { useState, useEffect } from 'react';
 import { Snackbar } from '@mui/material';
 import { useGridSettings } from '../menus/TopBar/SubMenus/useGridSettings';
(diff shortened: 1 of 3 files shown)


src/ui/components/PresentationModeHint.tsx
```diff
@@ -21,7 +20,7 @@ export default function PresentationModeHint() {
         setOpen(false);
       }}
       autoHideDuration={5000}
-      message={`Press “${KeyboardSymbols.Command}.” to exit presentation mode.`}
+      message={`Press “Escape” to exit presentation mode.`}
```
jimniels, line 23: Minor: I feel like we should make this be “ESC” since we use that elsewhere to denote "exit", like in the code editor close tooltip:

CleanShot 2023-03-03 at 18 58 07@2x

Then when we revamp keyboard shortcuts, this will just be another appearance of that functionality.

jimniels (approved):
Looks good. One comment about the "Escape" but otherwise LGTM 👍

PR #342, opened 2023-03-06
feat: allow user to edit filename

jimniels:
- Allow user to edit file name in menu bar (a la Figma)
- Don't allow a rename to empty filename
- Truncate at a really long name when saved

Mar-06-2023 14-09-55

### DIFF
--- a/src/ui/menus/TopBar/TopBar.tsx
+++ b/src/ui/menus/TopBar/TopBar.tsx
@@ -1,4 +1,5 @@
-import { Box, Typography, IconButton } from '@mui/material';
+import { useEffect, useRef, useState } from 'react';
+import { Box, Typography, IconButton, InputBase } from '@mui/material';
 import { useRecoilState } from 'recoil';
 import { editorInteractionStateAtom } from '../../../atoms/editorInteractionStateAtom';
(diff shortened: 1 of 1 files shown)


src/ui/menus/TopBar/TopBar.tsx
```diff
@@ -27,8 +28,26 @@ interface IProps {
 
 export const TopBar = (props: IProps) => {
   const [editorInteractionState, setEditorInteractionState] = useRecoilState(editorInteractionStateAtom);
-  const { currentFilename } = useLocalFiles(props.sheetController);
+  const { currentFilename, renameFile } = useLocalFiles(props.sheetController);
+  // @ts-ignore TODO why would currentFilename be possibly undefined?
```
jimniels, line 32: Why does `currentFilename` come back as `string | undefined`? What is the case where we would support a file not having a name?

jimniels: Which is causing a bunch of typescript errors and react errors in the console related to going from a controlled to uncontrolled component

jimniels:
Also doesn't appear to be working in the file list? If i rename a file, it doesn't actually show up as being renamed in the file list (even though i'm using the `renameFile` function)

Mar-06-2023 14-15-57

PR #348, opened 2023-03-11
refactor: load from url context

jimniels:
Use context instead of multiple versions of `useLocalFiles` hook

Also: consider whether we want to make this a generic, app-wide context for also using `app` and `sheetController` or if we want those to be separate contexts....


src/storage/useLocalFiles.ts
```diff
@@ -41,11 +40,12 @@ function log(...s: string[]): void {
 }
 
 export const useLocalFiles = (sheetController: SheetController): LocalFiles => {
+  const [hookLoaded, setHookLoaded] = useState<boolean>(false);
+  // Move this recoil state into the hook
```
jimniels, line 44: By making this a hook we only use once, we can move all state into the hook itself...

src/ui/QuadraticUI.tsx
```diff
@@ -18,6 +18,16 @@ import { useGridSettings } from './menus/TopBar/SubMenus/useGridSettings';
...
 import { CSVImportHelpMessage } from './overlays/CSVImportHelpMessage';
+import { createContext } from 'react';
+import { LocalFiles } from '../storage/useLocalFiles';
+
+// TODO move into its own file...
+interface AppContextProps {
+  localFiles: LocalFiles;
+  app: PixiApp;
+  sheetController: SheetController;
+}
```
jimniels, line 29: Would we want to put all top-level state like this in one place? Drawbacks? Benefits?

Or should we have separate contexts, like

`SheetControllerContext` and `AppContext` and `LocalFilesContext`...

src/ui/QuadraticUI.tsx
```diff
@@ -43,40 +53,42 @@ export default function QuadraticUI(props: Props) {
...
         }}
       >
-        <FileUploadWrapper sheetController={sheetController} app={app}>
-          <QuadraticGrid sheetController={sheetController} app={app} />
-        </FileUploadWrapper>
-        <CodeEditor editorInteractionState={editorInteractionState} sheet_controller={sheetController} />
-      </div>
+        {editorInteractionState.showCellTypeMenu && <CellTypeMenu></CellTypeMenu>}
+        {showDebugMenu && <DebugMenu sheet={sheetController.sheet} />}
+        {!presentationMode && <TopBar />}
```
jimniels, line 67: I used this as an example where we don't have to pass `app` and `sheetController` anymore

src/ui/menus/TopBar/TopBar.tsx
```diff
@@ -11,24 +11,22 @@ import { NumberFormatMenu } from './SubMenus/NumberFormatMenu';
...
-
-export const TopBar = (props: IProps) => {
+export const TopBar = () => {
   const [editorInteractionState, setEditorInteractionState] = useRecoilState(editorInteractionStateAtom);
-  const { currentFilename, renameFile } = useLocalFiles(props.sheetController);
+  const {
+    localFiles: { currentFilename, renameFile },
+    app,
+    sheetController,
+  } = useContext(AppContext);
```
davidkircos, line 29: All this component does is pass down the app and sheetController. we wouldn't even need the context in this file, just the sub-components where it is actually used.

jimniels: Right. Yeah now it's still there because i didn't bother to wire up `useContext` in all this component’s children, but in theory once i did, yes i wouldn't need it in this component.

it's kind of a nice upgrade path because you don't have to touch every file that uses sheetController or app to start integrating this pattern into the app

src/ui/QuadraticUI.tsx
```diff
@@ -8,35 +8,31 @@ import BottomBar from './menus/BottomBar';
...
-
   useEffect(() => {
     sheetController.setApp(app);
   }, [sheetController, app]);
 
+  // Temporary way to attach sheet to global for use in GetCellsDB function
+  useEffect(() => {
+    GetCellsDBSetSheet(sheetController.sheet);
+  }, [sheetController.sheet]);
+
```
jimniels, line 35: Moved this here, [from here](https://github.com/quadratichq/quadratic/pull/348/files#diff-5ebd1e5a57539f916c7b54ba6551eaaf1d11ba2b59440a5b9276451450d11379L56), so we can put all the `app`, `sheetController` and `localFiles` stuff together in one context wrapper.

src/ui/QuadraticUIContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <LocalFilesContext.Provider value={localFiles}>
+      {/* <PixiAppContext.Provider value={app}>
+        <SheetControllerContext.Provider value={sheetController}> */}
+      <QuadraticUI app={app} sheetController={sheetController} />
+      {/* </SheetControllerContext.Provider>
+      </PixiAppContext.Provider> */}
+    </LocalFilesContext.Provider>
+  );
+}
```
jimniels, line 28: My thinking here is: let's put all app-level contexts into the same file.

For now, everyone that needs stuff from the `localFiles` state can do `useContext(LocalFilesContext)` and, in a separate PR, we can add in the contexts for `app` and `sheetController` which will live here in this file and every subsequent file that needs them will do `useContext(PixieAppContext | SheetControllerContext)`

src/ui/QuadraticUI.tsx
jimniels: React’s official docs show where they’re using [a reducer + context](https://react.dev/learn/scaling-up-with-reducer-and-context) and they split the reducer and dispatch into separate contexts, so that’s what i’m gonna do (not use a reducer, just split up the contexts).

I’ll have a context for `LocalFiles` and i’ll have a separate context for app and sheet controller in the same file but different contexts.

Issue #347, opened 2023-03-10
feat: zoom to selection


jimniels:
Today Quadratic has:

- Zoom to 100% `CMD` + `0` - zooms the grid to 100%
- Zoom to fit `CMD` + `9` - zooms the grid to fit all the data that exists on it

Many infinite canvas apps have an additional control called "Zoom to selection" which allows you to zoom to the currently active selection.

![Mar-10-2023 16-10-57](https://user-images.githubusercontent.com/1316441/224445016-4e771aa8-5ff4-4ac1-9a2d-99802b0e2a6e.gif)

We can add this in Quadratic so the current selection on the grid can be zoomed into view (with a sensible buffer around the edges)

Hotkey can be one down from "zoom to fit", so "Zoom to selection" is `CMD` + `8`

<img width="245" alt="CleanShot 2023-03-10 at 16 05 12@2x" src="https://user-images.githubusercontent.com/1316441/224444415-65638bf8-0892-4eb6-ac97-0ac6ebaf81cc.png">

PR #349, opened 2023-03-11
Zoom to selection

davidfig:
Adds #347


jimniels (approved):
Nice! Works just as I would expect. 👍

Issue #328, opened 2023-02-28
feat: Command + Arrow Keys to Select Whole Dataset


davidkircos:
For example when in a cell Command + Shift + DownArrow would select all cells below in a column.

PR #350, opened 2023-03-12
cmd arrow (+ selection)

davidfig:
Handles #328


davidkircos (approved):

jimniels (approved):
I'm not really a power user of this feature in excel, but playing with it in Quadratic then Excel it works as I would expect. This looks good enough to me to get into people's hands 

👍

PR #351, opened 2023-03-12
cmd+wheel scroll

davidfig:
Supports cmd + wheel for zooming. (Replaces #234). This should be tested with a non-magic mouse and a Windows mouse (to ensure I didn't break anything).


HactarCE:
Working well on Windows! Could we also get shift+scroll to scroll sideways?

davidkircos (approved):

PR #353, opened 2023-03-12
resizing headings to the left/above

davidfig:
Resizing headings always works in the same direction


davidkircos (changes requested):
Strange bug where if you are resizing a negative column and let go of the mouse when it's moving the viewport takes inertia and moves a bit (similar to letting go when moving while in drag mode). @davidfig 
Mar-14-2023 16-23-44

davidkircos (approved):

PR #354, opened 2023-03-12
left over quadrants

davidfig:
Fixes bug with quadrant texture not clearing when resizing


davidkircos:
Still able to reproduce this bug. Or did this PR fix a different issue? @davidfig 
Mar-14-2023 16-23-44

davidfig:
This was a different issue (the one where parts of the triangle outputs were left over). This may be fixed in another PR. The black bars are caused by textures of size < 0.

davidkircos:
Ah ok, this does appear to fix that other issue. Nice. I'll do a full review now

davidkircos (approved):

PR #360, opened 2023-03-16
Fix async grid access during formulas + refactors

HactarCE:
- [x] Fix async grid access during formula
- [x] Refactor formula functions in preparation for LSP integration
- [x] Expose `column_name()` and `column_from_name()` to JS

To QA
- [ ] `INDIRECT` function works in Formulas; e.g., `=INDIRECT("En6")`
- [ ] existing formulas seems to work ok


jimniels (approved):
I can't speak for the code, but functionality wise, everything seems to work for me.

CleanShot 2023-05-09 at 11 02 27@2x

PR #363, opened 2023-03-18
Viewport Gets Suck in Zoom Mode 

### LINKED ISSUE #359: bug: Viewport Gets Suck in Zoom Mode
To reproduce: Command + Tab to another app and return via mouse.

Sometimes when returning to Quadratic, I find that the scroll wheel zooms instead of scrolls without Control or Command pressed. Pressing and releasing Control resets this and fixes the issue.

davidfig:
Fixed #359 (maybe?)

- Added the blur handler for wheel. 

I wasn't able to reliably repro this, so it should be tested probably on a window's machine @HactarCE

### DIFF
--- a/src/debugFlags.ts
+++ b/src/debugFlags.ts
@@ -2,7 +2,7 @@
 export const debug = process.env.REACT_APP_DEBUG === '1' ? true : false;
 
 // skip python load (used to speed up debugging)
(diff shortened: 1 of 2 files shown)


davidkircos:
Seems to fix the issue.

davidkircos (approved):

Issue #159, opened 2022-11-16
Right Click - Copy as PNG


davidkircos:
Low-priority feature. 

<img width="205" alt="Screen Shot 2022-11-16 at 12 14 10 PM" src="https://user-images.githubusercontent.com/3479421/202272056-c0d063ab-4c2c-4edb-831d-b1af0bc11490.png">

PR #366, opened 2023-03-18
Copy as png

davidfig:
Implements #159 

- [x] copy to clipboard from menu
- [x] improve menu UI
- [x] add to palette
- [x] bug: only works at 100%
- [x] should only include cells and formatting (not gridlines and cell type outlines)
- [ ] Add a max export size and visible error when failing
- [ ] Always export at same resolution per cell


davidkircos:
This is a sweet feature! Nice.

Couple pieces of feedback 

1) if the section is off screen the PNG only contains what is on screen.
2) I think we should not render grid lines. Have it export like presentation mode.

jimniels:
This is _very_ cool! 

Here's a few things that popped out to me after looking at it:

1) the formatting bar icon should read sentence case "More commands" (not "More Commands") and be a horizontal set of dots instead of vertical

CleanShot 2023-03-21 at 15 55 06@2x

2) If you have the "more" option selected and you close the formatting menu, then make a new selection, the formatting bar pops up again in the same state (it should close the "Copy as PNG" fly out menu)

Mar-21-2023 15-42-52

3) Sentence case the label in the command bar 

CleanShot 2023-03-21 at 15 44 41@2x

4) The style of the formatting bar flyout menu (More...) is different than that of the formatting bar itself, e.g. text color is different, icon sizes appear different, etc. 

CleanShot 2023-03-21 at 15 53 11@2x

Also, the icon in the formatting bar flyout is different than the one in the command palette

CleanShot 2023-03-21 at 15 52 06@2x

We should probably make those icons black like the body text, but for now let's at least just make them consistent in color and let's use a different icon so as to differentiate it from the "Copy" command in the command palette. Example:

CleanShot 2023-03-21 at 16 06 26@2x

And in the formatting bar:

CleanShot 2023-03-21 at 16 06 22@2x

That's a custom icon I made. Happy to do this part once everything else in this PR is resolved. (Note to self: I feel like we're going to quickly outgrow shoving icons into every command, we might want to re-think how/where we apply icons with more judiciousness...)

Also i'm spitballing on that graphic. I'm assuming the discrepancy stems from the formatting bar being a custom implementation (which appears to be 34px tall with 2px padding top/bottom) whereas the flyout menu is using the react-menu library (which appears to be 36px tall with 8px padding top/bottom).

Feels like we really should try to make these appear more cohesive...

6) _I think_ the way this feature works should be uniform regardless of zoom level. For example, these are two screenshots I got using the feature that had the same selection but I was zoomed way out and zoomed way in. Seems like "Copy selection as PNG" should always copy the thing uniformly at 100% regardless of what the user's zoom in the application is.

CleanShot 2023-03-21 at 16 14 51@2x

Here's another example where I zoomed way out and put content way in the upper left and way in the bottom right then made a selection and did "copy as png" and it seemed to export a transparent background?

CleanShot 2023-03-21 at 16 19 11@2x

jimniels:
I updated UI styles so that:

1. The formatting bar has the same style (e.g. color) as the icons in the header and the flyout menu has the same styles as the formatting bar
2. The "more..." button is horizontal
3. Sentence casing in the tooltip

Before:

CleanShot 2023-03-22 at 11 41 37@2x

After:

CleanShot 2023-03-22 at 11 41 48@2x

jimniels:
Updated the command palette UI with a new icon and sentence casing.

Before:

CleanShot 2023-03-22 at 13 50 06@2x

After:

CleanShot 2023-03-22 at 13 49 41@2x

Also updated the formatting flyout menu to use the same icon:

CleanShot 2023-03-22 at 13 52 43@2x

davidkircos:
The screenshot includes content from bordering cells. The frame should go precisely to the selected cells.

image

davidfig:
I'm not sure I understand the 100% comment. Regardless of zoom level, it always copies to the same DPI. (I fixed a bug where it was clipping when some of the selection was offscreen, which was unrelated.) Do you want the zoom level to change on the png based on the screen zoom? That would feel weird.

jimniels:
> Regardless of zoom level, it always copies to the same DPI

Yup, that's what I would expect 💯

davidkircos (changes requested):
Exports are the same resolution per unit area. So in large exports you can zoom in and the content looks good at a cell level.

jimniels:
[I updated the context menu to use the default icon size](https://github.com/quadratichq/quadratic/pull/366/commits/abd32b2e076a4d7f01433096fe2b578af99a72b9) (rather than small).

This helps with the flyout "Copy as PNG" menu from looking disabled (it was using the same color as all the other menus in the app, but because it was smaller — small icon, 14px type — than the other menus – default icon, 16px type – it had the look of being disabled.

CleanShot 2023-03-27 at 11 12 38@2x

Now the floating context menu uses the same icons as elsewhere in the app. I think this is an improvement overall.

CleanShot 2023-03-27 at 11 17 08@2x

davidkircos (approved):

Issue #177, opened 2023-01-06
Horizontal scrolling using Shift+scroll


dumblob:
### Motivation

Spreadsheets have largely been the most common app users need to navigate horizontally. Unfortunately Quadratic seems to only offer horizontal scrolling by arrow keys navigation. So quick mouse navigation is not possible (there are no scrollbars which is IMHO a good UX decision but there must be a "mouse way" to scroll horizontally without putting your hand elsewhere from the mouse).

### Envisioned solution

Make `Shift+scroll` work as horizontal scroll :wink:.

davidkircos:
@jimniels what do you think of adding this?

jimniels:
Makes a lot of sense to me.

I've been using a magic mouse / trackpad for so long, I have no sense for physical scroll wheels on mice anymore. Surely this is a thing in other apps? e.g. does Figma allow horizontal scrolling/panning with SHIFT + scroll?

dumblob:
> Surely this is a thing in other apps?

Yeah, it is very widespread. It originates from Microsoft Windows IIRC.

>e.g. does Figma allow horizontal scrolling/panning with SHIFT + scroll?

IDK but you may try other apps - actually no need to go far away as all web browsers I know of have Shift+scroll as default :wink:.

dumblob:
Awesome. Thanks a lot!

PR #367, opened 2023-03-19
Horizontally scroll with shift + mouse wheel

davidfig:
Implements #177


HactarCE:
Working perfectly on my end!

davidkircos:
With a trackpad, this inverses the scroll. Is it possible to have this only effect the mouse scroll, not the trackpad?

HactarCE (approved):

jimniels (changes requested):
Per [David's comment](https://github.com/quadratichq/quadratic/pull/367#issuecomment-1475465233), this appears to affect panning on the trackpad when you hold down shift.

In prod today, if you hold down shift and pan with a trackpad it works the same as if you hold down shift and pan without it.

But with these changes, shift + pan on a trackpad breaks (it sort of pans back and forth just a tiny bit, regardless of the direction you pan)

jimniels:
This is being implemented in #533

Issue #231, opened 2023-02-08
Cell Type Outlines Extend on Overflowing Cells


davidkircos:
<img width="308" alt="Screenshot 2023-02-07 at 5 39 19 PM" src="https://user-images.githubusercontent.com/3479421/217399213-e9552e87-6a69-4058-a436-c66a2a4a02fd.png">

PR #368, opened 2023-03-19
Extends cell type border to clipping bounds

davidfig:
Implements #231 

Note: this does not extend the Cells Array border since it's not clear how that would work.


davidkircos:
It seems to me the cell outline should quantize to the width of the next cell instead of clipping mid-cell.

PR #370, opened 2023-03-19
Hide portion of Quadrants when typing input

davidfig:
- [x] hide cache under input box
- [x] merged https://github.com/quadratichq/quadratic/pull/369


davidkircos:
@davidfig can you provide more details about what this branch does? I am not sure what to test here.

davidkircos:
Does not appear to work for me 
Mar-21-2023 13-34-20

davidfig:
Can you try again? Seems to work fine for me.

jimniels:
This is not working for me either.

You can see here if I zoom out and edit a cell, it only re-renders once i've edited the cell and hit enter.

Apr-28-2023 10-21-37

jimniels:
This still doesn't work for me.

Have a cell with a value in it, zoom out, start editing it and the previous value doesn't disappear until you hit "ENTER"

May-08-2023 10-28-30

There's like a threshold of zooming out where, once you pass it, it shows the old value _while you're editing_

May-08-2023 10-26-40

PR #373, opened 2023-03-20
Minor improvements

davidkircos:
Vertically center header numbers better
image

Make the header color for selection more distinct.
image

Update Tab names and Documentation
CleanShot 2023-03-21 at 12 13 02@2x


jimniels (approved):
Made a few tweaks to casing, layout, style, etc., in the "documentation" tabs.

Python

CleanShot 2023-03-21 at 15 17 36@2x

Formulas:

CleanShot 2023-03-21 at 15 28 10@2x

Otherwise, this looks good to me! 👍

davidkircos:
We need to return 'result' in the second Python Example.

jimniels:
Good catch. I updated both of these:

CleanShot 2023-03-21 at 16 26 01@2x

PR #376, opened 2023-03-22
feat: show return type for python code

davidkircos:
The goal of this branch is to make Python work better in Quadratic. 

- [x] Clearly display to a user what from their code is being returned to the grid.
- [x] Add back code prettier. 
- [ ] Small output icon next to code once run (disappears when editing)
- [ ] OUTPUT tab showing output details and inspection


jimniels:
This PR attempts to add an indication of live types as you work in the code editor, so you can tell what's being returned to the grid. Working example:

CleanShot 2023-07-05 at 13 48 42@2x

jimniels:
Example "return" from Duckbook.

CleanShot 2023-07-07 at 10 08 26@2x

juanbercoff:
@jimniels Hey, given that the return type and line are already showing, what additional information would you like to show?

jimniels:
@juanbercoff sorry for being so late on this.

Honestly, I don't think I have a very precise answer for you. This was kind of opened as a proof-of-concept, like "can we provide _enough_ information for a piece of UI like this to be useful to the end user?"

In the process, we began to explore "what information is available for display?" 

If i remember correctly, part of the issue was that we don't know the return type _until_ the user runs the code. So there's this kind of inbetween state, e.g.

1. Says the return type
2. User edits code but hasn't run it yet (we don't know the return type)
3. User runs code, now we know the return type

I'm not sure that we ever fully figure out the right UI/UX. So this is definitely more of an exploratory PR than it is us trying to achieve something very specific, if that answers your question.

PR #379, opened 2023-03-26
selection auto complete

davidfig:
- [x] basic dragging corner
- [x] selecting the correct direction when dragging corner into open space (ie, not directly underneath or to the side of the current selection box)
- [x] escape cancels the drag
- [x] releasing activates the algorithms
- [x] delete if smaller than selection
- [x] simple counting algorithms
- [x] text series algorithms
- [x] copy cell instead of just value
- [x] bug w/probably last selection not clearing on release


src/gridGL/helpers/isStringANumber.ts
```diff
@@ -0,0 +1,4 @@
+// from https://stackoverflow.com/a/58550111
+export const isStringANumber = (s: string) => {
```
davidkircos, line 2: There is another function that does this in the code base too. We should probably use the same function everywhere.

https://github.com/quadratichq/quadratic/blob/fa756a30deca0c84e912b633b4c92b4f6afb81d0/src/grid/formatting/cellTextFormatter.ts#L6

davidkircos: I feel like we resolved this in the `alignment` branch, but that hasn't been merged yet

src/theme/colors.ts
```diff
@@ -11,6 +11,9 @@ export const colors = {
   headerBackgroundColor: 0xffffff,
   headerSelectedBackgroundColor: 0xe7f7ff,
   headerCornerBackgroundColor: 0xffffff,
+  boxCellsDeleteColor: 0xff0000,
+  boxCellsColor: 0x0000ff,
+  boxCellsAlpha: 0.5,
```
davidkircos, line 16: @jimniels I'd love to get your feedback on these colors

davidkircos (changes requested):
@davidfig Attempting to auto complete an empty range results in the app crashing 

CleanShot 2023-04-17 at 16 58 55@2x

jimniels:
image

To me, it feels a bit strange that the current selection gets highlighted as "remove" even when nothing is being removed.

I feel like I would expect red if we're removing, green if we're adding, and none if we're not doing either.

image

jimniels:
### Style changes

After playing with this some more, and looking at what excel/sheets do, I feel like a good interaction style here is to keep the blue outline for additive dragging, and then a red outline at the point when a subtractive drag is taking place.

Apr-20-2023 17-00-21

This means that, when dragging the selection for a single cell, you'll never see the "remove" red highlight, because you're never doing a subtractive drag with just one cell.

Apr-20-2023 17-05-17


**Outstanding bug**: this is not working for horizontal dragging. Will have to investigate why the logic is wrong. The positioning seems to be triggering at the wrong place, as you can see here:

Apr-20-2023 16-57-29

### Deleting one, then two or more

When deleting a range down to 2 or more, things work as expected:

deleting-to-2-or-more

But when deleting a range down to a single cell, it doesn't work:

deleting-to-1

src/gridGL/helpers/isStringANumber.ts
davidfig: copy & pasted

jimniels:
Deleting a range down to one cell still doesn't work

https://user-images.githubusercontent.com/1316441/233504607-8d085907-2b72-4c32-93ce-0cd01e97f1fa.gif

jimniels:
horizontal dragging still doesn't highlight red cells properly as you drag

Apr-22-2023 08-33-27

jimniels (changes requested):
horizontal dragging still doesn't highlight red cells properly as you drag column by column

Apr-22-2023 08-33-27

Deleting a range down to one cell still doesn't work

davidkircos (approved):

PR #382, opened 2023-03-28
ignore caps for cell type filter

davidkircos:


jimniels (approved):
CleanShot 2023-03-28 at 09 49 46@2x

Ah great catch. idk how i missed that.

PR #386, opened 2023-03-28
Language gpt4 api

davidkircos:
Checklist
- [x] be able to turn off AI Features if we don't have an API URL Env
- [x] Mark AI features as `experimental`
- [x] Edit tab for AI cell to have docs


src/grid/computations/runCellComputation.ts
```diff
@@ -28,6 +29,18 @@ export const runCellComputation = async (cell: Cell, pyodide?: any): Promise<Cel
...
+    return {
+      success: result.success,
+      std_out: undefined,
+      std_err: result.error_msg,
+      output_value: result.output_value,
+      cells_accessed: [],
+      array_output: result.array_output || [],
+      formatted_code: '',
+      error_span: null,
+    };
```
jimniels, line 43: Kinda feels weird we're not just spreading in what we get back from `runAI`? e.g.

```
let result = await runAI(...);
return {
   ...result,
   // extra stuff here _not_ returned by `runAI`, e.g. `std_out`
}
```

It's like we're selectively picking out pieces of result and not using the defaults returned by `runAI`.

For examples: `array_output`. Inside of `runAI` it is either `parsedResult` or if an error throws its `null` but then here we check if it's exists and default to `[]`. I'm unsure of why this middle tier of logic needs to exist? Why not have `runAI` define the its return values _and defaults_ and then here you just return that plus any extra that's not part of that result?

src/ui/menus/CellTypeMenu/CellTypeMenu.tsx
```diff
@@ -52,6 +53,25 @@ const CELL_TYPE_OPTIONS = [
...
+      <>
+        {' '}
+        <span
+          style={{
+            fontStyle: 'italic',
+            fontWeight: 'bold',
+          }}
+        >
+          Experimental:
+        </span>{' '}
```
jimniels, line 70: What do you think about adding a badge for this rather than this stylized text? We already use these badges to qualify the other languages (where necessary), e.g. "Coming soon" or "Experimental"

CleanShot 2023-04-04 at 12 15 08@2x

.env_example
```diff
@@ -3,3 +3,6 @@ REACT_APP_SENTRY_DSN=https://xxxxxxxxxxxxxxxxxx@xxxxxxxxxxxx.ingest.sentry.io/xx
 REACT_APP_DEBUG=0 // use =1 to enable debug flags
 REACT_APP_AUTH0_DOMAIN=auth-dev.quadratic.to
 REACT_APP_AUTH0_CLIENT_ID=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+REACT_APP_AUTH0_AUDIENCE=https://localhost:8000
+REACT_APP_AUTH0_ISSUER=https://auth-dev.quadratic.to/
+REACT_APP_QUADRATIC_API_URL=http://localhost:8000
```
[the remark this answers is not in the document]
davidkircos: You don't actually need these env vars unless you want to run the API too and do AI requests.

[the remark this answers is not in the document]
davidkircos: If you run locally without them everything should work and AI will just be hidden in the Quadratic UI

src/ui/menus/CodeEditor/AITab.tsx
```diff
@@ -0,0 +1,269 @@
...
+import { CodeBlockParser } from './AICodeBlockParser';
+
+interface Props {
+  editorMode: EditorInteractionState['mode'];
+  evalResult: CellEvaluationResult | undefined;
+  editorContent: string | undefined;
+}
+
+type Message = {
+  role: 'user' | 'system' | 'assistant';
```
jimniels, line 19: Is this just a duplicate version of [this](https://github.com/quadratichq/quadratic/pull/386/files#diff-13b36bf7efa5cd43f15568d53f0e0e4148530d06cf684f1469c91ae2b8d9400bR27)?

src/ui/menus/CodeEditor/AITab.tsx
```diff
@@ -0,0 +1,269 @@
...
+    setPrompt('');
+    try {
+      await fetch(`${apiClientSingleton.getAPIURL()}/ai/chat/stream`, {
+        method: 'POST',
+        headers: {
+          Authorization: `Bearer ${token}`,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(request_body),
+      }).then((response) => {
```
jimniels, line 71: Minor: everything in here feels really big for being inline in a component....maybe this could be split out? e.g.

`.then((response) => streamAiResponse(response, setMessages))`

src/ui/menus/CodeEditor/AITab.tsx
```diff
@@ -0,0 +1,269 @@
...
+            onKeyDown={(e) => {
+              if (e.key === 'Enter') {
+                submitPrompt();
+              }
+            }}
+            placeholder="Ask a question"
+            endAdornment={
+              <InputAdornment position="end">
+                {loading && <CircularProgress size="1.25rem" sx={{ mx: '1rem' }} />}
+                <IconButton size="small" onClick={submitPrompt} edge="end" disabled={loading}>
```
jimniels, line 170: I feel like we should set the color of this to be primary

CleanShot 2023-04-04 at 12 27 36@2x

jimniels (approved):
Left a few comments. Feel free to incorporate or ignore, I'm going to go ahead and approve it now 👍 

Would luv to spend a little time polishing some UI details in this, but can do that in a separate PR.

This is a really neat addition for exploring AI capabilities in the product! 🎉

src/ui/menus/CodeEditor/AITab.tsx
davidkircos: yes one is defined in the API one is defined in the Client. I'm not sure how to directly share types

src/grid/computations/runCellComputation.ts
davidkircos: I cleaned this up a bit and removed unused variables

src/ui/menus/CodeEditor/AITab.tsx
davidkircos: I agree this could be improved, but let's leave it for now so it's very easy to modify. I think we will iterate more on this quickly

PR #388, opened 2023-03-29
fix: tweak async run indicator on code editor

jimniels:
Remove the layout shift when you hit the "Save & run" button by just disabling the icon and putting the async loading indicator next to it (with a little extra padding)

Before:

Mar-28-2023 20-56-50

After:

Mar-28-2023 20-53-18

Subtle — but small things add up!

### DIFF
--- a/src/ui/menus/CodeEditor/CodeEditor.tsx
+++ b/src/ui/menus/CodeEditor/CodeEditor.tsx
@@ -317,6 +317,7 @@ export const CodeEditor = (props: CodeEditorProps) => {
           </span>
         </div>
         <div style={{ display: 'flex', alignItems: 'center', gap: '.5rem' }}>
+          {isRunningComputation && <CircularProgress size="1.125rem" sx={{ m: '0 .5rem' }} />}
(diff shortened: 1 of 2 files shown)


src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -317,6 +317,7 @@ export const CodeEditor = (props: CodeEditorProps) => {
           </span>
         </div>
         <div style={{ display: 'flex', alignItems: 'center', gap: '.5rem' }}>
+          {isRunningComputation && <CircularProgress size="1.125rem" sx={{ m: '0 .5rem' }} />}
           <TooltipHint title="Save & run" shortcut={`${KeyboardSymbols.Command}↵`}>
             <IconButton
```
davidkircos, line 322: Wrap this in a <span> to get rid of the console warnings

davidkircos:
There is a problem with the preview branches based on me messing with some env vars.

Could you add this line `if (process.env.REACT_APP_SENTRY_DSN && process.env.REACT_APP_SENTRY_DSN !== 'none')` to index.js? I think that will fix it.

davidkircos (approved):

PR #389, opened 2023-03-29
fixes copy paste encoding bug

davidkircos:
Fixes recent sentry errors titled "InvalidCharacterError"


jimniels (approved):
Not sure if there were specific use cases I was supposed to test for, but copy/pasting around the app and nothing sticks out as buggy 👍

PR #395, opened 2023-04-03
Added Dockerfile

irjensen:
Added a Dockerfile to simplify setup.

Building image:
image

Running image:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM rust:latest
+
+# Install npm
+RUN apt-get update && apt-get install -y npm
+
+# Install wasm-pack
+RUN curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+
+# Add wasm32 target
+RUN rustup target add wasm32-unknown-unknown
+
+# Copy package.json and package-lock.json
+COPY package*.json ./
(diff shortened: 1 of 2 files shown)


davidkircos:
Quadratic currently has no dependencies besides node and npm. Was there some part of the project you had difficulty getting up and running?

ovizii:
> Quadratic currently has no dependencies besides node and npm. Was there some part of the project you had difficulty getting up and running?

A Dockerfile is just the first step, the next step would be an automated image build. :-)

The reason I am asking for a docker based deployment is that this way I could just run the container with one line, something along the lines of:

docker run -p 1234:80 -v /local/data/folder:/mountpoint/inside/container --restart always quadratiqhq:latest

No more manual installing and updating along these lines.

Run Quadratic Locally
Install npm, rustup, and wasm-pack
Run rustup target add wasm32-unknown-unknown
Build the Rust/WASM npm run build:wasm
run npm install
Run Web npm start

Run Electron npm run dev
You're basically asking about the benefits of Docker. In short, I'd say: I can test any app I like with a few commands and I can just remove everything with one command without having to install anything on the host except docker.
Benefits like: portable, scalable, lightweight, isolation come to mind.

PR #403, opened 2023-04-04
refactor: file schemas

jimniels:
## Goals

Refactor the file schemas, so that:

- Each file schema is versioned and has it's own file (so zod schemas and typescript types are fully independent of previous versions)
  - [version 1.0 schema](https://github.com/quadratichq/quadratic/pull/403/files#diff-a956fa702ea9564169f222f1b17e584378b7dba1ed7da784fd2ea70bbf910cf5)
  - [version 1.1 schema](https://github.com/quadratichq/quadratic/pull/403/files#diff-63372a6cc930c497114d0f3d293149da3814ca1f2fdcf8d67f624cd41070273e)
- Codebase always imports types from most recent file schema, these are accessible as imports without the version number in [`src/schemas`](https://github.com/quadratichq/quadratic/pull/403/files#diff-fd18542e2ce0ef39537cc5aa71c104717cb7c89cccf5e53c2abbbaa7e0c6d5dc)
  - Importing anything from a versioned file schema is essentially an anti-pattern (except for the upgrade functions which upgrade from one schema to the next)
- Types that are shared between elements in the code and grid file structures are [exported from `src/schemas`](https://github.com/quadratichq/quadratic/pull/403/files#diff-fd18542e2ce0ef39537cc5aa71c104717cb7c89cccf5e53c2abbbaa7e0c6d5dcR33-R43) as their own types.
  - Today these types are inferred from the structure of the grid file, but could conceivably vary in the future

## Examples

Anything in the code that mimics the structure of what goes into the grid files can import special types from `src/schemas`, e.g.

```js
/** @file src/schemas/index.js */

// Map the most recent version to the version-agnostic name `GridFile`
import { GridFileV1_1 } from "./GridFileV1_1";
export { GridFileV1_1 as GridFile };
// Export types from the current version that are re-used across the code base
export type CellTypes = GridFile['cells'][0]['type'];


/** @file src/atoms/editorInteractionStateAtom.js */

import { CellTypes } from "../schemas";
export interface EditorInteractionState = {
  mode: CellTypes
}
```

### ToDos

- [x] Tests: file validation works
- [x] Tests: files from v1 upgrade properly to v1_1, especially where `borders[0].horizontal.type` changes from an integer to an enumerated string, e.g. `0` to `line1`
  - [x] File validation for border type enum changes happens in the upgrade function, not in the one-off, pre-v1 file function
- [x] Tests: example files can be imported


src/atoms/editorInteractionStateAtom.ts
```diff
@@ -9,7 +9,7 @@ export interface EditorInteractionState {
   showGoToMenu: boolean;
   showFileMenu: boolean;
   selectedCell: Coordinate;
-  mode: CellTypes;
+  mode: GridFile['cells'][0]['type'];
```
jimniels, line 12: @davidkircos what do you think of this? (and [this](https://github.com/quadratichq/quadratic/pull/403/files#diff-26faa8e82fec0803c77574a0674b6fa0d874f0b7222d95426137b39c4fde82b3R1-R18) and [this](https://github.com/quadratichq/quadratic/pull/403/files#diff-565cacca1292ca964dec30714a5128d8d37acc1b52cb31beec18b558a3aed194R1-R17))

The idea here is: all types for the _current_ grid schema come out of `schemas/` and, rather than export individual types for parts of the GridFile, you always import the current `GridFile` type and then lookup the specific parts you need.

If we're good with this approach, I have like 20 files to go through and update like this :)

---

The alternative would be to create and name specific sub-types of the current `GridFile` in [the main `schemas/` export](https://github.com/quadratichq/quadratic/pull/403/files#diff-fd18542e2ce0ef39537cc5aa71c104717cb7c89cccf5e53c2abbbaa7e0c6d5dc) and then export those for use in the codebase, but I don't love that approach because you have to each sub-type you come up with, rather than always referencing one single type.

Example:

```js
// schemas/index.js
import { GridFileV1_1 } from "./GridFileV1_1";
export { GridFileV1_1 as GridFile };
export type CellTypes = GridFileV1_1['cells'][0]['type'];

// src/atoms/editorInteractionStateAtom.js
import { CellTypes } from "../schemas";
export interface EditorInteractionState = {
  mode: CellTypes
}
```

This stands in contrast to the example diff in this PR, which does:

```js
// schemas/index.js
import { GridFileV1_1 } from "./GridFileV1_1";
export { GridFileV1_1 as GridFile };

// src/atoms/editorInteractionStateAtom.js
import { GridFile } from "../schemas";
export interface EditorInteractionState = {
  mode: GridFile['cells'][0]['type']
}
```

jimniels: We're going to go with named types in `src/schemas/index.js` — see the PR description for more info

src/schemas/GridFileV1.ts
```diff
@@ -0,0 +1,101 @@
...
+            decimalPlaces: z.number().optional(),
+          }),
+        ])
+        .optional(),
+      wrapping: z.enum(['wrap', 'clip']).optional(), // default is overflow
+    })
+    .array(),
+
+  // todo: this goes away when alignment branch is merged
+  // because this goes away, we'll accept anything in runtime parsing
```
davidkircos, line 95: This wont go away in V1 right?

It'll go away in a new file version

src/schemas/GridFileV1_1.ts
```diff
@@ -0,0 +1,148 @@
...
+        .optional(),
+      wrapping: z.enum(['wrap', 'clip']).optional(), // default is overflow
+    })
+    .array(),
+  id: z.string().uuid(),
+  modified: z.number(),
+
+  // todo: this goes away when alignment branch is merged
+  // because this goes away, we'll accept anything in runtime parsing
+  render_dependency: z.any(), // Dependency[];
```
davidkircos, line 100: Will go away in a future file version.

davidkircos (approved):

src/schemas/GridFileV1.ts
jimniels: Good point.

[I refactored the schemas](https://github.com/quadratichq/quadratic/pull/403/commits/5c96cb8ee5205b920e231d1947ac379a344b2fee) to commit to this as a part of their respective file version.

So we should bump the file schema to version `1.2` once we do remove it in #318

Issue #410, opened 2023-04-07
Typing `$10` as a cell value doesn't create a numeric cell


HactarCE:

jimniels:
I think this is the same as #409 

@HactarCE  You ok if we just close this and include `$` as another character you should be able to enter when a cell is focused?

HactarCE:
Ah yep good point!

PR #411, opened 2023-04-07
Strip currency prefix to mitigate #410

HactarCE:
Mitigate #410 by stripping currency prefix when using a string as from .

This still isn't ideal because the result isn't a currency value.


davidkircos:
I think the desired behavior is that when a dollar sign is submitted into a cell that we change the cell formatting to Currency and strip the $ from the underlying value stored.

davidkircos (approved):

PR #418, opened 2023-04-10
Add Amplitude Analytics

davidkircos:


jimniels (approved):
LGTM 👍 

Works in Chrome, but fails in Arc browser (based on Chromium) — presumably because of ad blockers extensions ha

CleanShot 2023-04-10 at 12 57 19@2x

PR #428, opened 2023-04-14
feat: add read-only dialog for mobile

jimniels:
For people who show up to the site on a mobile device, show a dialog indicating that the app is best on desktop.

To test:
- Should not fire for a desktop device
- Should only show the first time you come to the app (on mobile) and, once dismissed, you should never see this message again (unless you clear your cache, as it's stored in localStorage)

CleanShot 2023-04-14 at 12 06 24@2x


davidkircos (approved):
Code looks good. Need to review UI once the preview builds.

PR #433, opened 2023-04-21
fix: message for failure to load local file (and global snackbar)

jimniels:
Today, if you try to load a file that has become corrupted, the UI becomes unresponsive when you click on one of your files in memory.

This adds a message to the user noting that something went wrong:

Apr-21-2023 16-45-48

Also: this PR refactors the app so these kinds of "global notifications" can now be triggered from a single component. Previous messages that were making their own snackbar:

- CSV import help
- CSV import error
- Copied PNG

Apr-21-2023 16-37-25

Notifications replace any that are currently visible. This can be refactored later, as necessary, if we need a queue of messages that stack. For now, this seems fine given users shouldn't trigger a lot of these.

Other code changes:

- Refactored contexts & providers, now we have a `/contexts` folder with the contexts for the app (`sheetController` and `app` can come later in another PR)
- `useLocalFiles` was renamed to `useGenerateLocalFiles` to denote its one time use, and then we stick that state in the context and that's where we pull the values from elsewhere in the app


src/ui/menus/FileMenu/FileMenu.tsx
```diff
@@ -130,8 +133,19 @@ export function FileMenu(props: FileMenuProps) {
...
+                          loadFileFromMemory(id).then((loaded) => {
+                            if (loaded) {
+                              onClose({ reset: true });
+                            } else {
+                              addGlobalSnackbar('Failed to load file.');
+                              Sentry.captureEvent({
+                                message: 'User file became corrupted',
+                                level: Sentry.Severity.Info,
+                                // TODO send along the corrupted file
+                                // extra: { file: ... }
```
jimniels, line 145: We should probably pass along the file here, but this function wasn't architected to pass that info back when it pulls it out of localStorage. 

Given that this is a rare thing, i figured we'll just note it in Sentry and refactor `localFileFromMemory` to return this info for posting later as necessary.

davidkircos:
Code looks good. 

What do you think about removing the "x" on the Snackbar?

CleanShot 2023-04-24 at 11 51 24@2x

jimniels:
I think we should keep it by default. We could make it optional if we really want, but I like that it auto-dismisses after a few seconds, but you as the user always have the option to dismiss it yourself, like what Gmail does.

CleanShot 2023-04-24 at 13 21 48@2x
CleanShot 2023-04-24 at 13 21 52@2x

davidkircos:
I like Figma's. 

It just closes when you click it, but doesn't explicitly have an "x"

CleanShot 2023-04-24 at 15 31 52@2x

jimniels:
Personally, I still prefer the explicitness of the 'x' because I prefer being more obvious and explicit in UIs so it's clear what you can and cannot do. Clicking to hide just isn't obvious — but maybe it doesn't need to be?

What do you like about not showing the 'x'? Just because there's less UI? I like it because it's obvious "oh i can close that".

What if we diminished the visual prominence of the 'x' by lowering its opacity and shrinking its size a bit?

Apr-24-2023 16-07-06

davidkircos (approved):

Issue #426, opened 2023-04-14
Keyboard shortcuts should use `key` rather than `code`


HactarCE:
Users of non-QWERTY keyboard layouts expect keybinds such as <kbd>Ctrl</kbd>+<kbd>Z</kbd> and <kbd>Ctrl</kbd>+<kbd>O</kbd> to be based the symbol produced by the key rather than physical key. For example, I press the key labeled <kbd>;</kbd> to produce the letter <kbd>O</kbd>, so I expect the "Open" command to be bound to <kbd>Ctrl</kbd>+that key.

Sidenote: this is why I thought that <kbd>Ctrl</kbd>+<kbd>O</kbd> wasn't being overridden correctly.

jimniels:
THis should be fixed for current shortcuts in the app when we merge https://github.com/quadratichq/quadratic/pull/434#pullrequestreview-1398751126

jimniels:
This was fixed in #434

PR #434, opened 2023-04-22
Fix clipboard issue

bartbroere:
Just now I was using the Quadratic spreadsheet but I didn't get copying and pasting to work while using the Dvorak layout. It didn't work for `Ctrl + QWERTY's C location` and for `Ctrl + Dvorak's C location`.

The interchanged use of `event.code` and `event.key` seemed intentional, but because `Ctrl + QWERTY's C` location on my pc translates to `Ctrl + j`. In Chrome this opens my Downloads view.

It might be less confusing if we don't have the shortcut as  `Ctrl + QWERTY's C location`, but simply as `Ctrl + c`.

If we want to change this for the following key shortcuts, I could also do it for all the other shortcuts that may have unintended side effects.


src/gridGL/interaction/keyboard/keyboardClipboard.ts
```diff
@@ -19,7 +19,7 @@ export function keyboardClipboard(props: {
   const { event, interactionState, sheet_controller, app, snackbar } = props;
 
   // Command + V
-  if ((event.metaKey || event.ctrlKey) && event.code === 'KeyV') {
+  if ((event.metaKey || event.ctrlKey) && event.key === 'v') {
```
davidkircos, line 22: Should we do `event.key.toLowerCase()` ?

davidkircos:
Screenshot 2023-04-23 at 8 26 26 AM
Sounds like for letter keys that change in different layouts using .key is better and for modifier keys the .code is better

src/gridGL/interaction/keyboard/keyboardClipboard.ts
bartbroere: Good idea, I think the reader can spot more easily that SHIFT also needs to be pressed. (And I'm not sure how CAPS LOCK would influence this.)

src/gridGL/interaction/keyboard/keyboardClipboard.ts
```diff
@@ -28,7 +28,7 @@ export function keyboardClipboard(props: {
   }
 
   // Command + Shift + C
-  if ((event.metaKey || event.ctrlKey) && event.shiftKey && event.code === 'KeyC') {
+  if ((event.metaKey || event.ctrlKey) && event.key === 'C') {
```
bartbroere, line 31:
```suggestion
if ((event.metaKey || event.ctrlKey) && event.shiftKey && event.code.toLowerCase() === 'c') {
```

For a bit more readable code.

src/gridGL/interaction/keyboard/keyboardClipboard.ts
jimniels: Looks like CAPS LOCK doesn't affect it when it's used by a modifier key, e.g. `v` with CAPS LOCK is `V`, but `v` with CMD + SHIFT + v is lowercase `v`

jimniels: Which means we should probably [not do `.toLowerCase()` on Ctrl + Shift + C, right](https://github.com/quadratichq/quadratic/pull/434/files#diff-c43cfe0602c21d80a8b855bbbf6593adca98152b645724eb700f09d5e7075818R31)?

jimniels: That one was actually still using `e.code` so [I changed it](https://github.com/quadratichq/quadratic/pull/434/commits/a03b432e87b0f6642b2b6ab6e341ec613ebe894d)

bartbroere: Interesting, so Cmd + Shift + 'V' (uppercase) also works with Caps Lock enabled?

And thanks for fixing the other occurrences. I did a search for `event.code` but not for `e.code`

jimniels: Well, AFAIK we don't implement Cmd + Shift + V as a shortcut (though maybe we should!) But yes, from what I can tell, if you have CAPS LOCK on and hit an alpha character, you get the uppercase version with `e.key`. Same with SHIFT + alpha character. But if you use a modifier key, like Cmd or Ctrl, with CAPS LOCK or shift + an alpha character, `e.key` will be lowercase (you could play with it on [this nifty site](https://www.toptal.com/developers/keycode))

Yeah no worries! Thanks for bringing this to our attention!

jimniels (approved):
I added to this by using `.key` for all places in the codebase where we do `.code` as well as fixing one or two minor things.

Otherwise, this is a great PR. Thanks @bartbroere for opening this and bringing it to our attention. Once all the tests etc., merge, I'm good to see this merge cc @davidkircos

jimniels:
Fixes #426

PR #441, opened 2023-04-25
refactor: swap out uuid lib for native randomUUID

jimniels:
Generate a UUID via the browser's built-in [`crypto.randomUUID`](https://caniuse.com/mdn-api_crypto_randomuuid)

This results in less JS code to ship to the client.

Resolves #421

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,6 @@
         "@types/react": "^17.0.33",
         "@types/react-color": "^3.0.6",
         "@types/react-dom": "^17.0.10",
-        "@types/uuid": "^9.0.1",
         "ace-builds": "^1.4.13",
         "axios": "^0.24.0",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -17299,9 +17292,8 @@
       }
     },
     "node_modules/quadratic-core": {
-      "version": "0.1.3",
-      "resolved": "file:quadratic-core/pkg",
-      "license": "MIT"
+      "resolved": "quadratic-core/pkg",
+      "link": true
```
jimniels, line 17296: Not sure why these are changing? Is this problematic?

jimniels:
Turns out, the uuid is using the browser's native version of this where applicable https://github.com/uuidjs/uuid/pull/600

So this probably isn't necessary. Closing.

PR #447, opened 2023-04-27
Live Cell

SueHeir:
Right-click a cell then click the time icon. The code is set to update every minute but can be changed in QuadraticUI.tsx for now. Does not play well with the current cyclic Python bug.  Works well with something along the lines of generating a random number. 

Best,
Liz


davidkircos:
Thanks for your contribution. This is a cool feature! I'd love to think through how this could also be made to run on a node server in the background.

Not sure why a preview branch wasn't created. I'll have to look into that.

SueHeir:
#
>  I'd love to think through how this could also be made to run on a node server in the background.
That would be really cool!

Also, I added a DFS method to find cyclic cell dependencies, it makes the Live Cells play well with Python scripts updating their own values.  I wouldn't add this code, the whole graph is searched every time you change anything, just thought I'd point out the bug:)

PR #451, opened 2023-04-29
fix: allow wider array of characters for initial cell input

### LINKED ISSUE #409: feat: allow wider array of characters for initial cell input
Today, you can begin to edit a cell with alphanumeric character or `.` [Code](https://github.com/quadratichq/quadratic/blob/main/src/gridGL/interaction/keyboard/keyboardCell.ts#L146)

This means, for example, that you cannot type a negative number by typing `-` first. You'd have to enter the number, then move your cursor to the beginning and type `-`.

Other non-alphanumeric characters are also affected, for example typing someone's username into a cell, e.g. `@jimniels`. You cannot begin editing a cell by type `@`. Same with inputing a dollar amount, you can't start with `$` to type `$10`

We should support a wider array of characters for this.

jimniels:
Fixes #409 (mimics how sheets/excel work, which allow for basically any character as the first character in a cell)


src/gridGL/interaction/keyboard/keyboardCell.ts
```diff
@@ -12,7 +11,7 @@ export function keyboardCell(options: {
   setInteractionState: React.Dispatch<React.SetStateAction<GridInteractionState>>;
   editorInteractionState: EditorInteractionState;
   setEditorInteractionState: React.Dispatch<React.SetStateAction<EditorInteractionState>>;
-  app?: PixiApp;
+  app: PixiApp;
```
jimniels, line 14: I went ahead and made this not optional, since that seems to be an old way the app was architected. Typescript didn't yell at me, so yolo.

src/gridGL/interaction/keyboard/keyboardCell.ts
```diff
@@ -103,6 +103,7 @@ export function keyboardCell(options: {
       });
     }
     event.preventDefault();
+    return false;
```
jimniels, line 106: I added a `return false` to all the above because, eventually they all return false anyway. And if i don't have it, then I have to explicitly check for `event.key.length === 1 && ` then "does not equal anything that was pressed up above".

src/gridGL/interaction/keyboard/keyboardCell.ts
```diff
@@ -140,10 +141,11 @@ export function keyboardCell(options: {
...
   }
 
-  // if key is a letter number start taking input
-  if (isAlphaNumeric(event.key) || event.key === '.') {
+  // Any single character key, start taking input
+  if (event.key.length === 1) {
```
jimniels, line 148: Logic here is: just take _anything_ (that wasn't already pressed above) that's a single character (this excludes it from taking keys like "Shift" and inserting the word "Shift" into the sheet.

I think this is simple enough and all we'll need. Now people can basically type anything as the first character in their cell, including things they'll likely do (like `$`) or things who knows why somebody might do it (like ` ` (space) or `}`)

davidkircos: The space bar triggers panning. Does this conflict?

We may want to explicitly define which chars can start a cell so we can be mindful that it doesn't conflict with other keyboard commands.

jimniels: Are there more than just `Space`? I couldn't think of any that weren't without modifier keys. So i figured it was easier to say "any single digit key _minus_ these ones" rather than say "all these single digit keys" cause there could be a lot. Excel/sheets let you type any character, why shouldn't we? Do we really want to do something like:

```
if (
  isAlphanumeric(e.key) ||
  e.key === '!' ||
  e.key === '@' ||
  e.key === '#' ||
  e.key === '$' ||
  e.key === '%' ||
  // a bunch more...
)
```

jimniels: Updated this to be any single char key except space.

davidkircos (approved):

PR #452, opened 2023-04-29
Selection auto complete multiple

davidfig:
Horizontal + Vertical auto-complete


jimniels:
When I try to delete a range of cells down to just a single row, the "delete" selection disappears. I would expect it to show as a delete if i go from `n` down to just 1 (not just 2)

May-01-2023 13-17-06

jimniels:
Perhaps related to the [above comment](https://github.com/quadratichq/quadratic/pull/452#issuecomment-1530101580)

If I drag down to just 1 row and then _do a mouseup_, the active dragging selection remains (e.g. I move my mouse around without pressing anything but the software is responding as if I am still pressing down)

May-01-2023 13-16-08

jimniels:
Dragging an empty cell over one with an existing value doesn't empty that cell.

May-01-2023 13-35-49

jimniels:
When you drag left to right (or top to bottom and bottom to top), where you end your selection becomes the terminal location for the new active selection

May-01-2023 13-38-49

However, this isn't the case when you drag right to left (for a single cell or a range of cells).

May-01-2023 13-42-08

jimniels:
FWIW: [I made a small visual tweak](https://github.com/quadratichq/quadratic/pull/452/commits/4e82e0a3b863e305b71216239951c340ee6afd7c) when deleting. I really like the solution David came up with to put a red highlight on the cells that will be deleted. However, when you delete one row but add others, it feels weird that the entire selection has a red border, like a delete operation will take place. That's not true. Both a delete and an add operation is happening.

May-01-2023 13-23-52

I tweaked this so the border always stays the blue of the cursor. And then the red highlight only occurs for cells that will be deleted.

May-01-2023 13-25-28

This feels like a more honest mapping to what is happening when you drag.

jimniels:
If you have two cells next to each other and you drag the top one _without your selection ever leaving the row you're dragging in_, it will repeat as expected:

May-01-2023 13-55-51

But if you have those same two cells and you drag the top one _while letting the selection go outside a single row_ but you still end the selection as a single row, it populates both rows

May-01-2023 13-56-06

This seems to only happen with the first two rows (if you do it on a wider selection)

May-01-2023 13-59-01

HactarCE:
Also check that formatting is copied from the original cell(s)

jimniels:
- [This bug](https://github.com/quadratichq/quadratic/pull/452#issuecomment-1530101580) is still present
- [This bug](https://github.com/quadratichq/quadratic/pull/452#issuecomment-1530103312) is still present

jimniels:
Per [Andrew's comment](https://github.com/quadratichq/quadratic/pull/452#issuecomment-1538610845) formatting isn't copied

May-08-2023 11-04-20

jimniels (approved):

jimniels:
Filed a separate [bug to get formatting working on drag](https://github.com/quadratichq/quadratic/issues/505)

PR #453, opened 2023-04-29
added delete button

davidfig:
Delete button works same as backspace to delete cells.

### DIFF
--- a/src/gridGL/interaction/keyboard/keyboardCell.ts
+++ b/src/gridGL/interaction/keyboard/keyboardCell.ts
@@ -43,7 +43,7 @@ export function keyboardCell(options: {
     event.preventDefault();
   }
 
-  if (event.key === 'Backspace') {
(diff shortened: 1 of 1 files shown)


jimniels (approved):
`fn` + `Backspace` on mac also triggers the `Delete` key.

In prod today, if I press `fn` + `Backspace` nothing happens. With this change, if I hit `fn` + `Delete` the cell now deletes. And the code looks on, so while I don't have a physical `del` key, i'm gonna go ahead and approve/merge this.

---

FWIW: `control` + `D` on mac is also [a "system" shortcut](https://support.apple.com/en-us/HT201236#:~:text=Control%2DH%3A%20Delete%20the%20character,right%20of%20the%20insertion%20point.) for deleting characters to the right of the insertion point. Sheets doesn't implement this for deleting a cell but excel does (personally, I don't think its worth it for us)

PR #454, opened 2023-04-30
Replaced Input with Div + contenteditable

### LINKED ISSUE #432: Bug: text overflow when editing neighboring cell
https://user-images.githubusercontent.com/3479421/233183335-983a5bd0-68b0-42a6-a3b2-807fff8ffb16.MOV

Chrome / MacOS

davidfig:
fixes #432


jimniels:
Changing formatting while you're editing characters in the cell works but doesn't apply to the saved cell when you hit enter.

For example, take a cell and begin editing it and hit `CMD` + `I` to make the text italic. It appears italic in the contenteditable but once you hit "ENTER" the saved value doesn't have italics applied (presumably it's only applying the italics to the contenteditable element and not the attribute of the cell itself)

May-08-2023 11-38-50

You can see this play out if you try to apply formatting to only certain parts of the text of a cell but not others

May-08-2023 11-42-15

I looked at excel (on web) to see what they do. They do appear to allow formatting to independent parts of text in a cell, though some of their formatting gets lost between editing/saving change too ha

May-08-2023 11-43-38

Sheets also allows formatting to independent pieces of text

May-08-2023 11-45-48

jimniels:
📝 Playing with this functionality in Excel and Sheets, I noticed that when you begin to edit a cell, the regular cursor highlight disappears and the cursor style outline applies around the entire box that's in focus.

For example, notice how in Sheets the little square in the bottom right disappears when you're editing (so you can't "drag/autocomplete" while editing a cell). The active cell gets highlighted with the style of the cursor but without the draggable circle.

May-08-2023 12-48-28

Same thing happens in Excel:

May-08-2023 12-52-28

In quadratic, however, as you edit, the draggable cursor stays on screen and stays fixed on the original width of the cell regardless of how much text is being entered.

May-08-2023 12-48-41

I feel like I would expect quadratic to act similar to these other apps, e.g. when a cell becomes active for editing, the "drag" cursor disappears and the _entire length of the input_ gets the focus style of the active cell. Example mockup:

CleanShot 2023-05-08 at 12 56 26@2x

src/gridGL/interaction/CellInput.tsx
```diff
@@ -21,26 +21,64 @@ export const CellInput = (props: CellInputProps) => {
...
+  const copy = sheetController.sheet.getCellAndFormatCopy(cellLocation.x, cellLocation.y);
+  const cell = copy?.cell;
+  const format = copy?.format;
+
+  const fontFamily = useMemo(() => {
+    if (!format) return 'OpenSans';
+    if (format.italic && format.bold) return 'OpenSans-BoldItalic';
+    if (format.italic) return 'OpenSans-Italic';
+    if (format.bold) return 'OpenSans-Bold';
+  }, [format]);
```
jimniels, line 39: Is putting this in `useMemo` getting us something here? Why not have `fontFamily` be a normal piece of derived state, e.g.

```js
let fontFamily = 'OpenSans';
if (format.italic && format.bold) {
  fontFamily = 'OpenSans-BoldItalic';
} else if (format.italic) {
  fontFamily = 'OpenSans-Italic';
} else if (format.bold) {
  fontFamily = 'OpenSans-Bold';
}
```

I think this is causing some kind of bug because _when you're edit a cell without a background_, the font appears as Open Sans

CleanShot 2023-05-08 at 13 04 44@2x

jimniels:
🐞 (small) there's a tiny but noticeable overflow when you edit a cell with a background that happens when you go into "edit" mode on a cell. Note how, upon entering to edit the cell, the background color overflows the width of the cell.

May-08-2023 13-09-54

src/gridGL/interaction/CellInput.tsx
davidfig: Doesn't get us very much, but I'm not sure how it causes a bug. I'll switch it, though.

jimniels:
🐞 [This comment](https://github.com/quadratichq/quadratic/pull/454#issuecomment-1538903220) seems to have been fixed in most places, but I am still seeing it in a few contexts. Like this one: this is from the default example file.

CleanShot 2023-05-16 at 13 46 30@2x

When you click to edit the cell, the background goes longer than the cell

May-16-2023 13-47-21

Honestly doesn't seem like the biggest of deals, as it's working as expected across most cells (click to edit, and it's the size of the current cell) but in this one particular case it's longer than the current cell for some reason.

jimniels:
🐞 When you add spaces to the beginning of a cell, it's like the spaces aren't the same width between the cell input and the cell on the grid (when you go to edit it, the spacing isn't the same width)

May-16-2023 13-51-24

jimniels:
🐞 If you format a cell with a background color, that background color doesn't extend the width of the cell input. It gets cut off. [Note how excel and sheets](https://github.com/quadratichq/quadratic/pull/454#issuecomment-1538880796) both have the cell input's style go the width of its contents only once you hit "Enter" and it goes into the grid does it truncate the background to the width of the cell.



That is not how Quadratic works. Quadratic cuts off the background color while you're editing the cell.

May-16-2023 14-02-34

It makes sense to me that we mimic how excel/sheets work where formatting is preserved to fit the contents of the cell input (and can truncate once the user hits 'Enter')

jimniels:
🐞 This appears to be a bug with what's in prod, so I suppose if you want to file an issue and talk about it outside the scope of this PR, that's fine by me?

Maybe the background of a cell input, if not set, should be white? That way when you have two cells next to each other whose contents overlap as you edit, the text remains legible 

May-16-2023 14-26-59

I would expect it to work like this (example from Sheets)

May-16-2023 14-28-53

jimniels:
🐞 I'm still seeing this bug where, sometimes, when I click outside of the cell input after having typed something, my cursor acts like i'm on a mousedown, so as I drag the cursor it's making a selection on the grid as I move my mouse (even though i'm not holding down the mouse).

Having a hard time reproducing it, but it's happening on and off. Will report back when I nail down a reproducible cause....

davidkircos:
Played around with this for a few minutes. It seems to work great! I noticed a small visual bug where the cursor sometimes has a gap with the background color.

https://github.com/quadratichq/quadratic/assets/3479421/28d70942-db3a-49eb-8836-b529a7062a9e

jimniels:
🐞 When you have two cells next to each other and the first one's length is longer than itself _and the one next to it_, then when you edit the one next to it, the first one's overflow becomes visible.

May-22-2023 11-58-44

That seemed like a bug to me, but I double checked excel/sheets to see how they do it and they don't allow the clipping of other cells to become visibile when you're editing a cell. So it is a bug in my mind. This is what I would expect (from Sheets):

May-22-2023 11-58-07

HactarCE:
Ctrl+scroll doesn't zoom in when my cursor is over the ContentEditable

jimniels:
Zooming while your cursor is over the cell input is doesn't appear to be specific to this PR. It's happening in prod today. We can probably file a separate issue for that apart from this chunk of work.

jimniels (changes requested):
I'm still seeing these two bugs:

1. [This comment](https://github.com/quadratichq/quadratic/pull/454#issuecomment-1557664457)
2. [This comment](https://github.com/quadratichq/quadratic/pull/454#issuecomment-1550272965)

PR #455, opened 2023-04-30
Save cursor w/undo

davidfig:
Implements #292


jimniels:
This works for the most basic case of "type something into a field, hit return, then hit `CMD` + Z`"

May-01-2023 12-09-23

But try something beyond that and it seems to break. For example, type a value into a cell, then rather than hit enter, click into a cell somewhere else, then hit `CMD` + `Z`. It will undo the value you entered but not move the cursor.

May-01-2023 12-10-33

And for more complicated actions, like typing a value, then another value somewhere else, then another value and dragging it, it just breaks completely.

https://user-images.githubusercontent.com/1316441/235503722-035bf386-277e-4aa0-a918-e035e8d729de.mp4

jimniels (approved):
Nice, clicked around a bunch and changed things and undo/redo worked as expected in each case. 👍  Love this feature!

May-08-2023 11-22-55

jimniels:
Functionality works in the app and code looks good to me, but unit tests are failing...appears to be something with [these lines](https://github.com/quadratichq/quadratic/pull/455/files#diff-36cc19b2079af0c1d520a944a10db6f032788d8babfe83a500c51cc5ae526c83R115) where you check for `this.app`

src/grid/controller/sheetController.ts
```diff
@@ -135,12 +142,16 @@ export class SheetController {
     // add reverse transaction to redo stack
     this.redo_stack.push(reverse_transaction);
 
+    this.app.settings.setInteractionState?.(transaction.cursor);
+
     // TODO: The transaction should keep track of everything that becomes dirty while executing and then just sets the correct flags on app.
     // This will be very inefficient on large files.
-    if (this.app) this.app.rebuild();
+    this.app.rebuild();
```
jimniels, line 149: Remove checks at beginning of function and move any calls to `this.app` inside an `if` statement (this should make the tests pass...)

src/grid/controller/transaction.ts
```diff
@@ -1,5 +1,7 @@
+import { GridInteractionState } from '../../atoms/gridInteractionStateAtom';
 import { Statement } from './statement';
 
 export type Transaction = {
   statements: Statement[];
+  cursor: GridInteractionState;
```
davidkircos, line 6: Might have to be optional

jimniels:
I actually just noticed that hitting "Delete" on a blank cell logs that as an event in the undo stack.

For example, in this gif, I hit the "Delete" key on a cell with a value, then I hit "Delete" on the cells below it, and then hit CMD + Z a number of times and each delete in a blank cell is an event in the stack and it moves the cursor to that cell (which feels weird, I would expect only the first "delete" that actually deleted a value to be in the stack of undos)

May-09-2023 11-38-03

jimniels:
I fixed the tests not passing in [this PR](https://github.com/quadratichq/quadratic/pull/480) which makes `this.app` optional instead of throwing an error. Feel free to merge and/or resolve in your own way.

PR #465, opened 2023-05-02
fix: close editor when creating new file

### LINKED ISSUE #462: bug: create new file from Files menu doesn't close previous editor state
If you have a file open with the editor, and you go to the files menu and create a new file, the editor remains open with the old code in it.

https://user-images.githubusercontent.com/1316441/235541936-dfc921cd-94c8-4755-b6f7-f5c1131a87a4.mp4

This same problem exists if you create a new file from the command palette.

jimniels:
Fixes #462 

Note: there are currently 3 different places that can create a new file:

1. Quadratic menu (File -> New)
2. Command palette
3. File menu (File -> Open -> Create file)

No. 2 & 3 suffer from this bug where if the code editor is open when you create a new file it doesn't disappear when the new file is opened.

### DIFF
--- a/src/hooks/useGenerateLocalFiles.tsx
+++ b/src/hooks/useGenerateLocalFiles.tsx
@@ -14,6 +14,7 @@ import { editorInteractionStateAtom } from '../atoms/editorInteractionStateAtom'
 import { DEFAULT_FILE_NAME, EXAMPLE_FILES, FILE_PARAM_KEY } from '../constants/app';
 import apiClientSingleton from '../api-client/apiClientSingleton';
 import mixpanel from 'mixpanel-browser';
+import { focusGrid } from '../helpers/focusGrid';
 const INDEX = 'file-list';
 
 export interface LocalFile {
@@ -84,6 +85,7 @@ export const useGenerateLocalFiles = (sheetController: SheetController): LocalFi
       sheetController.sheet.load_file(grid);
       sheetController.app?.rebuild();
       sheetController.app?.reset();
+      focusGrid();
       const searchParams = new URLSearchParams(window.location.search);
       // If `file` is in there from an intial page load, remove it
       if (searchParams.get('file')) {
--- a/src/hooks/useGenerateLocalFiles.tsx
+++ b/src/hooks/useGenerateLocalFiles.tsx
@@ -83,6 +83,7 @@ export const useGenerateLocalFiles = (sheetController: SheetController): LocalFi
       sheetController.clear();
       sheetController.sheet.load_file(grid);
       sheetController.app?.rebuild();
+      sheetController.app?.reset();
       const searchParams = new URLSearchParams(window.location.search);
       // If `file` is in there from an intial page load, remove it
       if (searchParams.get('file')) {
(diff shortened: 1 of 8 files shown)


src/ui/menus/FileMenu/FileMenu.tsx
```diff
@@ -66,10 +66,10 @@ export function FileMenu(props: FileMenuProps) {
       app.reset();
     }
 
-    setEditorInteractionState({
-      ...editorInteractionState,
+    setEditorInteractionState((prevState) => ({
+      ...prevState,
       showFileMenu: false,
-    });
+    }));
```
jimniels, line 72: Was simply stale state 🤦‍♂️

src/ui/menus/TopBar/SubMenus/QuadraticMenu.tsx
```diff
@@ -67,9 +67,9 @@ export const QuadraticMenu = (props: Props) => {
         <MenuDivider />
         <SubMenu label="File">
           <MenuItem
-            onClick={() => {
+            onClick={async () => {
+              await createNewFile();
```
jimniels, line 71: Make this the same order as the others: create the new file, then reset the sheet.

src/hooks/useGenerateLocalFiles.tsx
```diff
@@ -83,6 +84,8 @@ export const useGenerateLocalFiles = (sheetController: SheetController): LocalFi
       sheetController.clear();
       sheetController.sheet.load_file(grid);
       sheetController.app?.rebuild();
+      sheetController.app?.reset();
+      focusGrid();
```
jimniels, line 88: Rather than call `app.reset()` everywhere we call `createNewFile`, this now runs the reset inside the function itself, so consumers of `createNewFile` only need to call it.

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -134,9 +134,11 @@ export const CodeEditor = (props: CodeEditorProps) => {
...
 
   useEffect(() => {
-    // focus editor on show editor change
-    editorRef.current?.focus();
-    editorRef.current?.setPosition({ lineNumber: 0, column: 0 });
+    if (editorInteractionState.showCodeEditor) {
+      // focus editor on show editor change
+      editorRef.current?.focus();
+      editorRef.current?.setPosition({ lineNumber: 0, column: 0 });
+    }
```
jimniels, line 141: We don't need to set the focus on the editor if it's not visible.

jimniels:
This is fixed, however it is worth noting the follow minor bugs that still remain:

1. If you enter a cell value for `0,0` in the code editor (say `1+1`), then create a new file, then leave the cursor at `0,0` and go to edit the cell, the old value from the previous sheet (`1+1`) will still be there (if you move the cursor once you've created the new sheet, it won't)
2. If you have unsaved changes in the code editor, and create a new file, those changes will be _discarded_ with no prompt to save them. This issue is in the app today and beyond the scope of this PR.

davidkircos (approved):

PR #466, opened 2023-05-03
feat: AI ui/ux enhancements

jimniels:
This PR includes two UI/UX improvements:

1. If the user scrolls to the bottom of the AI Assistant panel, the latest response will stay pinned to the bottom as the response streams in
2. The user is provided a "Stop" button that will cancel the streaming AI chat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 ![Twitter Follow](https://img.shields.io/twitter/follow/QuadraticHQ)
 ![quadraticlogo4 1](https://user-images.githubusercontent.com/3479421/162037216-2fea1620-2310-4cfa-96fb-31299195e3a9.png)
 
-## ![quardatic icon small](https://user-images.githubusercontent.com/3479421/162039117-02f85f2c-e382-4ed8-ac39-64efab17a144.svg) **_The Data Science Spreadsheet_**
+## ![quardatic icon small](https://user-images.githubusercontent.com/3479421/162039117-02f85f2c-e382-4ed8-ac39-64efab17a144.svg) **_The data science spreadsheet_**
 
 Infinite data grid with Python, JavaScript, and SQL built-in. Data Connectors to pull in your data.
 
@@ -39,9 +39,9 @@ Quadratic has no environment to configure. The grid runs entirely in the browser
 - Quickly mix data from different sources
 - Explore your data for new insights
 
-# Getting Started
+# Getting started
 
-### Run Quadratic Locally
+### Run Quadratic locally
 
 1. Install npm, [rustup](https://www.rust-lang.org/tools/install), and [wasm-pack](https://rustwasm.github.io/wasm-pack/installer/)
 2. Run `rustup target add wasm32-unknown-unknown`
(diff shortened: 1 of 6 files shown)


jimniels:
For code blocks, it kinda stinks you don't get the code highlighting until the full code block comes in, e.g. a response might look like this:

```
This is a code sample from me, an AI assistant:

```javascript
var x = 0;
// a cod
```

Only once you get everything between the code fences ``` will it turn into an editor preview.

There's probably a way we can eagerly begin styling a code block when we see the first set of code fences by changing [this regex](https://github.com/quadratichq/quadratic/blob/main/src/ui/menus/CodeEditor/AICodeBlockParser.tsx#L3) but that is beyond the scope of this PR.

---

Note how chatGPT triggers a formatting a code response at the first sign of a code block:

block-code-edit-highlight

It does now, however, do this for inline code. It waits until the final backtick

inline-code-no-highlight

README.md
```diff
@@ -79,6 +79,6 @@ There are more example files are in the application file menu. File > Open sampl
 
 You can download them and then open them in Quadratic via File > Open Grid
 
-## Quadratic is Hiring
+## Quadratic is hiring
```
jimniels, line 82: sentence casing tweaks because i need to re-trigger a build ha

davidkircos:
CleanShot 2023-05-03 at 21 51 17@2x

Sometimes the Send tool tip get stuck on after cancelling a request

src/ui/components/TooltipHint.tsx
```diff
@@ -3,7 +3,7 @@ import { Tooltip } from '@mui/material';
 interface TooltipHintProps {
   title: string;
   shortcut?: string;
-  children: JSX.Element;
+  children: React.ReactElement<any, any>;
```
jimniels, line 6: Changed this to match what mui wants for children of `<Tooltip>`

src/ui/menus/CodeEditor/AITab.tsx
```diff
@@ -159,17 +177,37 @@ export const AITab = ({ evalResult, editorMode, editorContent }: Props) => {
...
+                  <TooltipHint title="Stop generating">
+                    <IconButton size="small" color="primary" onClick={abortPrompt} edge="end">
+                      <Stop />
+                    </IconButton>
+                  </TooltipHint>
+                ) : (
+                  <ConditionalWrapper
+                    condition={prompt.length !== 0}
+                    // @ts-expect-error
+                    Wrapper={({ children }) => <TooltipHint title="Send">{children}</TooltipHint>}
```
jimniels, line 198: I couldn't figure out how to make typescript happy for this. 

[This](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18051) was as close as I got to figuring it out, but wrapping children in a fragment `<>{children}</>` doesn't actually work with with the underlying `<Tooltip>`. So I figured best thing was just to ignore the error — though happy for any typescript advice on this.

davidkircos (approved):

PR #470, opened 2023-05-04
fix: grid settings

jimniels:
**Problem**

Currently, if you open then close the editor and go into presentation mode, the header/footer don't disappear. Similarly, if you open the code editor and trigger presentation mode, it doesn't hide everything. Only once you go into presentation mode _again_ does it then work.

May-04-2023 12-58-13

**Solution**

This problem was caused because `useGridSettings` was a hook being called across multiple components. Each use of `useGridSettings` was its own instance of the hook with it's own state and setState functions. This caused state across components to fall out of sync (even though it was all sourced from the same place: localStorage).

The solution, in this case, was to lift that state into recoil and have `useGridSettings` return _the recoil state_ as well as discrete functions for manipulation that state.

Because this changes how grid settings state is manipulated, we should be sure to test all grid settings functionality.

**To test**

- Each grid setting can turn on/off (show grid axis, show grid lines, show row/column headers, etc)
- Grid settings are preserved across file reloads
  - _Except presentation mode_ which resets on page load
- Original bug case (toggle presentation mode when code editor is open)

### DIFF
--- a/src/quadratic/QuadraticApp.tsx
+++ b/src/quadratic/QuadraticApp.tsx
@@ -7,7 +7,6 @@ import { loadAssets } from '../gridGL/loadAssets';
 import { IS_READONLY_MODE } from '../constants/app';
 import { debugSkipPythonLoad } from '../debugFlags';
 import init, { hello } from 'quadratic-core';
-import { useGridSettings } from '../ui/menus/TopBar/SubMenus/useGridSettings';
 import { SheetController } from '../grid/controller/sheetController';
 import { useGenerateLocalFiles } from '../hooks/useGenerateLocalFiles';
 import { PixiApp } from '../gridGL/pixiApp/PixiApp';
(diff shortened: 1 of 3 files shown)


src/quadratic/QuadraticApp.tsx
```diff
@@ -32,14 +29,6 @@ export const QuadraticApp = () => {
...
   }, [itemsLoaded]);
 
-  // reset presentation mode when app starts
-  useEffect(() => {
-    if (!settingsReset) {
-      if (presentationMode) setPresentationMode(false);
-      setSettingsReset(true);
-    }
-  }, [presentationMode, setPresentationMode, settingsReset, setSettingsReset]);
-
```
jimniels, line 42: Because this is state in recoil now, we can [colocate this logic](https://github.com/quadratichq/quadratic/pull/470/files#diff-0271748ba2e6a81a84d93762b4c3e39e29c07eb13ca55b567f827ca7aea782aaR31) with how/when the state is set in our recoil file. No need for a side effect in react.

src/ui/menus/TopBar/SubMenus/useGridSettings.ts
```diff
@@ -20,6 +21,40 @@ export const defaultGridSettings: GridSettings = {
   presentationMode: false,
 };
 
+// Persist the GrdiSettings
```
davidkircos, line 24: GridSettings*

davidkircos (approved):

PR #480, opened 2023-05-09
fix: make this.app conditional

jimniels:
so the tests pass


src/grid/controller/sheetController.ts
```diff
@@ -142,16 +145,18 @@ export class SheetController {
     // add reverse transaction to redo stack
     this.redo_stack.push(reverse_transaction);
 
-    this.app.settings.setInteractionState?.(transaction.cursor);
+    if (this.app) {
```
davidkircos, line 148: Probably good to add a comment that explains what this does

PR #491, opened 2023-05-12
fix: support null in array_output file schema

### LINKED ISSUE #472: bug: cells() function returns "NaN" for empty cells and corrupts file
CleanShot 2023-05-05 at 09 08 04@2x

This is causing file corruption. If you close the file, open another one and make an edit, then try to re-open this one, you'll get an error saying it failed to load the file.

Looking at what zod schema checks, this is what i see:

CleanShot 2023-05-05 at 09 11 51@2x

It looks like this is because `null` is being stuck as a value in `array_output` and the schema only allows string, number, or boolean. If you download the file that won't open in the app and open it in a text editor, you can see the lines where it breaks:

CleanShot 2023-05-05 at 09 11 07@2x

Removing `null` and replacing it with, say, an empty string `""` allows you to re-upload the file through the GUI. It will load the sheet and you'll see the row of numbers and "NaN".

CleanShot 2023-05-05 at 09 18 26@2x

If you re-run the cell, then open another file, then open this one again, it will be corrupted again.

jimniels:
Fixes #472 by supporting `null` as a valid type for `array_output`


src/grid/computations/types.ts
```diff
@@ -1,17 +1,15 @@
 import z from 'zod';
-
-const ArrayOutputSchema = z.array(z.union([z.string(), z.number(), z.boolean()]));
```
jimniels, line 3: This is what [`ArrayOutputBaseSchema`](https://github.com/quadratichq/quadratic/pull/491/files#diff-63372a6cc930c497114d0f3d293149da3814ca1f2fdcf8d67f624cd41070273eR7) is in the base schema file, and then we're duplicate what we already do in the file schema [here](https://github.com/quadratichq/quadratic/pull/491/files#diff-786b8eb1d8f122847835f3faa2451f8d2e9ca3401a219c1f6dfca5c02d49db5aL11)

So this is essentially just pulling these types directly from the base file schema, as they are the same!

src/grid/computations/python/runPython.ts
```diff
@@ -1,10 +1,12 @@
+import { ArrayOutput } from '../../../schemas';
+
 export interface runPythonReturnType {
   cells_accessed: [number, number][];
   success: boolean;
   input_python_stack_trace: string;
   input_python_std_out: string;
   output_value: string | null;
-  array_output: (string | number | boolean)[][];
```
jimniels, line 7: I think this type was wrong! It was saying a 2d array of string, number or bool. But even [the previous type](https://github.com/quadratichq/quadratic/pull/491/files#diff-786b8eb1d8f122847835f3faa2451f8d2e9ca3401a219c1f6dfca5c02d49db5aL11) said it could be a 1 or 2d array. So we're updating this to match what's in the file schema: 1 or 2d array of string, number, bool, or null.

However! This doesn't say `array_output` can be optional, so we unwrap it

Although maybe this one isn't optional? Perhaps we need to `unwrap()` the value here since it wasn't optional before. Setting it as `ArrayOutput` makes its type:

`(string | number | boolean | null)[] | (string | number | boolean | null)[][] | undefined`

src/grid/actions/updateCellAndDCells.ts
```diff
@@ -120,13 +119,15 @@ export const updateCellAndDCells = async (args: ArgsType) => {
...
+        // result.array_output = [[], []]
         if (result.array_output !== undefined && result.array_output.length > 0) {
           if (Array.isArray(result.array_output[0])) {
             // 2d array
             let y_offset = 0;
             for (const row of result.array_output) {
+              // [[], []]
               let x_offset = 0;
-              for (const cell of row as ArrayOutput) {
+              for (const cell of row as ArrayOutputBase) {
```
jimniels, line 130: This still fails because `cell.toString()` can't be run if we specify cell to be a string, number, boolean, or null. Unsure yet how we should handle this....

CleanShot 2023-05-18 at 16 36 44@2x

src/grid/actions/updateCellAndDCells.ts
```diff
@@ -1,9 +1,8 @@
-import { Cell } from '../../schemas';
+import { Cell, ArrayOutputBase } from '../../schemas';
```
jimniels, line 1: I don't love this approach, but it seems the best way to handle this case since typescript can't infer what the code is doing below, especially when our type is a 1 or 2d array of the same values

src/grid/actions/updateCellAndDCells.ts
jimniels: This actually isn't failing anymore...

davidkircos (approved):

PR #494, opened 2023-05-13
Delete empty cells

### LINKED ISSUE #481: bug: delete on an empty cell shouldn't be in the undo/redo stack
Hitting "Delete" on a blank cell logs that as an event in the undo stack.

For example, in this gif, I hit the "Delete" key on a cell with a value, then I hit "Delete" on the cells below it, and then hit CMD + Z a number of times and each delete in a blank cell is an event in the stack and it moves the cursor to that cell (which feels weird, I would expect only the first "delete" that actually deleted a value to be in the stack of undos)

May-09-2023 11-38-03

davidfig:
Delete does not do anything if cells are empty. Fixes #481


jimniels (commented):
This works great for small situations.

But, per [David's comment in Slack](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1684171087984099?thread_ts=1683987196.510299&cid=C03VC0Q1MPA), it does appear to break when you do it for a big operation.

For example, open the the "Airports distance" example file and try doing select all / delete. On this branch, the page just freezes totally and the tab becomes unresponsive, triggering a dialog fro the browser:

CleanShot 2023-05-16 at 15 27 49@2x

If you leave it long enough, it'll eventually do it's thing and go back to being responsive.

This stands in contrast to what's in prod today. In prod, if you open "Airports distance" and select all / delete, it'll be laggy and you'll see things go row by row, but the tab won't freeze and trigger the browser warning.

May-16-2023 15-29-24

Can / should we fix this so it doesn't become totally unresponsive? That feels like a regression to me (even though what we have today isn't the best)

PR #497, opened 2023-05-15
Implement `ABS`, `PI`, `TAU`, and trigonometric functions

HactarCE:


davidkircos (approved):
In Excel SQRT(-1) returns an error. In Quadratic it returns NaN.
PI() in Quadratic does one more decimal of precision than Excel and GSheets. 3.14159265358979000000 
 vs 3.14159265358979300000

The trig functions appear to function correctly. Although I did not test every single one.

HactarCE:
The `NaN` vs. error thing is broader than just `SQRT` -- I'll open a separate issue for that.

Regarding precision, this is a matter of how values are printed. Excel will gladly show more decimals if you push it a little, but they're meaningless because anything after the last `9793` is just floating-point noise.
image

Modern float printers are smart enough to print the _minimum number of digits needed to uniquely identify the float_, whereas I suspect Excel could be printing the _number of digits it's confident in_ (or one less).

davidkircos:
That makes sense. You're good to merge!

PR #501, opened 2023-05-16
Add regression test for #499

HactarCE:


davidkircos (approved):
This looks like it will work.

@HactarCE did you verify this test fails when that bug is present?

HactarCE:
Yes, I did :+1:

PR #507, opened 2023-05-17
Ignore empty values in variadic formula functions (`SUM`, `PRODUCT`, `COUNT`, and `AVERAGE`)

HactarCE:
Previously blank cells and empty strings would count empty strings as `0` and error on non-numeric values. This is incompatible with Excel, which ignores empty strings and other non-numeric values for these functions.

For some functions, like the basic operators `+`, `*`, etc., an empty string is counted as `0` (and that still applies here)

I also added a bunch of tests for these functions

Examples of new behavior:
- If `A1` is an empty cell, then `3 * A1` should be `0` but `PRODUCT(3, A1)` should be `3`.
- `COUNT()` should count zeros, but not count empty or non-numeric cells. So if `A1:A3` contains a blank cell, the number `0`, and the text `hello`, then `COUNT(A1:A3)` should be `1`.
- `AVERAGE(3, A1)` should be `1.5` if `A1` contains `0`, but `3` if `A1` is blank.


jimniels (approved):
Functionality checks out for me 👍

PR #529, opened 2023-05-22
Improve performance update cells

### LINKED ISSUE #528: Improve performance of updateCellAndDCells
Currently the function is using an array to store the cells to update, it would be more efficient to use a set.

juanbercoff:
- fixes #528

I've tested the difference and it improved from 9777ms to 480ms updating text from 100*27 grid

### DIFF
--- a/src/grid/actions/updateCellAndDCells.ts
+++ b/src/grid/actions/updateCellAndDCells.ts
@@ -14,6 +14,18 @@ interface ArgsType {
   create_transaction?: boolean;
 }
 
+type StringId = `${string},${string}`;
+
+function getCoordinatesFromStringId(stringId: StringId): [number, number] {
+  // required for type inference
+  const [x, y] = stringId.split(',').map((val) => parseInt(val));
+  return [x, y];
+}
+
+function addToSet(deps: [number, number][], set: Set<StringId>) {
+  for (const dep of deps) set.add(`${dep[0]},${dep[1]}`);
+}
+
 export const updateCellAndDCells = async (args: ArgsType) => {
   const { starting_cells, sheetController, app, pyodide, delete_starting_cells, create_transaction } = args;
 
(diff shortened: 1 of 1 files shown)


src/grid/actions/updateCellAndDCells.ts
```diff
@@ -14,6 +14,18 @@ interface ArgsType {
...
+type StringId = `${string},${string}`;
+
+function getCoordinatesFromStringId(stringId: StringId): [number, number] {
+  // required for type inference
+  const [x, y] = stringId.split(',').map((val) => parseInt(val));
+  return [x, y];
+}
+
+function addToSet(deps: [number, number][], set: Set<StringId>) {
+  for (const dep of deps) set.add(`${dep[0]},${dep[1]}`);
```
davidkircos, line 26: there is this function `getKey` defined a few times in the codebase that does this same thing. Convering coordinates to a string.

Maybe it's worth standardizing a set of functions in one place?

src/grid/actions/updateCellAndDCells.ts
```diff
@@ -24,43 +36,29 @@ export const updateCellAndDCells = async (args: ArgsType) => {
   const updatedCells: Coordinate[] = [];
 
   // start with a plan to just update the current cells
-  let cells_to_update: [number, number][] = starting_cells.map((c) => [c.x, c.y]);
+  const cells_to_update: Set<StringId> = new Set(starting_cells.map((c) => `${c.x},${c.y}` as const));
```
davidkircos, line 39: could use a standard function to convert x,y to string

davidkircos (commented):
Using a Set instead of a array makes sense.

juanbercoff:
I used performance.now, it also feels a bit snappier, at least in my old pc haha

davidkircos:
> I used performance.now, it also feels a bit snappier, at least in my old pc haha

@juanbercoff 
I am curious about the file and data you used to test.  Could you export and share your `.grid` file?

juanbercoff:
I used an empty file, wrote something on cell 0,0 and drag it to cell 100,27

davidkircos:
Oh wow, that is 10x faster on my computer too.

Let me know what you think about my comments on the code. This seems like a great improvement.

src/grid/actions/updateCellAndDCells.ts
juanbercoff: Created new function in helpers

davidkircos (approved):

PR #530, opened 2023-05-22
feat: Quadratic Cloud Files V1

jimniels:
Because this PR is such a fundamental change, _everything_ about the app should be tested, including:

- Authentication flow (log in, log out)
- Authenticated & non-authenticated users only have access to appropriate routes
- Cloud files migration
- Cross browser, including mobile/tablet
- TODO list of things that we should ensure we test

Known outstanding issues:

- [x] Sync/debounce/cancel file sync requests
- [x] Refactor how analytics load so we can use them in loaders #637 
- [x] Update [the instructions for how people can run the app locally](https://github.com/quadratichq/quadratic/pull/530/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055), e.g. env variables, client and server tasks, etc.
- [x] Back button after logging in ([won't fix](https://github.com/quadratichq/quadratic/pull/530/files#r1290704529))
- [x] Support "open in new tab" for file list items (e.g. example file)
- [x] Comment out light/dark mode toggle
- [x] Consider removing rate limiting (delete lots of files will trigger it)
- [x] Cross browser testing (mobile too)
- [x] Fix TODOs in the current PR
- [x] Docker for development (David)
- [x] Make all routes (not defined in the root routes component) as string constants for use elsewhere in the app
- [x] Social share/open graph meta info in `index.html` #636 

David - Review
- [x] review the code
- [x] review files and upgrade path
- [x] test the upgrade flow

Deployment Review
- [x] migrate staging (test prisma migration flow) (edi


quadratic-api/src/routes/files.ts
```diff
@@ -65,8 +72,70 @@ files_router.post('/backup', validateAccessToken, ai_rate_limiter, async (reques
...
+files_router.get('/:uuid', validateAccessToken, ai_rate_limiter, async (request: JWTRequest, response) => {
+  console.time('GET /files/:uuid');
+
+  // Validate request format
+  const schema = z.string().uuid();
+  const result = schema.safeParse(request.params.uuid);
+  if (!result.success) {
+    console.log(result.error.issues);
+    return response.status(400).json({ message: 'Invalid file UUID.' });
+  }
```
jimniels, line 105: @davidkircos after we validate the request, we're going to need to change this to some kind of logic like this:

1. Fetch the file and see if it exists
2. If it exists, is it public? Return it
3. If it exists, but it's not public, see if the current user is the owner and return it
4. Otherwise return a 4xx

src/gridGL/pixiApp/PixiApp.ts
```diff
@@ -52,11 +51,10 @@ export class PixiApp {
   // for testing purposes
   debug: Graphics;
 
-  constructor(sheet_controller: SheetController, save: () => Promise<void>) {
+  constructor(sheet_controller: SheetController) {
     this.sheet_controller = sheet_controller;
     this.sheet = sheet_controller.sheet;
     this.sheet.onRebuild = this.rebuild;
-    this.save = save;
```
jimniels, line 59: I went ahead and remove all this.

From what I can tell, creating a new instance of the sheet controller doesn't require passing in a `save` function. It never gets called anywhere.

Instead we have [this logic](https://github.com/quadratichq/quadratic/pull/530/files#diff-36cc19b2079af0c1d520a944a10db6f032788d8babfe83a500c51cc5ae526c83R10-R99)

quadratic-api/src/routes/files/files.ts
```diff
@@ -0,0 +1,281 @@
...
+      const contents = Buffer.from(req.body.contents, 'utf-8');
+      await dbClient.file.update({
+        where: {
+          uuid: req.params.uuid,
+        },
+        data: {
+          contents,
+          updated_date: new Date(),
+          // TODO this needs to be pulled from body.contents if that is posted
+          version: 'unknown',
```
jimniels, line 168: @davidkircos this information _should be_ in the contents, not sure if there's a reason to be setting here to `unknown`?

Similarly, [line 185](https://github.com/quadratichq/quadratic/pull/530/files#diff-49043b2e494fcfc12ea5a1846c4b2325bd4f187dc4dbf3dffcc248f7c82726e2R185) has `version` being changed to `unknown` but i feel like that just shouldn't be changing at all. Changing the file's name shouldn't change the version.

davidkircos: I didn't know how to read the version from the contents because we changed the format. Maybe we should make ti an extra parameter?

jimniels: Hm....i don't think it should be an extra parameter. it's so intimately tied to the `contents` i think you should just pull it from there. like when would we ever change the version but not the contents?

in this case, if you did `JSON.parse(contents).version` that should give you what you want. i can't imagine that changing anytime in the near future...

davidkircos:
Organization notes:

It seems to me that the `routes` folder should just really contain the routes. 
It would be good to organize everything FileBrowser or Dashboard (we should name it) in one place. Currently file, account, examples, and mine are thrown into the routes folder. 
While the dashboard header and layout are located in another folder entirely.

src/routes/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+            await authClient.handleSigninRedirect();
+            let isAuthenticated = await authClient.isAuthenticated();
+            if (isAuthenticated) {
+              let redirectTo = new URLSearchParams(window.location.search).get('redirectTo') || '/';
+              return redirect(redirectTo);
+            }
+          } catch (e) {
+            console.error(e);
+          }
+          return redirect('/');
```
jimniels, line 146: Not a huge deal, but there's a little bit of weird ui/x as it relates to the "Back" button on this flow.

If you hit `/` from a fresh tab and aren't logged in, you'll be redirected to auth0's website and see two items immediately in your history stack: 

1. `/`
2. `/login?from=/`

Once you complete the stuff on on auth0's page, you'll have four items in the history stack (and you'll be logged in to the app):

1. `/` - initial login
2. `/login?from/` - initial login redirect cause you're not auth'd
3. `auth0.com` - auth0 login page
4. `/login-result?redirectTo=...&code=...&state` - where auth0 redirects to

This can result in a bit of strange behavior if you log in then immediately hit the back button. Not a huge deal, but I'm not sure there's a lot we can do about it.

From what I understand, react-router does a `push` when you have a fresh new tab (rather than a redirect) so we'd have to file an issue with them to resolve.

src/api-client/types.ts
```diff
@@ -0,0 +1,56 @@
...
+  )
+  .optional();
+export type GetFilesRes = z.infer<typeof GetFilesResSchema>;
+
+// GET /file/:uuid
+export const GetFileResSchema = z.object({
+  file: z.object({
+    ...fileMeta,
+    contents: z.string(), // Stringified Gridfile
+    version: z.string(), // TODO one of: ...
```
davidkircos, line 29: Todos still here

src/components/Theme.tsx
```diff
@@ -0,0 +1,63 @@
...
+  let mode: PaletteMode = colorModePreference === 'system' ? (prefersDarkMode ? 'dark' : 'light') : colorModePreference;
+
+  const theme = useMemo(
+    () =>
+      createTheme({
+        palette: {
+          mode,
+        },
+        typography: {
+          fontFamily: `OpenSans, sans-serif`,
```
davidkircos, line 37: Should this just default to system fonts?

src/constants/app.ts
```diff
@@ -1,52 +1,47 @@
...
-  {
+  'startup_portfolio.grid': {
     name: 'Startup portfolio (example)',
-    file: 'startup_portfolio.grid',
     description: 'Example with calculations from formulas and Python.',
   },
-];
-export const CSV_IMPORT_MESSAGE = 'Drag and drop a CSV file on the grid to import it.';
+};
+export const CSV_IMPORT_MESSAGE = 'Drag and drop a CSV id on the grid to import it.';
```
davidkircos, line 46: I'm confused by this message change. Was it intentional?

src/dashboard/FileRoute.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+    throw new Response('Invalid file that could not be upgraded.', { status: 400 });
+  }
+
+  // If the file version is newer than what is supported by the current version
+  // of the app, do a (hard) reload.
+  if (contents.version > GridFileSchema.shape.version.value) {
+    // @ts-expect-error
+    window.location.reload(true);
+  }
```
davidkircos, line 47: This could potentially create an infinite loop. Should we tell the user the issue and let them decide to reload manually?

src/dashboard/components/Layout.tsx
```diff
@@ -0,0 +1,243 @@
...
+import { NavLink, Outlet, useLocation, useNavigation, useRouteLoaderData } from 'react-router-dom';
+import { ROUTES } from '../../constants/routes';
+import { RootLoaderData } from '../../router';
+import { colors } from '../../theme/colors';
+import { ReactComponent as QuadraticLogo } from './quadratic-logo.svg';
+import { ReactComponent as QuadraticLogotype } from './quadratic-logotype.svg';
+
+const drawerWidth = 264;
+
+export const Component = () => {
```
davidkircos, line 13: This name is too generic

src/dashboard/components/Layout.tsx
```diff
@@ -0,0 +1,243 @@
...
+            <Menu />
+          </IconButton>
+        </Box>
+        <Outlet />
+      </Box>
+    </Box>
+  );
+};
+
+function Navbar({ handleDrawerToggle }: { handleDrawerToggle: Function }) {
```
davidkircos, line 102: should components be in their own files?

src/dashboard/components/Layout.tsx
davidkircos: Oh, is this a router Route component? Maybe rename file to LayoutRoute? I'm still a bit confused by the convention here

src/debugFlags.ts
```diff
@@ -65,3 +65,9 @@ export const debugShowCellsForDirtyQuadrants = debug && false;
 export const debugShowFileIO = debug && false;
 
 export const debugGridSettings = debug && false;
+
+// --------
+// Auth
+// --------
+
+export const debugLogAuth = true;
```
davidkircos, line 73: `debug && true` ?

src/router.tsx
```diff
@@ -0,0 +1,190 @@
...
+          if (debugLogAuth) console.log('[auth] / <loader>: isAuthenticated: %s', isAuthenticated);
+          initializeAnalytics({ isAuthenticated, user });
+
+          return { isAuthenticated, user };
+        })}
+        element={<Root />}
+        errorElement={<RootError />}
+        id="root"
+      >
+        <Route index element={<Navigate to="/files" replace />} />
```
davidkircos, line 62: use routeConstants

src/router.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+        <Route path="file">
+          {/* Check that the browser is supported _before_ we try to load anything from the API */}
+          <Route element={<BrowserCompatibility />}>
+            <Route index element={<Navigate to="/files" replace />} />
```
davidkircos, line 67: use ROUTES

src/router.tsx
```diff
@@ -0,0 +1,190 @@
...
+            <Route path=":uuid" lazy={() => import('./dashboard/FileRoute')} />
+          </Route>
+        </Route>
+
+        <Route
+          path="files/create"
```
davidkircos, line 73: use ROUTES

src/router.tsx
```diff
@@ -0,0 +1,190 @@
...
+          id="create"
+          loader={Create.loader}
+          action={Create.action}
+          shouldRevalidate={() => false}
+        />
+
+        <Route lazy={() => import('./dashboard/components/Layout')}>
+          <Route path="files" element={<Navigate to={ROUTES.MY_FILES} replace />} />
```
davidkircos, line 81: use ROUTES

src/ui/QuadraticApp.tsx
```diff
@@ -1,34 +1,25 @@
...
 import { webWorkers } from '../web-workers/webWorkers';
-import { useSetRecoilState } from 'recoil';
-import { loadedStateAtom } from '../atoms/loadedStateAtom';
+import QuadraticUIContext from './QuadraticUIContext';
+import { QuadraticLoading } from './loading/QuadraticLoading';
 
-type loadableItem = 'pixi-assets' | 'local-files' | 'wasm-rust' | 'quadrants';
-const ITEMS_TO_LOAD: loadableItem[] = ['pixi-assets', 'local-files', 'wasm-rust', 'quadrants'];
+type loadableItem = 'pixi-assets' | 'wasm-rust';
+const ITEMS_TO_LOAD: loadableItem[] = ['pixi-assets', 'wasm-rust'];
```
davidkircos, line 14: I think we still need to prerender quadrants?

src/ui/contexts/File.tsx
davidkircos, line 1: Rename to FileContext?

src/dashboard/FileRoute.tsx
jimniels: Logged the event in sentry and we'll at least see tons of events if it gets stuck in a loop. Seems good enough for now.

src/dashboard/components/Layout.tsx
jimniels: Yeah it has to be named `Component` as that's the convention (that's the name react-router looks for).

The way this component is used in React Router, it's called a "Layout Route". So really, it's a "layout route for the dashboard" so the most descriptive name would probably be `DashboardLayoutRoute`. Maybe i'll change the components inside of `/dashboard` to be more descriptive of that, e.g. `Header` -> `DashboardHeader` etc.

src/components/Theme.tsx
```diff
@@ -0,0 +1,64 @@
...
+        palette: {
+          mode,
+        },
+        typography: {
+          // We set a font family on the body in the main `index.css` file
+          // so we just inherit from that.
+          fontFamily: `inherit`,
+          // For some reason, if you change the default font-family, this gets
+          // lost on the original overline style
+          overline: {
```
davidkircos, line 41: @jimniels is this code still needed?

Issue #518, opened 2023-05-19
Array dimensions should expand independently in formulas


HactarCE:
For example, `A2:A4 * B1:D1` results in a 3x3 array (in this case, it does matrix multiplication).

PR #532, opened 2023-05-23
Refactor formula functions and values

HactarCE:
This refactor will enable better Excel compatibility in formula evaluation:

## Arrays

- [x] Independent horizontal & vertical expansion (`{1,2,3} + {10;20;30}` returns a 3x3 array)
- [x] `SUM()`, `PRODUCT()`, etc. error on invalid single values (`SUM(1, "text")` returns an error)
- [x] `SUM()`, `PRODUCT()`, etc. ignore invalid values in arrays (`SUM(1, {"text"})` returns `1`)
- [x] Error values are propagated, even in arrays (`SUM(1, {1/0})` returns an error)
- [x] Cell references are 1x1 arrays (if A1=`"text"`, `SUM(1, A1)` returns `1`)
- [x] 1x1 arrays returned from operations are converted to single values (if A1 and A2=`"text"`, `SUM(1, A1&A2)` returns an error)

## NaN and Inf

- [x] Division by zero produces a nice pretty error message
- [x] NaN and Inf in calculations are converted to proper errors

## Nulls (possible future behavior)

This is untested for now since we need a more solid type system to make it useful

- [ ] Empty cells contain `null`
- [ ] `null` is coerced to zero when used numerically (if A1 is null, then `A1+2` returns `2`)
- [ ] Empty strings do not coerce to number (if A1 contains a formula producing an empty string, then `A1+2` returns an error)
- [ ] `null` is coerced to an empty string when used as text (if A1 is null, then `A1+"text"` returns `"text"`)


HactarCE:
This also fixes #518

davidkircos (approved):
Looks good to me.

PR #533, opened 2023-05-23
Horizontal scroll

### LINKED ISSUE #177: Horizontal scrolling using Shift+scroll

juanbercoff:
- fix #177 
Extracted keys to array
Added horizontal scroll keys
If both keys  types are pressed no action is taken


juanbercoff:
> Figma seems to disable this behavior on MacOS, but it exists on Windows and Linux.
> 
> This is probably what we should do as well.

Done in last commit

src/gridGL/pixiOverride/Wheel.ts
```diff
@@ -293,13 +313,14 @@ export class Wheel extends Plugin {
     } else if (e.ctrlKey && this.options.trackpadPinch) {
       this.pinch(e);
     } else {
+      const isMac = navigator.userAgent.search('Mac') !== -1;
```
davidkircos, line 316: We have a function for this in `quadratic/src/utils/isMac.ts`, could you switch to using that function @juanbercoff

davidkircos (changes requested):

HactarCE:
Working well on Windows!

src/gridGL/pixiOverride/Wheel.ts
juanbercoff: Done!

PR #545, opened 2023-05-27
Improve selection and misc. performance

davidfig:


davidkircos:
@davidfig This feels just as laggy as before in a large file zoomed out.

Try the example Airports distance fully zoomed out

davidkircos:
It's strange. The cursor is pretty fast near the origin. However, if you open Airports Large and zoom out enough to where the Context Menu disappears. Then move the cursor around far from the origin (but keep some data in view) the cursor becomes very laggy.

jimniels (changes requested):
This seems to introduce a problem where, if you load a file (a fresh load in the browser), and then click somewhere on the sheet to move the cursor, data on the screen disappears. Then once you pan, it reappears.

https://github.com/quadratichq/quadratic/assets/1316441/4514965a-5c29-49b1-a786-6e7beaf3aae3

Issue #301, opened 2023-02-21
bug: paste doesn't doesn't work in safari/firefox


jimniels:
In Safari/Firefox, you can cut/copy within Quadratic but when you try to paste nothing happens (nothing in the console).

You can tell it is cutting (you see it disappear) and copying because when you cut/copy, the content ends up on your computer’s clipboard. But when you try to paste, nothing happens.

![copy-paste](https://user-images.githubusercontent.com/1316441/220456128-c45cdf58-bbb5-4ec3-a180-934da94d358e.gif)

davidkircos:
This is also an issue in Microsoft Edge

jimniels:
Cut/paste works in safari now, but it doesn't preserve formatting

![May-17-2023 13-35-01](https://github.com/quadratichq/quadratic/assets/1316441/9e141114-1e34-4e6d-aa90-d463b1588c7f)

We might want to create a high-level ticket to track feature parity across browsers, e.g. "these are features which work in chrome but not yet in other browsers (fully): 1), 2), 3)"

PR #547, opened 2023-05-27
wip fix paste does not work in firefox

juanbercoff:
- #301 

This change use an event based approach instead of using the Clipboard API. This allows copying/pasting to be browser independent so now works on firefox/edge. 

It's WIP, if you like the approach I could polish the code and  handle some edge cases I noticed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,11 +15,20 @@ If you have any problems getting the project to run locally, please create an is
 3. Install [wasm-pack](https://rustwasm.github.io/wasm-pack/installer/)
 4. `rustup target add wasm32-unknown-unknown` to install the WASM toolchain
 
-## Run Quadratic locally
+## Run Quadratic front-end locally
 
 1. `npm run build:wasm` to compile the Rust code
 2. `npm install` to install dependencies (run again when updating Rust)
 3. `npm start` to run in browser or `npm run dev` to run with Electron
+
+### Run Quadrtic server locally
+
+1. `cd quadratic-api`
+2. `npm i`
+3. Install [postgress.app](https://postgresapp.com/) (follow instructions on website)
+4. Configure `./.env` & `./quadratic-api/.env` values, including update to `DATABASE_ENV` variable to point at your local postgres db.
+5. `npm run prisma:migrate`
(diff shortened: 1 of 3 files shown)


src/gridGL/QuadraticGrid.tsx
```diff
@@ -9,18 +9,28 @@ import { CellInput } from './interaction/CellInput';
...
 import { PanMode } from '../atoms/gridInteractionStateAtom';
+import { ClipboardData, generateClipboardStrings, pasteFromTextOrHtml } from '../grid/actions/clipboard/clipboard';
+import { Coordinate } from './types/size';
 
 interface IProps {
   sheetController: SheetController;
   app: PixiApp;
 }
 
+export type CopiedData = {
```
davidkircos, line 20: move type declaration to clipboard file

src/gridGL/QuadraticGrid.tsx
```diff
@@ -105,6 +115,42 @@ export default function QuadraticGrid(props: IProps) {
     app: props.app,
   });
 
+  useEffect(() => {
```
davidkircos, line 118: This approach seems to be more robust

davidkircos:
@juanbercoff I like this approach. However, because the clipboard is kept in the sate, it will not work across sessions and browser tabs, will it?

juanbercoff:
> @juanbercoff I like this approach. However, because the clipboard is kept in the state, it will not work across sessions and browser tabs for formatted cells, will it?

Yes, you are right. You can solve that using localforage, but I don't think it's worth it. A huge downside is that you can't copy structured content cross-browser/application. With the current implementation that's working.

The main issue with Firefox is that the Clipboard API is not fully supported. I'll try to find a workaround.

davidkircos:
Maybe the best approach is an optimistic one? If we are on a supported
platform - then use the Clipboard API. If we are not - use your current
approach which should work on all platforms.

davidkircos:
In Figma you can copy and paste structured content from the website version
to the electron version of the app. If the user is on Chrome we can do this
with the current approach.

juanbercoff:
Ok, that should work.

PR #549, opened 2023-05-31
fix paste does not work on firefox

juanbercoff:
Related comments #547 
- #301 
Replaced document.execCommand('copy') fallback with localforage

I have yet to test the issue on safari


juanbercoff:
Fixed not pasting format on safari. This issue was caused by safari when writing the clipboard, it added some style that broke the regex.

Get the metadata directly from the attribute to avoid the regex.

src/grid/actions/clipboard/clipboard.ts
```diff
@@ -210,7 +210,7 @@ export const copyToClipboard = async (sheet_controller: SheetController, cell0:
   );
   const quadraticString = btoa(String.fromCharCode(...quadraticData));
 
-  const clipboardHTMLString = `<span data-metadata="<--(quadratic)${quadraticString}(/quadratic)-->"></span>${htmlClipboardString}`;
+  const clipboardHTMLString = `<span id="quadratic-data" data-metadata="${quadraticString}"></span>${htmlClipboardString}`;
```
davidkircos, line 213: I copied this implementation from Fimga.

I am not sure why they wrap the metadata in a `<--(name) (/name)-->` if there is some good reason or not.

juanbercoff: I guess because it easies matching the expression with a regex.

davidkircos:
fixed

PR #553, opened 2023-06-03
properly resize negative rows 

davidfig:


davidkircos (approved):

davidkircos:
@davidfig This works great.

FYI - I noticed there is no way to resize the -1st row. Where you would drag the origin.

PR #554, opened 2023-06-04
Clip when cell has error

davidfig:


davidkircos:
This seems to clip all cells at the width of the column

PR #555, opened 2023-06-06
Allow omitting arguments in formula functions

HactarCE:
For example, `AVERAGE(1,,2)` contains a blank argument in the middle, which is coerced to zero, so that function call returns `1`.

Other examples of blank/omitted arguments:
- one blank: `AVERAGE(1,2,)`, `AVERAGE(,1,2)`
- two blank: `AVERAGE(,,3)`, `AVERAGE(,3,)`, `AVERAGE(3,,)`
- three blank: `AVERAGE(,,)`

Note that `AVERAGE()` contains no arguments, but `AVERAGE(,)` contains two blank arguments.


davidkircos:
When returning a blank argument, we may want to convert it to something other than the string`[blank]`
CleanShot 2023-06-21 at 15 40 49@2x

davidkircos (approved):

PR #559, opened 2023-06-09
Highlight cells referenced in code

### LINKED ISSUE #47: Highlight Cells Referenced in Code
When code references cells, highlight those cells and the code in the same color so it's clear what is being referenced.

Screen Shot 2022-04-14 at 10 50 26 AM

juanbercoff:
- resolves #47


jimniels:
I just approved this so I could play with the preview branch...wow, this is pretty neat @juanbercoff!

I'm playing with it / browsing the code ATM to see how you approached it...looks like you only implemented it for formulas. I'm curious: how difficult would it be to leverage the same logic to add it for python cells too that use the `cell()` or `cells()` functions?

davidkircos:
> [](/apps/github-actions) ** Lint Check / lint (pull_reque**

I think if we did this with Python, we would only want it to work for function calls with constants as the cell reference. For example, if they used a variable for a cell reference, we shouldn't try to highlight it.

juanbercoff:
Sure, I'll work on it.

davidkircos:
@juanbercoff just spent some more time testing this PR with formulas. The feature works great!

src/ui/menus/CodeEditor/FormulaLanguageModel.ts
```diff
@@ -50,8 +53,11 @@ export const FormulaTokenizerConfig = {
   ],
   tokenizer: {
     root: [
-      // cell references
-      [/\$?n?[A-Z]+\$?n?\d+/, 'constant.other.cell'],
+      // cell refence multicursor
+      [/\$?n?[A-Z]+\$?n?\d+:\$?n?[A-Z]+\$?n?\d+/, CELL_REFERENCE_MULTICURSOR],
+
+      // cell reference
+      [/\$?n?[A-Z]+\$?n?\d+/, CELL_REFERENCE],
```
davidkircos, line 60: @HactarCE does this REGEX look correct to you for matching a cell reference and a multi cell reference?

HactarCE: The regex itself looks correct, but it shouldn't be necessary. `parse_formula()` from `quadratic-core` returns a list of cell references from the same parser that will run the formula.

src/helpers/formulaNotation.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function parseNegative(formulaNotation: string) {
+  const isNegative = formulaNotation[0] === 'n';
+  formulaNotation = isNegative ? formulaNotation.substring(1) : formulaNotation;
+  return {
+    isNegative,
+    formulaNotation,
+  };
+}
```
HactarCE, line 58: This functionality is already implemented in Rust. Use `parse_formula()` from `quadratic-core`.

src/ui/menus/CodeEditor/FormulaLanguageModel.ts
HactarCE: Actually, the regex is not quite correct -- it should allow spaces around `:` (but really, this PR shouldn't need to modify the regex)

src/styles.css
HactarCE, line 1: Is there some way to read these colors from `theme.ts`, or have the theme fetch them from here?

HactarCE (dismissed):
Looks good for the most part! I left a few comments of things to change

jimniels:
I'm not sure if this is just a hot code reloading thing, but I'm noticing locally that if I change some code in `useEditorCellHighlights`, the code in monaco disappears, and then if i go try to edit another cell on the grid, the whole app crashes.

See video:


https://github.com/quadratichq/quadratic/assets/1316441/1702f597-b2c6-460e-814a-a14d0bead404

src/styles.css
juanbercoff: I don't think there is a define in the theme multiple decorations to different instances of the same token, that's why I'm adding the decorations on the hook.

src/helpers/formulaNotation.ts
juanbercoff: Do you have a sync version of `parse_formula`? All of the updates are sync and it's not showing the highlights. The alternative is to transform `cursor.updates`  and `Updates` to async but that could have some unintended consequences.

juanbercoff:
> I'm not sure if this is just a hot code reloading thing, but I'm noticing locally that if I change some code in `useEditorCellHighlights`, the code in monaco disappears, and then if i go try to edit another cell on the grid, the whole app crashes.
> 
> See video:
> 
>  CleanShot.2023-06-15.at.14.54.52.mp4

Could you try to comment the effect cleanup `return () => editor.dispose();` in both of the hooks?

src/styles.css
jimniels: I've got a commit reworking colors where I try to do this...will push it up a little later

src/helpers/formulaNotation.ts
juanbercoff: Nevermind, I think all the parsing can be done in the hook.

jimniels:
I made a few design tweaks to the branch, including:

1. Match the dashed cursor thickness to match the default cursor thickness 

CleanShot 2023-06-15 at 17 05 47@2x

2. Tweak color values to 1) be pulled from base MUI color palette, 2) meet AA accessibility guidelines for text (since not only is the cell highlighted but the textual reference to the cell in the editor), and 3) be pulled all from one place.

CleanShot 2023-06-15 at 17 07 37@2x

Cell border is solid highlight color, with the active cell referenced in the code editor getting a background highlight of the same color with 10% opacity

https://github.com/quadratichq/quadratic/assets/1316441/25076e86-ed07-4dea-9557-e640128ca8ef

src/styles.css
jimniels: It's not the cleanest, but I did this in [my design tweaks commit](https://github.com/quadratichq/quadratic/pull/559/commits/38bbfcee55a04a1c34dd1710b37e04bb2b0d289f#diff-cc96e0860bf406f3947d0be247e9bf5490613137e5e40c9cf1e19124d8c1a6a0R30-R43). It pulls the cell reference color values from `theme.js`, dynamically creates the `.cell-reference-{i}` classes, and then sticks them in the DOM as a `<style>` tag (it felt right to stick that functionality in the `useEditorCellHighlights.ts` so its colocated with the feature itself).

src/helpers/formulaNotation.ts
juanbercoff: Done

juanbercoff:
@HactarCE Do you have rust parser for python cells?

HactarCE:
> @HactarCE Do you have a rust parser for python cells?

Nope

jimniels:
> @HactarCE Do you have a rust parser for python cells?

@davidkircos you've done more of the python stuff, maybe you know?

juanbercoff:
I added the python parser and fixed an issue when changing the range the fill was triggered on the old cellmatch range.

davidkircos:
@juanbercoff this branch looks pretty good for formulas, and I think we could merge it quicky.

I am not confident in the python parser and implementation, I think it's a bit more complex and we should split it into a separate branch. Can you do that?

This is a great feature, I'd love to get it in merged!

juanbercoff:
@davidkircos Sure, I'll revert to Jim's commit and only leave the fill fix

jimniels (approved):
This looks good to me 👍  Such a great UX enhancement, TY for adding! Looking forward to the python version :)

FWIW: while playing with the feature, I noticed that you can be typing into the code editor, and the shift your focus to the grid and edit a cell and the cell outlines for fomulas disappear while you're editing a cell on the grid. That's a nice touch — TY for accounting for that!

Jun-22-2023 13-53-33

PR #566, opened 2023-06-20
feat: post user feedback to slack

jimniels:


quadratic-api/src/routes/feedback.ts
```diff
@@ -25,6 +27,30 @@ files_router.post('/', validateAccessToken, async (request: JWTRequest, response
...
+  if (process.env.SLACK_FEEDBACK_URL) {
+    const payload = {
+      text: [
+        `📣 ${process.env.NODE_ENV === 'production' ? '' : '[STAGING]'} New product feedback`,
+        `*From:* ${userEmail ? userEmail : `[no email]`} (${user.auth0_user_id})`,
+        '*Message*:',
+        feedback,
+      ].join('\n\n'),
+    };
+    // @ts-expect-error we're on node18 so global fetch is ok
```
jimniels, line 42: I spent like 10 mins trying to figure out how to fix this, and it appears to be some typescript configuration we need to update, as the API ts config says `@tsconfig/node16/tsconfig.json` (node 16) but the package.json says we're running on node 18 (which I believe is what's running in heroku). So I eventually just added this comment to expect the error (node 18 ships with `fetch` as a global).

quadratic-api/src/routes/feedback.ts
```diff
@@ -10,13 +10,15 @@ const prisma = new PrismaClient();
 
 const RequestBodySchema = z.object({
   feedback: z.string(),
+  userEmail: z.string().optional(),
 });
 type RequestBody = z.infer<typeof RequestBodySchema>;
 
 files_router.post('/', validateAccessToken, async (request: JWTRequest, response) => {
-  const { feedback }: RequestBody = RequestBodySchema.parse(request.body);
+  const { feedback, userEmail }: RequestBody = RequestBodySchema.parse(request.body);
```
jimniels, line 18: There's probably a way we can look up the user's email on the server side based on the auth0 id? But i wasn't sure how to do it, and since we already have it on the client, I figured I'd just pass it from client to server and then the server can pipe it into the Slack message (this makes the message in Slack much more helpful than having to manually look it up ourselves)

quadratic-api/src/routes/feedback.ts
jimniels: I will look into this

quadratic-api/src/routes/feedback.ts
```diff
@@ -25,6 +28,23 @@ files_router.post('/', validateAccessToken, async (request: JWTRequest, response
     },
   });
 
+  // Post to Slack
+  // SLACK_FEEDBACK_URL is the Quadratic product feedback slack app webhook URL
+  // https://api.slack.com/apps/A05E24P1WRW/incoming-webhooks
```
davidkircos, line 33: @jimniels is this our slack post URL? It feels like this shouldn't be checked in to the codebase?

davidkircos (changes requested):
One question about the URL in the comment.

quadratic-api/src/routes/feedback.ts
jimniels: @davidkircos no this is the url to slack's API docs where, if you're logged in, you can see/get the proper web hook URL. It's ok if this is checked in cause you'll just see an error page if you're not authenticated to quadratic's slack

davidkircos: remove from codebase

davidkircos (approved):

PR #569, opened 2023-06-21
Add `HLOOKUP`, `VLOOKUP`, and `XLOOKUP`

HactarCE:
- [x] Implementation
- [x] Tests
- [x] Examples & documentation


davidkircos (changes requested):
The Autocomplete for VLookup doesn't have spaces between the argument names.
CleanShot 2023-06-22 at 13 47 42@2x
The argument names are more complex than in Sheets. Could we use similar language for clarity?
CleanShot 2023-06-22 at 13 48 17@2x
I thought the final argument was for if the source is sorted not `approx_match`?

Looks like `VLOOKUP(L3, B3:E500, 4, FALSE)` works fine but `VLOOKUP(L3, B3:E500, 4, FALSE())` gives the error "ERROR: Expected right paren or argument separator (comma)".

HactarCE:
> The Autocomplete for VLookup doesn't have spaces between the argument names.
I just merged from main, which fixed this

> The argument names are more complex than in Sheets. Could we use similar language for clarity?
Sure! I've updated that.
image

> I thought the final argument was for if the source is sorted not `approx_match`?
I've been going based on [Excel's version](https://support.microsoft.com/en-us/office/vlookup-function-0bbc8083-26fe-4963-8ab8-93a18ad188a1) -- interesting that Sheets is subtly different! I can change it to `is_sorted` if we want that.

> Looks like `VLOOKUP(L3, B3:E500, 4, FALSE)` works fine but `VLOOKUP(L3, B3:E500, 4, FALSE())` gives the error "ERROR: Expected right paren or argument separator (comma)".
I've just merged from #572, which fixed this

davidkircos (approved):

PR #572, opened 2023-06-22
Add boolean parsing in formulas

HactarCE:
This PR makes `TRUE` and `FALSE` work directly in formulas. So `IF(TRUE(), 1, 2)` is equivalent to `IF(TRUE, 1, 2)`. This is consistent with Excel.

### DIFF
--- a/quadratic-core/Cargo.lock
+++ b/quadratic-core/Cargo.lock
@@ -384,7 +384,7 @@ dependencies = [
 
 [[package]]
 name = "quadratic-core"
-version = "0.1.11"
+version = "0.1.12"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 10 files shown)


davidkircos (changes requested):
@HactarCE functionality looks good! Should we update the docs to show TRUE?

It's also a little weird that TRUE autocompletes to TRUE(). Maybe we can turn that off? 

Also the tests fail.

CleanShot 2023-06-22 at 13 39 03@2x

HactarCE:
Good idea with the autocomplete snippet! I've implemented that, and fixed the tests (I made a mistake with parsing priority that I thought I had committed but hadn't)

davidkircos (approved):

PR #580, opened 2023-06-27
fix: allow any for array_output in older versions of the schema

jimniels:
v1.2 version of the file schema allowed `any` for `array_output` to fix a bug (#561)

but if older versions of files have those same errors but are otherwise valid, they'll fail to load.

this makes it so (retroactively) we make previous file versions support `any` for `array_output` (as the code would, sometimes, support writing files out in an "invalid" way with more than just string|number|boolean in `array_output`)

**To test:**

- Try to import a v1.1 version of a file where the `array_output` is a value that's supported in v1.2 ([test-file.grid.txt](https://github.com/quadratichq/quadratic/files/11885332/test-file.grid.txt))
- Ensure that the file imports correctly

### DIFF
--- a/src/schemas/validateGridFile.test.ts
+++ b/src/schemas/validateGridFile.test.ts
@@ -35,7 +35,27 @@ const v1File: GridFileV1 = {
 
 const v1_1File: GridFileV1_1 = {
   version: '1.1',
-  cells: [],
+  cells: [
+    // Test to allow `any` for `array_output`
+    {
+      x: 0,
+      y: 0,
+      type: 'PYTHON',
+      value: '1,2,3',
+      last_modified: '2023-06-27T16:54:40.619Z',
+      evaluation_result: {
+        success: true,
+        std_out: '',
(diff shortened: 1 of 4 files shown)


src/schemas/validateGridFile.test.ts
```diff
@@ -35,7 +35,27 @@ const v1File: GridFileV1 = {
...
+        output_value: '[[[1, 2, 3]]]',
+        cells_accessed: [],
+        array_output: [[[1, 2, 3]]],
+        formatted_code: '[[[1, 2, 3]]]\n',
+        error_span: null,
+      },
+      python_code: '[[[1,2,3]]]',
+      array_cells: [[0, 0]],
+    },
+  ],
```
jimniels, line 58: Added a test for this to ensure older versions of files will support `any` for `array_output`

---

(Before this fix, this addition to the test would've made the tests fail)

CleanShot 2023-06-27 at 10 56 20@2x

davidkircos (approved):

PR #587, opened 2023-07-01
python web-worker port

davidfig:
Move Python loading and execution to a web worker

Todo:

- [x] Bug with "Python is Loading" in CodeEditor doesn't go away after loading.
- [x] Python loading indication in the BottomBar. Spinner or similar while loading and checkmark once loaded.
- [x] Fix tests

To Review:

- [ ] If Python fails to load, it's handled gracefully
- [ ] while loading makes sense


src/grid/computations/runCellComputation.ts
```diff
@@ -1,7 +1,8 @@
 import { Cell } from '../../schemas';
+import { webWorkers } from '../../web-workers/webWorkers';
 import { runAI } from './ai/runAI';
 import { runFormula } from './formulas/runFormula';
-import { runPython } from './python/runPython';
+// import { runPython } from './python/runPython';
```
davidkircos, line 5: remove

jimniels:
FWIW: I made a couple small visual tweaks, including:

1. Changed the "checks" in the footer to be icons from our icon pack rather than the unicode checkmark

Before: CleanShot 2023-07-12 at 10 33 35@2x

After: CleanShot 2023-07-12 at 10 35 28@2x

2. Changed the "Python is loading" to the slightly more terse "Loading Python" and updated the color to use a theme orange

Before: CleanShot 2023-07-12 at 10 37 19@2x

After: CleanShot 2023-07-12 at 10 36 42@2x

src/atoms/loadedStateAtom.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export interface LoadedState {
+  pythonLoaded: boolean | 'error';
+}
+
+export const loadedStateDefault: LoadedState = {
+  pythonLoaded: false,
+};
+
+export const loadedStateAtom = atom({
```
davidkircos, line 11: `loadingStateAtom`?

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -357,14 +361,20 @@ export const CodeEditor = (props: CodeEditorProps) => {
         </div>
         <div style={{ display: 'flex', alignItems: 'center', gap: '.5rem' }}>
           {isRunningComputation && <CircularProgress size="1.125rem" sx={{ m: '0 .5rem' }} />}
+          {isLoadingPython && (
+            <div style={{ color: theme.palette.warning.main, display: 'flex', alignItems: 'center' }}>
+              Loading Python...
+              <CircularProgress color="inherit" size="1.125rem" sx={{ m: '0 .5rem' }} />
+            </div>
+          )}
```
davidkircos, line 369: @jimniels to review UI/UX here

davidkircos (approved):
The code looks good! Awesome loading time improvement

jimniels (approved):
Functionality-wise, this looks good to me. Huge improvement — and love how you handled the async loading of python in the background.

I'm happy to give this a 👍  but want to note one thing:

If Python is taking a long time to load, it's possible you can get into the editor and make some edits. If python still hasn't loaded and you go to close the editor, you'll get the popup that says "do you want to save your changes?" If you click "Save & Run" it won't actually save and run because python has loaded yet.

CleanShot 2023-07-17 at 13 55 17@2x

From my testing, it seems like it closes the editor and just doesn't save whatever changes you had in there (but you don't know this until you re-open the editor for that cell).

I'm not sure what the best way is to solve this (and it seems like a pretty big edge case?). Seems like the options would be:

1. Don't trigger that "What do you want to do with your unsaved changes?" modal when python is still loading (and people will just lose their changes), or
2. Just hide the controls to save/close the editor entirely when python is loading, e.g.

 
CleanShot 2023-07-17 at 13 59 01@2x

Or, 3) figure out a way to save changes to code that is separate from running code (but that seems like a big change i would guess...?)

Anyhow, it's a pretty big edge case I think, but the functionality there is technically incorrect in the current code, where it asks if you want to save your changes and if you say "yes" it won't because python hasn't loaded.

PR #588, opened 2023-07-02
fix copying of large data

### LINKED ISSUE #584: copying large amounts of data doesn't work
To reproduce:

- Enter `1..1000 * {1, 2, 3, 4, 5, 6}` in a formula to generate a large block.
- Copy the block (does not work -- clipboard keeps old contents)
- Paste the block

davidfig:
Fix copying of large data. Closes #584

### DIFF
--- a/src/debugFlags.ts
+++ b/src/debugFlags.ts
@@ -2,7 +2,7 @@
 export const debug = process.env.REACT_APP_DEBUG === '1' ? true : false;
 
 // skip python load (used to speed up debugging)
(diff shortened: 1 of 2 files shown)


jimniels:
This looks ok to me, but @HactarCE since you filed the original issue, can you test against the case that made you file the issue?

HactarCE:
Yep! Seems to be resolved.

HactarCE (approved):

PR #591, opened 2023-07-03
Perf cell labels

davidfig:
Experiments with perf for CellsLabels


davidfig:
This is still an ongoing experiment. Will move it to next branch.

PR #592, opened 2023-07-03
feat: display cell information in bottom bar

jimniels:
Display formatting and cell type information about the currently active cell in the bottom bar.

Jul-03-2023 11-56-37

Notes:

- Format is: `Format: {formatType} ({cellType})` where:
  - `formatType` is one of: `text`, `number`, `currency`, `percentage`, `exponential`
  - `cellType` is only displayed if it's no plain text and is one of: `formula`, 'python', `sql`, `javascript`, `ai` or `computed'
- The "Format: {x}" disappears when there is a multi-selection

I'm open to leaving off the `cellType` and only displaying the formatting type until we get real cell data types, but i also think it's useful to see something like "text (python)" or "text (computed)" to understand "this cell is formatted as text but comes from python or a computed output"

### DIFF
--- a/src/ui/menus/BottomBar/BottomBar.tsx
+++ b/src/ui/menus/BottomBar/BottomBar.tsx
@@ -3,7 +3,6 @@ import { colors } from '../../../theme/colors';
 import { useRecoilState } from 'recoil';
 import { gridInteractionStateAtom } from '../../../atoms/gridInteractionStateAtom';
 import { useEffect, useState } from 'react';
-import { Cell } from '../../../schemas';
 import { formatDistance } from 'date-fns';
 import { focusGrid } from '../../../helpers/focusGrid';
 import { isMobileOnly } from 'react-device-detect';
@@ -11,6 +10,7 @@ import { debugShowCacheFlag, debugShowFPS, debugShowRenderer, debugShowCacheCoun
 import { Sheet } from '../../../grid/sheet/Sheet';
 import { editorInteractionStateAtom } from '../../../atoms/editorInteractionStateAtom';
 import { ChatBubbleOutline } from '@mui/icons-material';
+import { CellAndFormat } from '../../../grid/sheet/GridSparse';
 
 interface Props {
   sheet: Sheet;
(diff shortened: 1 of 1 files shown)


HactarCE:
It says `text` for me always, even for numeric values. Is this intentional?

jimniels:
Ah ok sorry for the confusion.

Yes, so this indicator is meant to communicate cell _format_ not data type.

That means that, by default, yes every cell with a value will display `Format: text`. The only way to change that is to explicitly use the formatting controls, e.g.

CleanShot 2023-07-05 at 10 20 33@2x

So if you had two seemingly identical columns with "numbers" in them, only the ones that have been formatted by the user as "number" will show `Format: number` whereas the other will show `Format: text`

Jul-05-2023 10-22-26

Maybe this is not totally clear to the user and we should wait until we have data types....although I'm not sure that solves the issue either, as you might still want to know format, e.g. if you see "$12" will you want to see: "Type: number · Format: currency".

I do think that hinting at a cell's formatting is useful today, though maybe we should only do it when a formatting is explicitly defined by the user, e.g. if somebody types "1" you don't see any format in the bottom bar, but once they set the cell "1" to be formatted as a "number", you would see "Format: number" in the bottom bar. In that case, perhaps by not explicitly stating "Format: text" it is implied and doesn't cause confusion when you type "1" and see "Format: text".

waveminded:
If this supposed to be async? I tried updating the cell and it doesn't update until I unfocus the cell then return to it.

jimniels:
I'm tempted to say we should just show this for cells with explicit formatting:

Jul-05-2023 12-56-57

But in a sense, that's kind of already clear just by looking at the cell (does the value look like "$10.05"? it's probably currency...)

Probably the most useful implementation of this would be some combo of data type + formatting, e.g. 

| Cell value | Bottom bar |
|-----------|------------|
| `ten`            | `Data: string`|
| `10`| `Data: number`|
|`$10` | `Data: number, formatted as currency`|

Given that we won't have data types until later, I think i'm going to close this PR until we can do this kind of implementation. I think that's the right way to do this and doing it any earlier might just confuse people.

PR #593, opened 2023-07-03
feat: show/hide cursor when browser window goes blur/focus

jimniels:
In a native app like Excel, when your window is focused you see the cursor, but when you focus another window (at the OS-level), the cursor disappears.

Jul-03-2023 12-49-04

This mimics OS-level functionality for the UI/UX around windowing. It's useful, for example, on large monitors where you have two windows open side-by-side and you're copy/pasting information between the two documents. It helps you keep track of which window you're focused in and dealing with.

https://github.com/quadratichq/quadratic/assets/1316441/f6643f02-8843-4022-95eb-d7c0e55ff143

Thinking that it would be useful to do the same in Quadratic. In this prototype, I copied what excel does. The cursor should disappear when the browser window loses focus (single or multi cell selection)

Jul-05-2023 15-52-35

### DIFF
--- a/src/gridGL/UI/Cursor.ts
+++ b/src/gridGL/UI/Cursor.ts
@@ -32,6 +32,7 @@ export class Cursor extends Graphics {
   }
 
   private drawCursor(): void {
+    if (!document.hasFocus()) return;
     const { settings, viewport } = this.app;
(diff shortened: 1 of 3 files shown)


davidkircos:
I don't know of any web apps that do this, I wonder if there is a good reason.

PR #594, opened 2023-07-04
feat: show active cell selection formula results in bottom bar

jimniels:
## Background

Sheets and Excel both have a feature that runs six formulas (like sum and average) for an active multi-cell selection. In this way, you don't have to write a formula to see its output. You can just make a selection on the sheet and see the output change as you change your selection.

### Formulas

1. `AVERAGE`
2. `COUNTA`
3. Numerical count
4. `MIN`
5. `MAX`
6. `SUM`

### Screenshot examples

Excel web:

CleanShot 2023-07-03 at 22 56 10@2x

Excel native:

CleanShot 2023-07-03 at 22 51 32@2x

Sheets:

CleanShot 2023-07-03 at 22 57 06@2x
CleanShot 2023-07-03 at 22 57 14@2x

## Quadratic

Excel and Sheets implement this feature slightly differently, but the main idea is that the UI becomes configureable and you can choose which formula you want to see in the bottom bar (sheets only ever allows once choice, while excel is configureable and lets you choose which of the 6 you want to see).

To start in Quadratic, we'll provide Sum, Average, and Count. We can add more later as we deem necessary.

Working example:

Jul-03-2023 23-04-34

### To Test

Make various selections across the grid and ensure the following:

- Only show `count` if at least two or more cells with data are selected.
- Only show `sum` and `average` if at least one or more cells of the active selection are numbers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22218,7 +22218,7 @@
     },
     "quadratic-core/pkg": {
       "name": "quadratic-core",
-      "version": "0.1.7",
+      "version": "0.1.13",
       "license": "MIT"
     }
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -17311,8 +17311,9 @@
       }
     },
     "node_modules/quadratic-core": {
-      "resolved": "quadratic-core/pkg",
-      "link": true
+      "version": "0.1.7",
+      "resolved": "file:quadratic-core/pkg",
+      "license": "MIT"
     },
     "node_modules/querystringify": {
       "version": "2.2.0",
@@ -22215,11 +22216,6 @@
       "funding": {
         "url": "https://github.com/sponsors/colinhacks"
       }
-    },
-    "quadratic-core/pkg": {
-      "name": "quadratic-core",
-      "version": "0.1.13",
-      "license": "MIT"
     }
   },
   "dependencies": {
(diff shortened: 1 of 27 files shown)


jimniels:
Todos:

- Hide the info when it gets small (responsiveness)
- Put to the right hand side (instead of left)

jimniels:
I've revised the remaining items on this PR. The bugs that exist at this point seem to stem from the bug in `COUNT` so I am re-assigning this to @HactarCE

HactarCE:
@jimniels Is it possible to limit the float precision when printing?
image

jimniels:
@HactarCE I'm sure there is. Excel seems to approximately match the level of precision in the inputs with the level of precision in the output. 

So if you do the average of two whole numbers, you might get one decimal in the output. But if you do the average of 0.0000007 and 0.0000008, you'll probably get a number with six zeros or so. 

David and I talked about this briefly, but ultimately just decided to leave it unoptimized for now.

davidkircos (changes requested):
There are performance concerns with this branch with large selections.

- [x] When moving the cursor over a large selection, this lags significantly
- [x] When selecting all in a large doc, this freezes the whole doc while running

Suggested solution:
Only calculate for a max selection size of 100 cells

davidkircos (approved):

davidkircos (approved):

PR #596, opened 2023-07-05
refactor: tweak color of heading grid lines

jimniels:
I never loved that the accent color (blue) is being used for the grid heading lines.

This changes it from being the accent color (used to indicate "active"-ness in the grid, e.g. cursor color, active cell heading background color, etc.) to our neutral "gray" which is used for the other heading lines.

CleanShot 2023-07-05 at 15 59 50@2x


HactarCE:
Good catch!

davidkircos (approved):

PR #597, opened 2023-07-07
autoformat currency

### LINKED ISSUE #590: fix: automatically set "$10" to currency
[Per David's comment](https://github.com/quadratichq/quadratic/issues/590#issuecomment-1618909880)

> Yeah, I think the desired behavior is when entering "$10" to strip the value to 10 and set the formatting to right aligned currency

Like excel does it:

Jul-03-2023 11-21-28

---

**Old description** (see thread below)

When you have a column formatted as currency and change the decimal formatting, it changes any amounts already with cents, but not for whole dollar amounts

Jul-03-2023 11-07-29

davidfig:
Closes #590 

When users inputs "$n" where n is a valid float number:
1.  strips the $ in the value
2. adds currency formatting
3. adds right-aligned formatting

Note, there is no way to add $10 as a string currently. It will always convert. A few possibilities:
1. make the formatting a separate transaction so you can undo the formatting (this is how Word works)
2. do not apply formatting when editing a cell that already has a value
3. don't worry about it

### DIFF
--- a/src/gridGL/interaction/CellInput.tsx
+++ b/src/gridGL/interaction/CellInput.tsx
@@ -10,7 +10,7 @@ import { EditorInteractionState } from '../../atoms/editorInteractionStateAtom';
 import { CellFormat } from '../../schemas';
 import { useFormatCells } from '../../ui/menus/TopBar/SubMenus/useFormatCells';
 import { CURSOR_THICKNESS } from '../UI/Cursor';
-import { isStringANumber } from '../../helpers/isStringANumber';
+import { isStringANumberWithCommas } from '../../helpers/isStringANumber';
 
 interface CellInputProps {
   interactionState: GridInteractionState;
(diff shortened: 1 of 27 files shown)


davidkircos:
I am not sure if this is robust enough of a solution.

What if a formula or code returns "$10"? How should we handle that?

src/gridGL/interaction/CellInput.tsx
```diff
@@ -160,6 +164,12 @@ export const CellInput = (props: CellInputProps) => {
             create_transaction: false,
           });
       } else {
+        // convert to dollars
+        if (value[0] === '$' && !isNaN(parseFloat(value.substring(1).trim()))) {
+          value = value.substring(1);
+          textFormatSetCurrency();
+          changeAlignment('right');
+        }
```
davidkircos, line 172: It should be possible to do this all in one transaction.

I think we'd only want to update the Alignment if it wasn't previously set by the user.

davidfig: This all happens in one transaction (all set transactions are false). I'll add the alignment check and not change it unless it's originally undefined.

davidfig:
Added automatic format on paste as well

davidkircos:
I think we should wait on this until we have a type system on the grid.

PR #602, opened 2023-07-10
Rebuild Rust on `npm start`

HactarCE:
This works, but I don't know if it's idiomatic. feels kinda hacky.

I know `cargo` is smart and only rebuilds what it needs to, so don't worry about unnecessary rebuilds

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
   },
   "main": "public/electron.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


davidkircos:
What if the project is running on a computer without the Rust build chain setup?

davidkircos:
@andrew This is resolved in the `next` branch?

PR #606, opened 2023-07-12
Strip formatting on paste while in CellInput

### LINKED ISSUE #601: bug: Pasting Formatted Content into Cell
When pasting rich text into the ContentEditable input, it visually shows the HTML format. However, when saving, the format disappears. This issue does not occur when pasting into the sheet, only into the ContentEditable input.

Jul-08-2023 09-11-45

Here is my proposed solution:
- [x] Strip formatting on paste into content editable
- [x] Strip new lines when pasting into content editable, and strip new lines when saving a cell value. We currently don't support multi-line cells.

davidfig:
Closes #601. 

To replicate: double click a cell and then paste either a html formatted input or multiline input. it should strip both


src/gridGL/interaction/CellInput.tsx
```diff
@@ -221,6 +221,15 @@ export const CellInput = (props: CellInputProps) => {
   // set input's initial position correctly
   const transform = updateInputCSSTransform();
 
+  const handlePaste = (event: ClipboardEvent) => {
+    const text = event.clipboardData.getData('text');
+    const parsed = new DOMParser().parseFromString(text, 'text/html');
+    const result = parsed.body.textContent || '';
+    // event.clipboardData.setData('text', result.replaceAll('\n', ''));
```
davidkircos, line 228: @davidfig should this line be removed?

src/gridGL/interaction/CellInput.tsx
```diff
@@ -221,6 +221,15 @@ export const CellInput = (props: CellInputProps) => {
   // set input's initial position correctly
   const transform = updateInputCSSTransform();
 
+  const handlePaste = (event: ClipboardEvent) => {
+    const text = event.clipboardData.getData('text');
```
davidkircos, line 225: does this work on all browsers?

davidkircos: Could It be cleaner to watch for value to change and strip all non supported chars?

Quadratic only supports a small set of chars at the moment, alphanumeric, and symbols. Maybe we should clean the input of everything that isn't supported on the grid?

davidkircos (changes requested):
Have a few comments and questions on the PR

Functionality seems to work great!

src/gridGL/interaction/CellInput.tsx
davidfig: Just tried it on Firefox and Edge. Pretty standard API.

davidkircos:
added to `next`

PR #628, opened 2023-07-28
Python return types

### LINKED ISSUE #627: feat: show return type for python code
From #376 
The goal of this branch is to make Python work better in Quadratic.

- [x]  Clearly display to a user what from their code is being returned to the grid.
- [x]  Add back code prettier.
- [ ]  Small output icon next to code once run (disappears when editing)
- [ ]  OUTPUT tab showing output details and inspection

juanbercoff:
From #376 
Resolves #627 

- Adapt inspect_python to work with service worker.
- Fix bug in windows having to save the editor 2 times: the onChange event of the editor handles the line ending different in windows and adds '\r\n'. This makes unsavedChanges true and prompts an additional save.
- Removed logs from console when no editor content after saving.


jimniels:
This was superseded by the work merged in #1160. Closing.

PR #632, opened 2023-08-03
Preview Branches

davidkircos:
Properly configured Heroku for API previews and Vercel for front end preview branches.


quadratic-api/src/routes/files/files.test.ts
davidkircos, line 1: All the changes here do is fix the tests

PR #634, opened 2023-08-05
Electron autoupdate

### LINKED ISSUE #629: feat: electron build and auto update

juanbercoff:
Resolves #629 

I ended up using forge as recommended https://www.electronjs.org/blog/forge-v6-release

When you open the build it throws an error because it's not code signed. I added the auto updater with the default config that check for updates on startup and every 10 min and automatically downloads them on the background. This could be configured but I think it's best to code sign, test that's working and then configure the auto updater as necessary.

I try building it on windows but there's an issue with the symlink of quadratic-core. I think that's happening because I have the project in a different hard drive.

Forge automatically detects your OS and uses the corresponding maker.

Use `electron:make` to build the electron app.

This PR also fixes a bug in electron where you couldn't close the app.

I had to add the homepage key to package.json and modify various imports for the build to work, please check that's deploying as expected.

- [ ] Update server and feed url 
- [ ] Update forge config with S3 bucket

### DIFF
--- a/.github/workflows/electron-release.yml
+++ b/.github/workflows/electron-release.yml
@@ -1,0 +1,55 @@
+name: Build electron on release
+
+on:
+  workflow_dispatch:
+  #release:
+  #  types: [published]
+
+jobs:
+  build-mac:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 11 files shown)


davidkircos:
deprioritized

PR #635, opened 2023-08-07
David UI Refinements

davidkircos:


src/shared/dashboard/FileListItem.tsx
```diff
@@ -17,7 +16,7 @@ export default function FileListItem({
   const theme = useTheme();
 
   return (
-    <Box sx={{ '&:hover': { background: colors.canvasLayer2, cursor: 'pointer' } }}>
+    <Box sx={{ '&:hover': { background: theme.palette.grey[100], cursor: 'pointer' } }}>
```
jimniels, line 19: Note to self: we need these to have a bottom margin (or clear border) so when you hover over elements that are next to each other they don't blend in to one another)

CleanShot 2023-08-09 at 19 29 58@2x

src/shared/dashboard/Layout.tsx
```diff
@@ -203,20 +203,25 @@ function SidebarNavLink({ to, children, style, isLogo }: any) {
     to === navigation.location?.pathname;
 
   return (
-    <NavLink to={to} style={{ ...style, position: 'relative' }}>
-      {isActive && (
-        <div
-          style={{
+    <NavLink to={to} style={{ ...style, position: 'relative' }} className={'navLinkQuadratic'}>
```
jimniels, line 206: what is this classname for? i don't see it referenced anywhere else in this PR

src/shared/dashboard/Layout.tsx
```diff
@@ -203,20 +203,25 @@ function SidebarNavLink({ to, children, style, isLogo }: any) {
...
+      <Box
+        sx={[
+          {
             position: 'absolute',
             top: '0',
             left: '0',
             right: '0',
             bottom: '0',
-            backgroundColor: theme.palette.primary.light,
+            backgroundColor: isActive ? theme.palette.grey[400] : 'inherit',
```
jimniels, line 215: I think this is a good effect for the navigation items, but i don't love it for the logo...

I'm still debating if the logo should be clickable. Does it seem weird that you click it and it goes to "My files"? 


If we keep it, we should fix the padding so it's equal, right now you get uneven spacing around it when you hover (note how there's more padding on the left than the right)

CleanShot 2023-08-09 at 19 37 18@2x

src/shared/dashboard/Header.tsx
```diff
@@ -24,7 +24,7 @@ export default function Header({ title, actions }: { title: string; actions?: Re
         },
       }}
     >
-      <Typography variant="h5" sx={{ py: theme.spacing(2) }} color="text.primary">
+      <Typography variant="h6" sx={{ py: theme.spacing(2), fontWeight: 400 }} color="text.primary">
```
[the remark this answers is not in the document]
davidkircos: I am just making the font smaller. I think it's a little bit large. 

I like how apps like Linear have small fonts. Smaller fonts give the UI a sense of higher precision IMO.

src/shared/dashboard/FileListItem.tsx
davidkircos: I looked at a few apps, and what they do is raise the Icon slightly for the hovered item - while keeping the highlight color exactly the same.

src/shared/dashboard/Layout.tsx
davidkircos: Oh I was trying to add the hover over with CSS. You an remove this.

src/shared/dashboard/Layout.tsx
davidkircos: I like it for the logo. It just indicates that it will navigate you somewhere. Linear also highlights the logo on mouseover 🤷 

Ah yeah, we should fix the padding.

src/shared/dashboard/FileListItem.tsx
jimniels: Yeah what I mean is add a separator between the navigational elements, so they the active one doesn't blend in to the one you're hovering, e.g. here's linear:

CleanShot 2023-08-10 at 12 33 07@2x

I'll add this...

jimniels: Updated these to include a subtle hover effect:

Aug-10-2023 13-05-45

This is the same effect you see on the file list item as well:

Aug-10-2023 13-06-10

And it now uses the color mode agnostic value from the theme, so these work in dark mode too

Aug-10-2023 13-07-20

src/shared/dashboard/Header.tsx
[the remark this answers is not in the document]
jimniels: I feel like if we're going to go tighter, we should do it more universally. Let me play with this a bit...

jimniels (approved):
Made a few more changes and tweaks. Figured we'd go ahead and merge this and then if we want to make further adjustments, we'll do it in the `share-file` branch.

PR #636, opened 2023-08-10
feat: social share meta/image

jimniels:


public/index.html
```diff
@@ -22,6 +22,19 @@
...
     -->
     <title>Quadratic</title>
+    <!-- 
+      Social share meta info
+    -->
+    <meta property="og:type" content="website" />
+    <meta property="og:title" content="Quadratic" />
+    <meta property="og:description" content="Modern spreadsheet software." />
+    <meta property="og:url" content="https://%VERCEL_URL%" />
+    <meta property="og:image" content="https://%VERCEL_URL%%PUBLIC_URL%/images/social.png" />
```
jimniels, line 32: @davidkircos what are your thoughts on this?

These image URLs need to be fully qualified URLs, which means to truly test them we need to have the environment URL in there.

The way (I believe) this works is you can do `%ENV_VAR%` and `react-sripts` which does our front-end build will either replace that with the corresponding environment variable, or do nothing.

For our vercel builds, we should be able to do `VERCEL_URL` but then when amplify runs, idk if there's a corresponding environment variable for that one? If there was, we could do both:

`https://%VERCEL_URL%%AMPLIFY_URL%` cause both of those values should never be present, but that is kinda ugly and a bit confusing...

The simpler option is we could just stick this image in our [quadratic-media repo](https://github.com/quadratichq/quadratic-media) and then just have one consistent URL across all our environments for the social images. That seemed like the easiest option, though it's out of this code base (and I don't have write access to that repo, so i'd need your help there)

PR #640, opened 2023-08-15
Ajf/controller typescript

davidfig:


quadratic-core/src/controller.rs
```diff
@@ -37,6 +37,11 @@ impl GridController {
         self.grid.sheet_from_id(sheet_id)
     }
 
+    pub fn set_sheet_name(&mut self, sheet_id: SheetId, name: String) {
```
HactarCE, line 40: We'll need to make a transaction for this too.

quadratic-core/src/grid/ids.rs
```diff
@@ -18,6 +18,16 @@ macro_rules! uuid_wrapper_struct {
...
+
+            pub fn from_string(s: &str) -> Self {
+                $name {
+                    id: Uuid::parse_str(s).unwrap_or_default(),
+                }
+            }
+
+            pub fn id_to_string(&self) -> String {
+                self.id.to_string()
+            }
```
HactarCE, line 30: `impl FromStr` (which will automatically give `.parse()`) and `Display` (which will automatically give `.to_string()`) instead of these methods.

Also, I don't know what a default UUID is but I think I'd rather panic than have one of those.

quadratic-core/src/grid/mod.rs
```diff
@@ -152,12 +152,21 @@ impl Grid {
     pub fn sheets(&self) -> &[Sheet] {
         &self.sheets
     }
+    pub fn sheets_to_sorted_sheet_ids(&self) -> Vec<SheetId> {
```
HactarCE, line 155: Good call factoring this out! `sheet_ids()` would be more concise though

quadratic-core/src/grid/mod.rs
```diff
@@ -152,12 +152,21 @@ impl Grid {
...
+        let sheets = &self.sheets.to_vec();
+        sheets.iter().map(|sheet| sheet.id).collect()
```
HactarCE, line 157: No need to allocate again with `.to_vec()`. `self.sheets.iter().map(...).collect()` works fine.

quadratic-core/src/grid/mod.rs
```diff
@@ -152,12 +152,21 @@ impl Grid {
...
         &mut self.sheets
     }
     /// Adds a sheet to the grid. Returns an error if the sheet name is already
     /// in use.
-    pub fn add_sheet(&mut self, sheet: Sheet, index: Option<usize>) -> Result<(), ()> {
+    pub fn add_sheet(&mut self, sheet: Option<Sheet>, index: Option<usize>) -> Result<SheetId, ()> {
+        let sheet = sheet.unwrap_or({
+            let sheet_id = SheetId::new();
+            Sheet::new(sheet_id, format!("Sheet {}", self.sheets.len().to_string()))
+        });
```
HactarCE, line 168: This is constructing a new `Sheet` even when the caller provides one. Use `.unwrap_or_else(|| { ... })` instead.

Also, this will fail if the user already has a sheet with the same autogenerated name. We should figure out where in the code to pick a new name that is guaranteed to be unique. Here _would_ be good, except for multiplayer shenanigans where we need to know the real sheet name at the transaction level.

quadratic-core/src/grid/mod.rs
```diff
@@ -152,12 +152,21 @@ impl Grid {
...
+        let id = sheet.id.clone();
```
HactarCE, line 169: `SheetId` is `Copy`, so `.clone()` is not necessary.

quadratic-core/src/grid/sheet.rs
```diff
@@ -631,4 +631,20 @@ impl Sheet {
...
         todo!("unspill cells from {source:?}")
     }
+
+    pub fn get_name(&self) -> String {
+        self.name.clone()
+    }
+
+    pub fn get_color(&self) -> Option<String> {
+        self.color.clone()
+    }
```
HactarCE, line 641: [Rust convention](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) omits the `get_`. Also, no need to `.clone()` if the caller just needs a reference. So:
```rs
pub fn name(&self) -> &str { ... }
pub fn color(&self) -> &Option<String> { ... }
```

HactarCE: Also, these fields are already public, so there's no need for getters and setters.

quadratic-core/src/wasm_bindings/controller.rs
```diff
@@ -226,13 +267,31 @@ impl GridController {
...
+    }
+
+    #[wasm_bindgen(js_name = "getSheetColor")]
+    pub fn get_sheet_color(&self, sheet_id: String) -> String {
+        let sheet_id = SheetId::from_string(&sheet_id);
+        let sheet = self.grid().sheet_from_id(sheet_id);
+        match &sheet.color {
+            Some(color) => color.clone(),
+            None => String::from(""),
+        }
```
HactarCE, line 292: This works. `sheet.color.clone().unwrap_or_default()` is equivalent; use whichever you prefer

HactarCE (changes requested):

quadratic-core/src/controller.rs
davidfig: already done. that was mostly for test

quadratic-core/src/grid/mod.rs
davidfig: I'm noticing a lot of concise naming in rust....

PR #641, opened 2023-08-15
refactor api client

jimniels:
When the API client was small we just copy/pasted boilerplate code between the different endpoint calls, but now we have more API calls and we run the risk of making our reporting not uniform. And our code styles in between function calls is different as well.

This attempts to handle that all uniformly by making all the API calls use the same `safeFetch` function which supplies base options to all API as well as a zod schema, and then that function (which is used on all API calls) handles errors uniformly, reports them to sentry uniformly, and validates every request against a specified zod schema.


src/api-client/apiClientSingleton.ts
```diff
@@ -35,151 +50,54 @@ class APIClientSingleton {
...
   }
 
   // Fetch a file from the DB
   async getFile(id: string): Promise<GetFileRes | undefined> {
-    try {
-      const base_url = this.getAPIURL();
-      const response = await fetch(`${base_url}/v0/files/${id}`, {
-        method: 'GET',
-        headers: {
-          Authorization: `Bearer ${await authClient.getToken()}`,
```
jimniels, line 73: Minor: but this is an example where we do things differently between API calls. Here's we await the token right inline, but elsewhere we [create a var for it](https://github.com/quadratichq/quadratic/pull/641/files#diff-6b20f7b3bd9463e58b5e1b4c201292fe211e586d05528c182b4f8f654aa2d8beL41) then [stick it in the Authorization header using a template literal](https://github.com/quadratichq/quadratic/pull/641/files#diff-6b20f7b3bd9463e58b5e1b4c201292fe211e586d05528c182b4f8f654aa2d8beL45)

quadratic-api/src/routes/feedback.ts
```diff
@@ -46,7 +46,7 @@ files_router.post('/', validateAccessToken, userMiddleware, async (req: Request,
     });
   }
 
-  res.status(200).end();
+  res.status(200).json({ message: 'Feedback submitted!' });
```
jimniels, line 49: This is what we do on other endpoints (return a message that says it was done) so this is being uniform.

src/api-client/apiClientSingleton.ts
```diff
@@ -193,30 +111,53 @@ class APIClientSingleton {
...
+  }
+}
 
-      if (!response.ok) {
-        throw new Error(`API Response Error: ${response.status} ${response.statusText}`);
-      }
+export default APIClientSingleton.getInstance();
 
-      const json: PostFilesRes = await response.json();
+// Borrowed from: https://gist.github.com/gimenete/dd1886288ee3d3baaeae573ca226048f
```
davidkircos, line 122: The referenced code is not necessarily available under a free-use license. We should make sure the author gives us permission or we just referenced and not copied the code.

davidkircos (commented):
looks great! I like this.

Couple suggestions in Review

src/api-client/apiClientSingleton.ts
jimniels: It's sort of conceptually where we jumped off of. I changed/refactored parts of it to fit our codebase. I think we're in the clear. I can remove the comment TBH because it's really not that useful of a reference.

src/api-client/apiClientSingleton.ts
```diff
@@ -193,30 +111,53 @@ class APIClientSingleton {
...
+export async function safeFetch<T>(path: string, init: RequestInit, schema: z.Schema<T>): Promise<T> {
```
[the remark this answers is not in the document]
jimniels: I think i'm going to rename it to `fetchFromApi()` and i'll put it in another file.

src/api-client/apiClientSingleton.ts
davidkircos: Sounds good, lets do that.

src/api/apiClient.ts
```diff
@@ -0,0 +1,101 @@
+import mixpanel from 'mixpanel-browser';
```
davidkircos, line 1: This isn't a Client anymore. Maybe rename to api.ts?

src/api/apiClient.ts
```diff
@@ -0,0 +1,101 @@
...
+  cells: [],
+  formats: [],
+  columns: [],
+  rows: [],
+  borders: [],
+  cell_dependency: '',
+  version: GridFileSchema.shape.version.value,
+};
+
+async function getFiles() {
```
davidkircos, line 34: It would be nice if you could look at all these functions and know at a glace that they are going to make an API call. 

Maybe some sort of naming convention or exporting them as properties of an object called `API`.

davidkircos (approved):

PR #650, opened 2023-08-17
refactor: centralized env variables

jimniels:
Rather than having `process.env.REACT_APP_VAR_NAME` spread throughout the app in many places (and in each place, having logic like "check if var is defined, log to sentry if not, continue doing thing" etc.)  this centralizes all env variables into one place in the app and the you have to explicitly import the constant where you use it.

Also: checking for env variables is now part of the build step. So the build won't compile if any env variables are missing.


src/constants/env.ts
```diff
@@ -0,0 +1,49 @@
...
+export const REACT_APP_QUADRATIC_API_URL = process.env.REACT_APP_QUADRATIC_API_URL || '';
+export const REACT_APP_VERSION = process.env.REACT_APP_VERSION || '';
+// Note: anytime you add another required var here, add it to the object below
+const requiredVarsByName = {
+  REACT_APP_AUTH0_DOMAIN,
+  REACT_APP_AUTH0_CLIENT_ID,
+  REACT_APP_AUTH0_AUDIENCE,
+  REACT_APP_AUTH0_ISSUER,
+  REACT_APP_QUADRATIC_API_URL,
+  REACT_APP_VERSION,
```
davidkircos, line 28: Remove REACT_APP_* prefix from all of these when returned as constants?

src/api/apiClient.ts
```diff
@@ -92,17 +92,7 @@ export const apiClient = {
   },
 
   getApiUrl() {
```
davidkircos, line 94: Seems like we don't need this function anymore? Just reference QUADRATIC_API_URL directly

src/constants/env.ts
jimniels: I actually did this in the first version of this code, basically mapping all these vars to their equivalent without the REACT_APP_ prefix, e.g.

```
VERSION: REACT_APP_VERSION,
QUADRATIC_API_URL: REACT_APP_QUADRATIC_API_URL
// etc
```

But then I decided against it because while it saved some characters and typing, it introduces another layer of abstraction that i don't think is really worth it. Keeping them the same name I think makes it easier to reason about env variables in the app cause they're the same name everywhere, from the `.env` file to the constant to where they're used in the code. same everywhere.

davidkircos:
@jimniels what's the status on this pr. I think we left off wanting the build to fail if the required envs are not present?

jimniels:
@davidkircos I couldn't come up with a good way to make the build fail. Because we're using react-scripts for the build, we don't really have a good way to hook into it and say "if these env vars aren't here, fail"

I actually thought it'd be easy, just run a script before the build, e.g.

`npm run build` and in `package.json` have a `prebuild` script that just runs the `env.ts` file, but for the life of me i couldn't get typesript to fire up with `dotenv` to run that single file with our env vars. Maybe there's a better way? I meant to step away from it and come back with fresh eyes. Maybe you have a better idea...

jimniels:
1. Get this working for the server too
2. Base it off `next` instead of `main`
3. See about [nx.dev](http://nx.dev/)

davidkircos:
Ship after we ship #609

jimniels:
Closing this. We can use it later to reference how to do this if we get the time

PR #653, opened 2023-08-18
feat: read only share

jimniels:
Allow users to share files and other users (or non-authed viewers) can open the file and view it.

Things to test:

- Responsiveness (works across devices)
- Permissions
  - There are three kinds of users to test for: 1) Owner, 2) Sharee (auth'd user but not owner), 3) Anon (non auth'd user)
  - For each, make sure they only have access to applicable actions in the app
- `/file` route is accessible to non-authenticated users, every other route is still protected.
- File-related actions are accessible from all three places: 1) Quadratic menu, 2) file name dropdown, 3) command palette

Todos:

- [x] Rename GlobalSnackbar to GlobalSnackbarProvider
- [x] Need name (& avatar where possible) of file owner
- [x] Properly permission actions uniformly across quadratic menu, file name dropdown, and command palette
- [x] Fix TODOs in PR code
- [x] Uniformly style topbar menu buttons
- [x] Note on file list what files are shared
- [x] Refine top menus with hover bg color
- [x] Don't draw the cursor when you're on a mobile device (as it's read only)
- [x] Read only dialogs
- [x] Disable keyboard shortcuts when in read-only mode (e.g. "ENTER" to edit the active cell, or "CMD + X" to cut, etc)
- [x] New file icons for new actions (rename, duplicate, share)

Launching

- [x]  Add new ENV Vars to Production
- [x]  Set up Auth0 (steps below)
- [x] Create an Auth0 Machine to Machine Application
- [x] Check that Grant Types include Client Credentials
- [x] Check that APIs include Auth0 Management API
- [x] Check that APIs Auth0 Management API Scopes (via dropdown) include read:users


quadratic-api/src/auth0/profile.ts
```diff
@@ -0,0 +1,32 @@
...
+    where: { id: user_id },
+  });
+
+  const auth0_user = await auth0.getUser({ id: user.auth0_id });
+
+  return {
+    name: auth0_user.name,
+    picture: auth0_user.picture,
+    email: auth0_user.email,
+  };
```
jimniels, line 31: @davidkircos  What are the guarantees, if any, on this data from the API? I know for the front-end auth0 lib says this could be `User | undefined`. But, from what I can tell, the server guarantees it'll always return a user and each of these will always be a string.


CleanShot 2023-08-30 at 19 02 00@2x

quadratic-api/src/routes/files/files.ts
```diff
@@ -255,10 +212,11 @@ files_router.post(
         name: true,
         created_date: true,
         updated_date: true,
+        public_link_access: true,
```
jimniels, line 215: @jimniels I'm debating having us remove this entirely from every response except the `/sharing` endpoint....

quadratic-api/src/routes/files/sharing.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    // change file link permissions to NOT_SHARED
+    const res4 = await request(app)
+      .post('/v0/files/00000000-0000-4000-8000-000000000000/sharing')
+      .send({ public_link_access: 'NOT_SHARED' })
+      .set('Accept', 'application/json')
+      .set('Authorization', `Bearer ValidToken test_user_1`)
+      .expect('Content-Type', /json/)
+      .expect(200); // OK
+
+    expect(res4.body).toMatchObject({ message: 'File updated.' });
```
jimniels, line 113: @davidkircos should we have a test that tries to change the `public_link_access` to something that's not one of the enums? e.g. `public_link_access: 'foo'`

jimniels: Or even an empty body...

quadratic-api/src/routes/files/sharing.ts
```diff
@@ -0,0 +1,85 @@
...
+    let owner = await getUserProfile(req.file.ownerUserId);
+    if (!req.user) {
+      owner.email = null;
+    }
+
+    return res.status(200).json({
+      owner: owner,
+      public_link_access: req.file.public_link_access,
+      users: [],
+      teams: [],
```
jimniels, line 46: Minor: I feel like we should just remove `users` and `teams` from this response. I left them out from the schema validation on the front-end because i don't need them ATM. Maybe we should just do the same for the backend (comment them out here?)

quadratic-api/src/routes/files/sharing.test.ts
jimniels: Actually [it doesn't look like we actually check](https://github.com/quadratichq/quadratic/pull/653/files#diff-496e7b22f56ea5e458aa2ef7214600a1288edbeff67b2ac60270e125abd9fbc5R69-R79) to see if it's an expect value, which we probably should...

jimniels (commented):
@davidkircos left a couple comments for you

quadratic-api/package.json
```diff
@@ -15,7 +15,8 @@
     "docker:down": "docker-compose down",
     "docker:dev": "npm run docker:up && dotenv -e .env.docker -- prisma migrate deploy && dotenv -e .env.docker -- ts-node-dev --watch ./src --project ./tsconfig.json ./src/server.ts && npm run docker:down",
     "docker:test:dev": "npm run docker:up && dotenv -e .env.docker -- prisma migrate deploy && dotenv -e .env.docker -- jest --watch --detectOpenHandles && npm run docker:down",
-    "docker:test:ci": "npm run docker:up && dotenv -e .env.docker -- prisma migrate deploy && dotenv -e .env.docker -- jest && npm run docker:down"
+    "docker:test:ci": "npm run docker:up && dotenv -e .env.docker -- prisma migrate deploy && dotenv -e .env.docker -- jest && npm run docker:down",
+    "job:update_users_auth0": "dotenv -e .env -- ts-node-dev jobs/update_users_auth0.ts"
```
davidkircos, line 19: remove

quadratic-api/src/routes/files/files.ts
davidkircos: Yeah let's remove it.

src/actions.ts
```diff
@@ -0,0 +1,139 @@
...
+  // ]
+  //
+  // shortcut: Shortcut[] | Shortcut
+};
+
+export const isOwner = (permission: Permission) => permission === OWNER;
+export const isEditorOrAbove = (permission: Permission) => permission === EDITOR || isOwner(permission);
+export const isViewerOrAbove = (permission: Permission) => permission === VIEWER || isEditorOrAbove(permission);
+
+export const createNewFile = {
```
davidkircos, line 53: This permission should be tied to isAuthenticated ?

src/api/fetchFromApi.ts
```diff
@@ -31,6 +32,7 @@ export async function fetchFromApi<T>(path: string, init: RequestInit, schema: z
   // Compare the response to the expected schema
   const result = schema.safeParse(json);
   if (!result.success) {
+    console.error(result.error);
```
davidkircos, line 35: Did you mean to throw?

src/actions.ts
jimniels: That's one way to do it. But if you're a `VIEWER` or above, it means you are authenticated. So it's the same thing. And given that all our conditional rendering is based on the `permission` value, i'm using that everywhere.

davidkircos (approved):

PR #660, opened 2023-08-23
Share Files (read only) - API

davidkircos:
- [x] Unauthenticated user can access GET file/:uuid - will return file+permissions or 404 or 403
- [x] Authenticated user and file owner OWNER can POST file/:uuid to update public_link_access to either READONLY or NOT_SHARED


quadratic-api/src/routes/files/files.ts
```diff
@@ -1,18 +1,22 @@
...
 import dbClient from '../../dbClient';
-import { validateAccessToken } from '../../middleware/auth';
-import { userMiddleware } from '../../middleware/user';
+import { userMiddleware, userOptionalMiddleware } from '../../middleware/user';
+import { validateAccessToken } from '../../middleware/validateAccessToken';
+import { validateOptionalAccessToken } from '../../middleware/validateOptionalAccessToken';
 import { Request } from '../../types/Request';
 
-const files_router = express.Router();
+type FILE_PERMISSION = 'OWNER' | 'READONLY' | 'EDIT' | 'NOT_SHARED' | undefined;
```
jimniels, line 10: We decided to remove `undefined` here, right? If `public_link_access` is in the response, it should always have a value, right?

jimniels (approved):

quadratic-api/src/routes/files/files.ts
davidkircos: Yeah, you are right, it is never undefined

PR #661, opened 2023-08-23
Update LICENSE

davidkircos:


davidfig:
quadratic-core/package.json needs to remove license

davidkircos:
> quadratic-core/package.json needs to remove license

There is no quadratic-core/package.json it's located in the build output `quadratic-core/pkg/package.json` which should no longer include the license because I believe it was built from Cargo.toml

jimniels (approved):

davidfig (approved):

PR #664, opened 2023-08-24
Next fractional indexing

davidfig:


quadratic-core/src/grid/mod.rs
```diff
@@ -172,41 +204,52 @@ impl Grid {
...
-        self.rebuild_sheet_id_map();
-    }
-    fn rebuild_sheet_id_map(&mut self) {
-        self.sheet_ids = IdMap::new();
-        for (i, sheet) in self.sheets.iter().enumerate() {
-            self.sheet_ids.add(sheet.id, i);
+            let after_sheet = self.sheet_from_id(to_before.unwrap());
+            let before = self.previous_sheet_order(after_sheet.id);
+            order = key_between(&before, &Some(after_sheet.order.clone())).unwrap();
         }
```
HactarCE, line 230: Since invalid sheet IDs are a possibility, we should make `sheet_from_id()` return `Option<Sheet>`. Then this code can be rewritten:
```rs
match to_before.and_then(|id| self.sheet_from_id(id)) {
  None => {
    let last_order = self.sheets.last().map(|last| last.order.clone());
    order = key_between(&last_order, &None).unwrap();
  },
  Some(after_sheet) => ...,
}
```

quadratic-core/src/grid/mod.rs
```diff
@@ -172,41 +204,52 @@ impl Grid {
...
+    /// Moves a sheet before another sheet
+    pub fn move_sheet(&mut self, target: SheetId, to_before: Option<SheetId>) {
+        let order: String;
+        // treat to_before as None if to_before's sheet no longer exists
+        if to_before.is_none() || !self.sheet_has_id(to_before) {
+            let last_order = match self.sheets.last() {
+                Some(last) => Some(last.order.clone()),
+                None => None,
+            };
+            order = key_between(&last_order, &None).unwrap();
```
HactarCE, line 225: In what cases does `key_between()` fail, and is that something we care about handling?

quadratic-core/src/grid/mod.rs
```diff
@@ -172,41 +204,52 @@ impl Grid {
...
+        self.sort_sheets();
     }
     pub fn sheet_id_to_index(&self, id: SheetId) -> Option<usize> {
-        self.sheet_ids.index_of(id)
+        self.sheets.iter().position(|sheet| sheet.id == id)
     }
     pub fn sheet_index_to_id(&self, index: usize) -> Option<SheetId> {
-        self.sheet_ids.id_at(index)
+        let sheet = self.sheets.get(index);
+        match sheet {
```
HactarCE, line 240:
```rs
self.sheets.get(index)?.id
```

quadratic-core/src/grid/mod.rs
```diff
@@ -172,41 +204,52 @@ impl Grid {
...
+            Some(sheet) => Some(sheet.id),
+            None => None,
+        }
+    }
+    pub fn sheet_has_id(&self, sheet_id: Option<SheetId>) -> bool {
+        if sheet_id.is_none() {
+            return false;
+        }
```
HactarCE, line 248:
```rs
let Some(sheet_id) = sheet_id else { return false };
```
Then you don't need to `.unwrap()` later.

quadratic-core/src/grid/mod.rs
```diff
@@ -172,41 +204,52 @@ impl Grid {
...
+        match &self.sheets.iter().position(|s| s.id == sheet_id.unwrap()) {
+            Some(_) => true,
+            None => false,
+        }
```
HactarCE, line 252: `.is_some()` instead of `match`

quadratic-core/src/controller.rs
```diff
@@ -129,19 +126,24 @@ impl GridController {
...
     ) -> TransactionSummary {
-        let ops = vec![Operation::DeleteSheet { sheet_id }];
+        let mut ops = vec![Operation::DeleteSheet { sheet_id }];
+        if self.sheet_ids().len() == 1 {
+            let id = SheetId::new();
+            let name = String::from("Sheet 1");
+            let order = self.grid.end_order();
+            let sheet = Sheet::new(id, name, order);
+            ops.push(Operation::AddSheet { sheet });
+        }
```
HactarCE, line 146: At some point the server will have to ensure that there's always at least one sheet, since clients can conspire to delete all the sheets at once without replacing them. This is fine for now tho.

quadratic-core/src/controller.rs
```diff
@@ -161,17 +163,17 @@ impl GridController {
...
         new_sheet.name = format!("{} Copy", new_sheet.name);
-        let ops = vec![Operation::AddSheet {
-            sheet: new_sheet,
-            to_before: sheet_after,
-        }];
+        let right = self.grid.next_sheet(sheet_id);
+        let right_order = match right {
+            Some(right) => Some(right.order.clone()),
+            None => None,
+        };
```
HactarCE, line 174: `let right_order = right.map(|right| right.order.clone());`

quadratic-core/src/controller.rs
```diff
@@ -277,51 +279,36 @@ impl GridController {
...
                         summary.sheet_list_modified = true;
-                        rev_ops.push(Operation::AddSheet {
-                            sheet,
-                            to_before: old_after,
-                        });
+                        rev_ops.push(Operation::AddSheet { sheet });
                     }
                 }
 
                 Operation::ReorderSheet { target, to_before } => {
```
HactarCE, line 301: `Operation::ReorderSheet` should probably have `order: String` instead of `to_before: SheetId`

quadratic-core/src/grid/mod.rs
davidfig: It will only fail if `order` is invalid. I expect panic is the right move in that case. We can validate order string on load or something like that down the road.

PR #667, opened 2023-08-25
Next breakout controller

davidfig:
I'm not sure if this is the best way to do it, but `controller.rs` was getting a bit long. I tried to break it up into separate files.


HactarCE (approved):
Looks good!

PR #668, opened 2023-08-25
fix: redirect to login when getting token silently fails

jimniels:
Was seeing this issue when you come back to the app after a day or so:

CleanShot 2023-08-25 at 13 40 02@2x

The app renders because `isAuthenticated` comes back as true.

_But_, the `.getToken()` fails because, I believe, it has expired. Because of this, the app renders but the part of the UI that relies on an API call fails because the auth client fails to get the token.

The best way to simulate this is to login and change the expires time to something in the past (e.g. `0`). Then reload the app and you'll see the same error screen above.

CleanShot 2023-08-25 at 13 44 46@2x

This fixes this problem by redirecting the user to login if the `getToken` call fails

### DIFF
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -105,7 +105,6 @@ export const router = createBrowserRouter(
       <Route
         path={ROUTES.LOGIN}
         loader={async ({ request }) => {
-          document.title = 'Login - Quadratic';
           let isAuthenticated = await authClient.isAuthenticated();
 
           // If they’re authenticated, redirect home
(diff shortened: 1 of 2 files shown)


src/router.tsx
```diff
@@ -125,6 +126,7 @@ export const router = createBrowserRouter(
       <Route
         path={ROUTES.LOGIN_RESULT}
         loader={async () => {
+          document.title = 'Login result - Quadratic';
```
jimniels, line 129: There's a dance that happens when you're not authenticated and it shows up in the browser history.

CleanShot 2023-08-25 at 13 20 57@2x

I'd like to clean this up one day — if possible — but it's not immediately apparent to me how you do it, so for another day.

src/router.tsx
```diff
@@ -136,8 +136,8 @@ export const router = createBrowserRouter(
             }
```
davidkircos, line 136: @jimniels can this path happen? if no error is thrown and isAuthenticated is false, then we don't have a return path

davidkircos (approved):

PR #680, opened 2023-09-01
feat: file actions on dashboard list

jimniels:
## File-related actions from the dashboard

On desktop we surface a few of the actions:

CleanShot 2023-09-06 at 14 43 16@2x

And on smaller screens we hide all actions under the dropdown:

CleanShot 2023-09-06 at 14 43 06@2x

## "Share File" menu re-used across dashboard and app

- The share menu now fetches its data from the `/file/:uuid/sharing` endpoint on-demand.
- It handles changes optimistically (by immediately rendering the change and syncing to the server).
- It shows the filename when on the dashboard (for additional context)

CleanShot 2023-09-13 at 16 51 35@2x

CleanShot 2023-09-13 at 16 51 26@2x

## Reusable "Dialog" component for dialogs 

Share & feedback use the same underlying component for displaying their content:

image

image

## Todos

- [x] Optimistically render actions as they take place
  - Delete - immediately removed from list (pops back in if API call fails)
  - Duplicate - immediately add new file to list (disappears if API call fails w/UI message)
  - Rename - immediately change file name (changes to old name if it fails w/UI message)
  - Download - does nothing, just waits for browser to download file
  - Share - immediately show share menu, skeleton loaders for incoming data
- [x] Keep renames in place by not updating the `updated_date` (see [slack thread](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1694035470838799))
- [x] Implement `<ShareFileMenu>` across the app (fetch its own data, or pass it in manually now...)
- [x] Mobile/responsive UI


src/dashboard/components/DashboardFileLink.tsx
```diff
@@ -39,31 +48,31 @@ export function DashboardFileLink({
...
+              {name}
             </Typography>
+            <Stack direction="row" gap={theme.spacing(1)}>
+              <Typography variant="caption" color="text.secondary">
+                {isShared && 'Public · '} {description}
+              </Typography>
+              {descriptionError && (
+                <Typography variant="caption" color="error">
+                  {descriptionError}
+                </Typography>
```
jimniels, line 70: @jimniels make these all separated by `·`

src/router.tsx
```diff
@@ -92,6 +102,8 @@ export const router = createBrowserRouter(
             element={<CloudFilesMigration.Component />}
             loader={CloudFilesMigration.loader}
           />
+
+          <Route path="/api/files/:uuid/sharing" action={shareFileMenuAction} loader={shareFileMenuLoader} />
```
jimniels, line 106: I'm not quite sure what to call this route. These are called "resource routes" in Remix, but that's less of a thing in react-router because it's all client side.

The idea is: we need URLs that react-router fetchers can access without doing navigations. So it's sort of like a client-side, react-router specific API. So for now i'm just having these mirror the actual server API routes.

davidkircos:
- [ ] Copy link button does not work on list view (copies from URL bar which is not the file URL)
- [ ] Add "Share" and "Delete" to ... menu on the dashboard.
- [ ] The Cannot View / Can view drop-down adds a new layer of background color

jimniels:
@davidkircos those have all been fixed (also, lmk what you think about the show/hide effect for each row item's icons, similar to google docs)

src/router.tsx
davidkircos: Do you actually need a whole route, or do you just need a fetcher?

src/ui/QuadraticUI.tsx
```diff
@@ -72,7 +74,21 @@ export default function QuadraticUI({ app, sheetController }: { app: PixiApp; sh
...
+      {editorInteractionState.showShareFileMenu && (
+        <ShareFileMenu
+          onClose={() => {
+            setEditorInteractionState((prevState) => ({
+              ...prevState,
+              showShareFileMenu: false,
+            }));
+            // TODO the button that triggers the share menu is getting focus, not the
+            // grid, even when you run this
+            focusGrid();
```
davidkircos, line 86: I think you may just need to put in a short delay.

davidkircos (approved):
Two comments on the code. I am not sure if they marit any changes.

src/router.tsx
jimniels: If we were fetching directly from the API with our fetchers, then we wouldn't need a route. 

But we have the `apiClient` for talking to the API, so we need client routes for the loaders/fetchers, and the client routes use the `apiClient` to talk to the API. So in this case, yes we need it.

Maybe in the future we can hit the API directly from our fetchers, but I haven't really thought about how that changes the architecture of the current codebase. So this was the most direct method to get this working.

PR #681, opened 2023-09-02
Rust implementation of clipboard

davidfig:
- [x] adds copy/paste functionality to rust
- [x] adds clear formatting to rust
- [x] adds copying/cutting/replacing column_data blocks
- [x] adds a bit of testing to block.rs
- [x] replaces TS copy/paste functionality
- [x] replaces TS clear formatting functionality


davidkircos:
I haven't really reviewed it, but it is very fast!

quadratic-core/src/controller/cells.rs
```diff
@@ -73,6 +73,33 @@ impl GridController {
...
+        cursor: Option<String>,
+    ) -> TransactionSummary {
+        let region = self.existing_region(sheet_id, rect);
+        let ops = match region.size() {
+            Some(_) => {
+                // need an empty list of columns to clear the range
+                let mut formats = vec![];
+                for _ in rect.x_range() {
+                    formats.push(Column::new());
+                }
```
HactarCE, line 89: Careful! `region` may be smaller than `rect` if some of the columns/rows don't exist, because we called `.existing_region()`. Destructure `region.size()` into `Some(region_size)`, then `let formats = vec![Column::new(); region_size.w];`

quadratic-core/src/controller/transactions.rs
```diff
@@ -196,6 +228,15 @@ pub enum Operation {
...
     },
+    ReplaceCellFormats {
+        sheet_id: SheetId,
+        pos: Pos,
+        formats: Vec<Column>,
+        height: u32,
+
+        // true for an undo; false for a paste
+        use_column_ids: bool,
+    },
```
HactarCE, line 239: This needs to store all of `region` and not infer the rectangle; if columns or rows move around or are deleted, then `pos` is useless.

This logic is already implemented in all the `SetCellFormats` handling, so I'd rather it not be duplicated. a couple options:
a) make `SetCellFormats` contain an optional array for _all_ formatting types.
b) submit multiple `SetCellFormats` operations when replacing formatting.

quadratic-core/src/grid/column.rs
```diff
@@ -257,6 +310,61 @@ impl<B: BlockContent> ColumnData<B> {
...
+    pub fn remove_range_to_column_data(
+        &mut self,
+        y_range: &Range<i64>,
+        column_data: &mut ColumnData<B>,
+    ) {
+        let mut blocks = self.remove_range(y_range.clone());
+        if blocks.len() != 0 {
+            blocks.iter_mut().for_each(|block| block.y -= y_range.start);
+            column_data.add_blocks(blocks);
+        }
```
HactarCE, line 322: no need to check whether `blocks.len() != 0` -- these functions will be fast in that case too.

quadratic-core/src/controller/transactions.rs
```diff
@@ -255,3 +297,16 @@ pub struct TransactionSummary {
...
+        TransactionSummary {
+            cell_regions_modified: vec![],
+            fill_sheets_modified: vec![],
+            code_cells_modified: vec![],
+            border_sheets_modified: vec![],
+            cursor: None,
+            sheet_list_modified: false,
+        }
+    }
+}
```
HactarCE, line 312: `impl Default for TransactionSummary { }` and then use autocomplete to fill in the implementation.

But also, in this case we can use a `#[derive(Default)]` since all the fields are just defaults for their own types. Good idea to add a `default()`!

HactarCE (changes requested):
See comments; we probably need to use `RegionRef` (or something equivalent) instead of `Rect` to keep values associated with column & row IDs in case they move around.

quadratic-core/src/controller/cells.rs
```diff
@@ -133,6 +133,82 @@ impl GridController {
...
+    }
+
+    pub fn delete_values_and_formatting(
+        &mut self,
+        sheet_id: SheetId,
+        rect: Rect,
+        cursor: Option<String>,
+    ) -> TransactionSummary {
+        let mut ops = self.delete_cell_values_operations(sheet_id, rect);
+        ops.append(&mut self.clear_formatting_operations(sheet_id, rect));
```
HactarCE, line 211: `ops.extend(self.clear_formatting...)` is slightly cleaner (`&mut` right before throwing away a value feels a little weird)

quadratic-core/src/grid/sheet.rs
```diff
@@ -213,6 +213,23 @@ impl Sheet {
...
 
+    pub fn get_code_cell_value(&self, pos: Pos) -> Option<CellValue> {
+        let column = self.get_column(pos.x);
+        if column.is_none() {
+            return None;
+        }
+        let block = column.unwrap().spills.get(pos.y);
+        if block.is_none() {
+            return None;
+        }
```
HactarCE, line 224:
```rs
let column = self.get_column(pos.x)?;
let block = column.spills.get(pos.y)?;
```
no `.unwrap()` needed :)

HactarCE (changes requested):
very good! just 1 or 2 small changes and then I'll approve

HactarCE (approved):
yay!

PR #682, opened 2023-09-04
Next currency

davidfig:
This is ready for review. I decided not to pass CellValue to TS to avoid the ts_rs/TS issue. TS only needs string representations.

- [x] updated .fnt generating script to generate all unicode characters in a font
- [x] currency is converted to a number and proper formatting during user input
- [x] use bigdecimal for number values
- [x] moved formatting from cells.rs to formatting.rs 
- [x] change decimal place 
- [x] format as percentage or currency
- [x] decimal places increase/decrease


quadratic-core/src/controller/cells.rs
```diff
@@ -33,8 +50,25 @@ impl GridController {
         let sheet = self.grid.sheet_mut_from_id(sheet_id);
         let cell_ref = sheet.get_or_create_cell_ref(pos);
         let region = RegionRef::from(cell_ref);
-        let values = Array::from(value);
-        let ops = vec![Operation::SetCellValues { region, values }];
+        let mut ops = vec![];
+        if let Some((currency, number)) = Self::is_currency(&value) {
```
HactarCE, line 54: Perhaps this method should take `value: String` instead and then also convert to number? If you don't think we should do that right now, then this is fine as-is.

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
         sheet.with_random_floats(region);
     }
 
-    pub fn grid(&self) -> &Grid {
-        &self.grid
+    /// tests whether a a CellValue::Text is a currency value
+    fn is_currency(value: &CellValue) -> Option<(String, f64)> {
+        if value.type_name() != "text" {
+            return None;
+        }
```
HactarCE, line 25: `if matches!(value, CellValue::Text(_))` or `if let CellValue::Text(_) = value` are preferred since they'll both error in case we ever change `"text"` to something else.

HactarCE: Even better here is `let CellValue::Text(s) = value else { return None; }` since we need `s` later on.

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
+        let s = value.to_string();
+        if s.len() == 0 {
```
HactarCE, line 27: `if s.is_empty()`

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
+            return None;
+        }
+        let possible_currency = s.chars().next().unwrap();
```
HactarCE, line 30: `let Some(first_char) = s.chars().next()?;` will return `None` if there is no first character, and then you don't even need the check above.

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
+        if CURRENCY_SYMBOLS.contains(possible_currency) {
+            let split: Vec<&str> = s.split(possible_currency).collect();
```
HactarCE, line 32: Use `if let Some(number) = s.strip_prefix(possible_currency)`.

Consider rewriting this whole logic to loop over the currency symbols (which will happen anyway in `.contains()`) and try `.strip_prefix()` on each one.

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
+            if split.len() == 2 {
```
HactarCE, line 33: In general, prefer `if let Some(number) = splits.get(1)`

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
 impl GridController {
     pub fn populate_with_random_floats(&mut self, sheet_id: SheetId, region: &Rect) {
         let sheet = self.grid.sheet_mut_from_id(sheet_id);
         sheet.with_random_floats(region);
     }
 
-    pub fn grid(&self) -> &Grid {
-        &self.grid
+    /// tests whether a a CellValue::Text is a currency value
+    fn is_currency(value: &CellValue) -> Option<(String, f64)> {
```
HactarCE, line 22: I would prefer a name like `unpack_currency()`. Reading `is_.*` makes me think the function will return a boolean.

quadratic-core/src/grid/formatting.rs
```diff
@@ -108,6 +108,17 @@ pub struct NumericFormat {
...
+impl NumericFormat {
+    pub fn new(kind: NumericFormatKind, decimals: Option<u32>, symbol: Option<String>) -> Self {
+        NumericFormat {
+            kind,
+            decimals,
+            symbol,
+        }
+    }
+}
+
```
HactarCE, line 121: This feels like a plain struct to me that doesn't have any implementation details to hide -- ok to just make these fields public and construct it with `NumericFormat { ... }` instead of using `NumericFormat::new(...)`?

HactarCE (changes requested):
I left some feedback. otherwise looks good!

quadratic-core/src/controller/cells.rs
```diff
@@ -1,28 +1,45 @@
...
+        let possible_currency = s.chars().next().unwrap();
+        if CURRENCY_SYMBOLS.contains(possible_currency) {
+            let split: Vec<&str> = s.split(possible_currency).collect();
+            if split.len() == 2 {
+                if let Ok(number) = split[1].parse::<f64>() {
+                    return Some((possible_currency.to_string(), number));
+                }
+            }
+        }
+        None
```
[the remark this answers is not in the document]
davidfig: Good catch. Moved to CellValue.

quadratic-core/src/controller/cells.rs
davidfig: That makes more sense.

quadratic-core/src/values/convert.rs
```diff
@@ -111,7 +116,8 @@ impl<'a> TryFrom<&'a CellValue> for f64 {
                     got: Some(value.type_name().into()),
                 })
             }
-            CellValue::Number(n) => Ok(*n),
+            // todo: this may be wrong
+            CellValue::Number(n) => Ok(n.to_f64().unwrap()),
```
HactarCE, line 120: Actually we should change this whole impl to `impl<'a> TryFrom<&'a CellValue> for BigDecimal` and then change all the formula functions to operate on `BigDecimal` ...

HactarCE: I can take on that in another PR. In the meantime, formulas just won't preserve precision well.

quadratic-core/src/values/convert.rs
```diff
@@ -145,7 +151,7 @@ impl<'a> TryFrom<&'a CellValue> for bool {
             CellValue::Blank => Ok(false),
             CellValue::Text(s) if s.eq_ignore_ascii_case("TRUE") => Ok(true),
             CellValue::Text(s) if s.eq_ignore_ascii_case("FALSE") => Ok(false),
-            CellValue::Number(n) => Ok(*n != 0.0),
+            CellValue::Number(n) => Ok(n.ne(&BigDecimal::from_str(&"0.0").unwrap())),
```
HactarCE, line 154: [`n.is_zero()`](https://docs.rs/bigdecimal/latest/bigdecimal/struct.BigDecimal.html#method.is_zero)

quadratic-core/src/values/mod.rs
```diff
@@ -17,10 +17,10 @@ pub use convert::CoerceInto;
 pub use isblank::IsBlank;
 pub use time::{Duration, Instant};
 
-use crate::{CodeResult, CodeResultExt, ErrorMsg, Span, SpannableIterExt, Spanned};
+use crate::{CodeResult, CodeResultExt, ErrorMsg, SpannableIterExt, Spanned};
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-#[cfg_attr(feature = "js", derive(ts_rs::TS))]
+// #[cfg_attr(feature = "js", derive(ts_rs::TS))]
```
HactarCE, line 23: remove

quadratic-core/src/values/mod.rs
```diff
@@ -85,14 +85,14 @@ impl Value {
...
-        }
-    }
+    // / Replaces NaN and Inf with errors; otherwise returns the value
+    // / unchanged.
+    // pub fn purify_floats(self, span: Span) -> CodeResult<Self> {
+    //     match self {
+    //         Value::Single(v) => v.purify_float(span).map(Value::Single),
+    //         Value::Array(a) => a.purify_floats(span).map(Value::Array),
+    //     }
+    // }
```
HactarCE, line 95: same as other one, not needed anymore!

quadratic-core/src/wasm_bindings/mod.rs
```diff
@@ -225,7 +226,7 @@ impl GridProxy for JsGridProxy {
             None => return CellValue::Blank,
         };
 
-        if let Ok(n) = string.parse::<f64>() {
+        if let Some(n) = BigDecimal::parse_bytes(&string.clone().into_bytes(), 10) {
```
HactarCE, line 229: `BigDecimal::from_str_radix(&string, 10)`

HactarCE (changes requested):

quadratic-core/src/bin/export_types.rs
```diff
@@ -16,63 +18,51 @@ fn main() {
...
-        // grid/code.rs
-        grid::CodeCellLanguage,
-        grid::CodeCellValue,
-        grid::CodeCellRunOutput,
-        grid::CodeCellRunResult,
-        // grid/formatting.rs
+        // grid::CodeCellLanguage,
+        // grid::CodeCellValue,
+        // grid::CodeCellRunOutput,
+        // grid::CodeCellRunResult,
```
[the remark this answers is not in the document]
davidfig: We actually need some of them. I'll pare them down and leave what's needed. (I cheated and used some any's as a short-term workaround.)

quadratic-core/src/grid/sheet.rs
```diff
@@ -616,6 +651,44 @@ impl Sheet {
...
+        } else {
+            None
+        } {
+            Some(numeric_decimals)
+        } else {
+            None
+        };
+        if decimals.is_some() {
+            return decimals;
+        }
```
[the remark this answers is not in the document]
davidfig: Wow. That was pretty terrible. Much better now.

quadratic-core/src/values/array.rs
```diff
@@ -82,10 +83,14 @@ impl Array {
...
     pub fn from_random_floats(size: ArraySize) -> Self {
         let mut rng = rand::thread_rng();
-        let values =
-            std::iter::from_fn(|| Some(CellValue::Number(rng.gen_range(-100..=100) as f64)))
-                .take(size.len())
-                .collect();
+        let values = std::iter::from_fn(|| {
+            Some(CellValue::Number(
+                BigDecimal::parse_bytes(&rng.gen_range(-100..=100).to_string().into_bytes(), 10)
+                    .unwrap(),
```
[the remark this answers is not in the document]
davidfig: I think I was having problems with ::from working earlier (for some reason). Much better.

PR #683, opened 2023-09-05
fix: conditionally load python on mobile devices

jimniels:
What we want to do is ensure that python only gets loaded for:

1. Desktop users
2. Users who have owner or editor rights

To ensure this happens, we need to:

- [x] Use the `isMobile` flag to not load python on phones/tablets
- [x] Check the `permission` that comes back from the API and use `isEditorOrAbove(permission)` to conditionally load python

### DIFF
--- a/src/atoms/loadedStateAtom.ts
+++ b/src/atoms/loadedStateAtom.ts
@@ -1,11 +1,11 @@
 import { atom } from 'recoil';
 
 export interface LoadedState {
-  pythonLoadState: 'initial' | 'loading' | 'loaded' | 'skipped' | 'error';
+  pythonLoaded: boolean | 'error';
 }
 
 export const loadedStateDefault: LoadedState = {
-  pythonLoadState: 'initial',
+  pythonLoaded: false,
 };
 
 export const loadedStateAtom = atom({
--- a/src/atoms/loadedStateAtom.ts
+++ b/src/atoms/loadedStateAtom.ts
@@ -1,11 +1,11 @@
 import { atom } from 'recoil';
 
 export interface LoadedState {
-  pythonLoaded: boolean | 'error';
+  pythonLoadState: 'initial' | 'loading' | 'loaded' | 'skipped' | 'error';
 }
 
 export const loadedStateDefault: LoadedState = {
-  pythonLoaded: false,
+  pythonLoadState: 'initial',
 };
 
 export const loadedStateAtom = atom({
--- a/src/atoms/loadedStateAtom.ts
+++ b/src/atoms/loadedStateAtom.ts
@@ -1,11 +1,11 @@
 import { atom } from 'recoil';
 
 export interface LoadedState {
-  pythonLoaded: boolean | 'error';
+  pythonLoadState: 'initial' | 'loading' | 'loaded' | 'error';
 }
 
 export const loadedStateDefault: LoadedState = {
-  pythonLoaded: false,
+  pythonLoadState: 'initial',
 };
 
 export const loadedStateAtom = atom({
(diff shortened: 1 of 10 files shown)


src/web-workers/pythonWebWorker/python.worker.ts
```diff
@@ -22,6 +23,13 @@ const getCellsDB = async (x0: number, y0: number, x1: number, y1: number): Promi
 let pyodide: any | undefined;
 
 async function pythonWebWorker() {
+  if (isMobile) {
+    self.postMessage({ type: 'python-loading-skipped' } as PythonMessage);
+    return;
+  }
```
jimniels, line 29: Presumably right here is where we need to say something like 

```
if (isMobile || isEditorOrAbove(pixi.app.settings.editorInteractionState.permission) {
  // return early and don't load python
}
```

The trick is: i'm not sure A) how to access that data from here, and B) if we access the `permission` state too early we're not going to get the right value — it's like we need to wait for the initial API call that tells us the user's rights to the document _before_ we initialize the web worker.

src/atoms/loadedStateAtom.ts
```diff
@@ -1,11 +1,11 @@
 import { atom } from 'recoil';
 
 export interface LoadedState {
-  pythonLoaded: boolean | 'error';
+  pythonLoadState: 'initial' | 'loading' | 'loaded' | 'error';
```
jimniels, line 4: Because this was previously just `false | true | 'error'` the baked-in assumption is that python starts by trying to load (e.g. if the value is `false` that means it's currently loading)

But we don't want that because it's not loading by default. It might never load at all. This was causing people who aren't logged in but looking at a sheet to see things like "Python is loading" if they opened the editor

CleanShot 2023-09-06 at 11 04 50@2x

So I had to change this to expand the scope of possible states for python loading. By default it doesn't try to load, but [if the proper conditions are met](https://github.com/quadratichq/quadratic/pull/683/files#diff-0477e2c4c9e88b95f3a82bdecb75fdf07e66777d96f9c3ff32f208faded5fcadR55-R58), this state gets changed to `loading` and the web worker fires up.

src/dashboard/FileRoute.tsx
```diff
@@ -74,8 +82,17 @@ export const loader = async ({ request, params }: LoaderFunctionArgs): Promise<F
...
+  const initializeState = ({ set }: MutableSnapshot) => {
+    set(editorInteractionStateAtom, (prevState) => ({
+      ...prevState,
+      permission,
+    }));
+  };
+
   return (
-    <RecoilRoot>
+    <RecoilRoot initializeState={initializeState}>
```
jimniels, line 95: Because [we initialize this with a static value from the recoil atom](https://github.com/quadratichq/quadratic/pull/683/files#diff-ce68af5e02ad903f6c16e3275180375cbf37046dc8e1d7268b29f7e6fac71231R25), we actually have the wrong value when the app first mounts.

I knew this was a small bug but it didn't bite us because we didn't really need this permissioning when the app first mounts.

But now that we need the right value to know whether to load python, I figured out how to initialize the `editorInteractionState` atom with the proper value when we first get it from the server.

src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -273,6 +273,17 @@ export const CodeEditor = (props: CodeEditorProps) => {
...
+    if (!(event.metaKey || event.ctrlKey) && event.key === 'Escape') {
+      event.preventDefault();
+      closeEditor();
+    }
+
+    // Don't allow the shortcuts below for certain users
+    if (!isEditorOrAbove(editorInteractionState.permission)) {
+      return;
+    }
+
```
jimniels, line 286: For people who can't run python, we hide the UI to hit the "Run" button, but we also need to disallow the keyboard shortcuts.

CleanShot 2023-09-06 at 11 44 28@2x

davidfig (approved):

PR #686, opened 2023-09-06
feat: add signup button, fix 'back' button on login

jimniels:
## Sign up

For users viewing a shared file who are not authenticated, we show a message saying they need to login.

If they hit "Log in" they get taken to the auth0 login page:

CleanShot 2023-09-06 at 16 37 52@2x

Where they can login, or if they are new, sign up.

This PR adds an additional "Sign up" button

CleanShot 2023-09-06 at 16 23 13@2x

Which allows them to jump directly to the sign up page:

CleanShot 2023-09-06 at 16 37 37@2x

## 'Back' button on login

Previous behavior:

- Load a shared file as a logged out user
- Click "log in"
- [app sends user to `/login` which then redirects them to auth0]
- Hit the back button
- End up on auth0 again

When you hit the back button, it's navigating you to `/login` which automatically redirects you back to auth0. So you get stuck in this loop where hitting "back" doesn't take you back to the file.

New behavior:

- Load a shared file as a logged out user
- Click "log in"
- [app sends user to `/login` which then redirects them to auth0]
- Hit the back button
- End up on the shared file

This properly replaces the `/login` entry in the history stack so when the user hits "log in" and ends up on auth0, they can hit 'back' in the browser and end up back on the shared file they were just looking at.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -40,7 +40,7 @@ async function getClient() {
 interface AuthClient {
   isAuthenticated(): Promise<boolean>;
   user(): Promise<undefined | User>;
-  login(redirectTo: string): Promise<void>;
+  login(redirectTo: string, isSignupFlow?: boolean): Promise<void>;
   handleSigninRedirect(): Promise<void>;
   logout(): Promise<void>;
   getToken(): Promise<string | void>;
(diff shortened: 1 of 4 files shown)


davidkircos:
Switch which button is outlined and which one is solid to make "Signup" the more pronounced button:
image

davidkircos (approved):

PR #687, opened 2023-09-08
feat: notification when app needs update

jimniels:
**What is this?**

Trigger a notification for the user when the app is out of date, e.g.

CleanShot 2023-09-08 at 17 03 30@2x

**How does it work?**

Every time the app changes, the "version" of the app is the current commit hash. This gets injected into the app in various ways (e.g. the footer)

CleanShot 2023-09-08 at 17 08 25@2x

It also gets put in the `index.html` file as a meta tag. And the app sets up an interval to poll the root `index.html` file and look for the app version. If it matches the current version, nothing happens. If it doesn't match that means the app has changed and we trigger a notification to the user to reload (and it continues to notify them every time the polling interval triggers).

Question: will this muck up our analytics for requests to `index.html`? Because everyone is going to be requesting that file many times to check for the newest version...

**Todos**

- [ ] Notification needs a "Reload" button
- [ ] Determine proper timing

**To test**

Open the preview app, push a new commit to this branch and watch vercel build the app. Once the app rebuilds and deploys, the app should pop up this notification (within the polling time) because it sees a new version of the app.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -6,6 +6,7 @@
     <meta content="True" name="HandheldFriendly" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0" />
     <meta name="theme-color" content="#FFFFFF" />
+    <meta name="version" content="%REACT_APP_VERSION%" />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
     <!--
       manifest.json provides metadata used when your web app is installed on a
--- a/src/components/UpgradeApp.tsx
+++ b/src/components/UpgradeApp.tsx
@@ -29,7 +29,7 @@ export const UpgradeApp = () => {
             nextVersion
           );
           if (needsUpdate) {
-            // TODO pass custom actions to snackbar so the page will refresh
+            // TODO pass custom action to snackbar so user can click button on notification to refresh
             addGlobalSnackbar('Quadratic has updates! Please reload.');
           }
         });
--- a/src/components/UpgradeApp.tsx
+++ b/src/components/UpgradeApp.tsx
@@ -1,9 +1,7 @@
-import { useState } from 'react';
 import { useInterval } from '../hooks/useInterval';
 import { useGlobalSnackbar } from './GlobalSnackbarProvider';
 
 export const UpgradeApp = () => {
-  const [stopPolling, setStopPolling] = useState<boolean>(false);
   const { addGlobalSnackbar } = useGlobalSnackbar();
 
   useInterval(
@@ -32,12 +30,11 @@ export const UpgradeApp = () => {
           );
           if (needsUpdate) {
             // TODO pass custom actions to snackbar so the page will refresh
-            addGlobalSnackbar('Quadratic has updates! You should reload.');
-            setStopPolling(true);
+            addGlobalSnackbar('Quadratic has updates! Please reload.');
           }
         });
     },
-    stopPolling ? null : 5000 /*3_600_000*/ // one hour
+    10000 // one hour? 3_600_000
   );
 
   return null;
--- a/src/components/UpgradeApp.tsx
+++ b/src/components/UpgradeApp.tsx
@@ -1,0 +1,44 @@
+import { useState } from 'react';
+import { useInterval } from '../hooks/useInterval';
+import { useGlobalSnackbar } from './GlobalSnackbarProvider';
+
+export const UpgradeApp = () => {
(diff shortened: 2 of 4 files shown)


public/index.html
```diff
@@ -6,6 +6,7 @@
     <meta content="True" name="HandheldFriendly" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0" />
     <meta name="theme-color" content="#FFFFFF" />
+    <meta name="version" content="%REACT_APP_VERSION%" />
```
davidkircos, line 9: We don't increment the app version very often, so I don't think this would work well. We could do it with the commit hash, but I don't love this approach because it would be easy to accidentally break.

jimniels: `REACT_APP_VERSION` is the current commit hash. This is what displays in the footer as the "version" of the app.

CleanShot 2023-09-12 at 12 43 25@2x

So every single time we merge something to `main` that's a new "version" of the app (a new commit hash), which makes sense given that we are doing continuous deployment so every single time we have a new commit we have a new version of the app and it goes to prod. 

So this "the app has updated" message works because it keys off this value.

davidkircos: Oh, interesting. Hmm ok. This approach should work.

What if we added a new file called `/public/version.txt` that would just return the commit hash, we could also add some comments into the file that describe how it is used so we know not to break it in the future.

jimniels: A standalone file could work — it helps keep analytics clean on requests to `index.html` — but the thing is it has to be generated at build/start time (otherwise you have to manually maintain it and it's destined to fall out of date)

One way to do that would be to add a `prebuild` script that runs before the build, e.g. inside `package.json`

```json
"scripts": {
  "prestart": "git rev-parse HEAD > public/version.txt",
  "start": "..."
}
```

You'd probably have to duplicate that for `prestart` for local dev too. 

Another option would be to do it in a pre-build script. I actually ended up making on of those for our `env` vars branch. We could so something similar there where we have a script that runs before every build (or local dev server) and it does all the stuff it needs to do, like check for env vars ([like it's doing in this PR](https://github.com/quadratichq/quadratic/pull/650/files#diff-c0fe3a168994c2821f2ef4c98eab2905c5cd66d366dd3bbd289badfbe39a891a)) and then also it generates a `version.txt` file in the build. The nice part about that is you can add as many comments as you want in that file as it's a JS script that runs before every `npm run build | start`

davidkircos:
What fills in %REACT_APP_VERSION% in index.html?

Couldn't you just make a file in public/version.txt with the contents
%REACT_APP_VERSION% ?

jimniels:
I don't think so.

Our project is using `create-react-app` which has [built-in support to inject environment variables in the `public/index.html` file](https://create-react-app.dev/docs/adding-custom-environment-variables#referencing-environment-variables-in-the-html). But that’s using create-react-app’s custom webpack configuration which is setup to do that specifically. From what I can tell, it doesn't do any other file processing for stuff in `public/` to inject vars.

public/index.html
jimniels: @jimniels add a comment around this

src/components/UpgradeApp.tsx
```diff
@@ -0,0 +1,41 @@
...
+        .then((res) => res.text())
+        .then((html) => {
+          const parser = new DOMParser();
+          const dom = parser.parseFromString(html, 'text/html');
+
+          const selector = "meta[name='version']";
+          const nextVersion = dom.querySelector(selector)?.getAttribute('content');
+          const currentVersion = document.querySelector(selector)?.getAttribute('content');
+
+          if (!(nextVersion && currentVersion)) {
```
jimniels, line 19: check that these are strings and they have values that look like shas

src/components/UpgradeApp.tsx
```diff
@@ -0,0 +1,41 @@
...
+          const needsUpdate = currentVersion !== nextVersion;
+          console.log(
+            '[<UpgradeApp>] needs update? %s | current version: %s | next version: %s',
+            needsUpdate,
+            currentVersion,
+            nextVersion
+          );
+          if (needsUpdate) {
+            // TODO pass custom action to snackbar so user can click button on notification to refresh
+            addGlobalSnackbar('Quadratic has updates! Please reload.');
```
jimniels, line 33: Log to sentry if we actually fire this (non-error event)

jimniels:
Given that we have larger ambitions for this, I'm going to close this until we're ready to do those things

PR #690, opened 2023-09-11
Port auto complete to rust

ddimaria:
- [x] Basic shell for the `expand_right` functionality
- [x] Complete `expand_right` functionality for values
- [x] Add tests for `expand_right` functionality for values
- [x] Repeat for `expand_up`, `expand_left`, and `expand_down`
- [x] Complete expand functionality for a series
- [x] Add tests for expand functionality for a series
- [x] Implement `shrink_horizontal` and `shrink_vertical`


davidkircos:
Expanding a series down and right doesn't expand the series to neighboring columns.

Sep-25-2023 14-56-29

davidkircos (changes requested):
- Expand series doesn't have the same behavior as production

Rust:
CleanShot 2023-09-25 at 19 32 48
Prod:
CleanShot 2023-09-25 at 19 33 23

- Does not expand into negative space 

Rust:
CleanShot 2023-09-25 at 19 34 35
Prod:
CleanShot 2023-09-25 at 19 35 12

quadratic-core/src/controller/auto_complete.rs
```diff
@@ -0,0 +1,1014 @@
...
+    Left,
+    Right,
+}
+
+impl GridController {
+    pub fn expand(
+        &mut self,
+        sheet_id: SheetId,
+        mut rect: Rect,
+        range: Rect,
```
davidkircos, line 29: Which one is which? Consider renaming variables.

Chat GPT recommends `start_rect` and `end_rect`

quadratic-core/src/controller/auto_complete.rs
```diff
@@ -0,0 +1,1014 @@
...
+
+            let ops = self.expand_right(sheet_id, &rect, &new_range, down_range, up_range)?;
+            operations.extend(ops);
+        }
+
+        Ok(self.transact_forward(operations, cursor))
+    }
+
+    /// Delete cell values and formats in a given range.
+    pub fn shrink(&mut self, sheet_id: SheetId, delete_range: Rect) -> Vec<Operation> {
```
davidkircos, line 141: The only public function is `expand`? The rest of these are only used as helper functions and should not be called outside of `expand`?

ddimaria: I made them public to be able to be tested by criterion (more info [here](https://bheisler.github.io/criterion.rs/book/user_guide/known_limitations.html)).  I'm happy to change them for now, though I think I may need to benchmark to see the bottleneck you're experiencing in large files in finer detail.

quadratic-core/src/controller/auto_complete.rs
ddimaria: It's the rectangle that represents the selection.  I agree that it's not a great variable name.  `selection` seems to make a lot of sense for the name.  Does that work?

ddimaria: or `selected`

quadratic-core/src/controller/auto_complete.rs
davidkircos: I'm guessing the bottleneck is in JS and not rust

quadratic-core/src/controller/auto_complete.rs
davidkircos: 'selection_start' and 'selection_end'?

ddimaria: Can you elaborate on the reasoning behind the `_start` and `_end`?

quadratic-core/src/test_util.rs
```diff
@@ -24,14 +24,16 @@ pub fn assert_cell_value(
...
+        None
     } else {
-        CellValue::Text(value.to_string())
+        Some(CellValue::Text(value.to_string()))
     };
 
     assert_eq!(
         cell_value,
-        Some(expected.clone()),
+        expected.clone(),
```
HactarCE, line 36: `.clone()` shouldn't be necessary. Equality comparison works on references, and `assert_eq!()` automatically takes a reference. If just `expected` doesn't work, then use `*expected`

quadratic-core/src/wasm_bindings/controller/auto_complete.rs
```diff
@@ -0,0 +1,25 @@
+use std::str::FromStr;
+use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
+
+use crate::{controller::GridController, grid::SheetId, Rect};
+
+#[wasm_bindgen]
+impl GridController {
+    /// Expand the contents of the selected rectangle up to the specified row.
```
HactarCE, line 8: This isn't super clear -- there's no `row` parameter in function, and it should work both horizontally & vertically. Maybe:
> Extend or shrink the contents of `rect` to `range` by inferring patterns.

quadratic-core/src/controller/auto_complete.rs
```diff
@@ -0,0 +1,1028 @@
...
+};
+use crate::{
+    grid::{
+        series::{find_auto_complete, SeriesOptions},
+        RegionRef, Sheet, SheetId,
+    },
+    Array, CellValue, Pos, Rect,
+};
+
+#[derive(PartialEq)]
```
HactarCE, line 16: `Debug`, `Copy`, `Clone`, and `Eq` would be nice, altho we don't need them right now.

quadratic-core/src/controller/auto_complete.rs
```diff
@@ -0,0 +1,1028 @@
...
+            attr: format.clone(),
+        })
+        .collect()
+}
+
+/// In a given rect, collect all cell values into an array.
+///
+/// TODO(ddimaria): determine if this should go in the cell.rs file or equiv
+/// TODO(ddimaria): is this necessary as it's more performant to just pluck the data from the sheet direclty
+pub fn cell_values_in_rect(&selection: &Rect, sheet: &Sheet) -> Result<Array> {
```
HactarCE, line 540: This'll be super useful in formula eval as well, so it should probably be defined directly on `Grid`.

quadratic-core/src/controller/auto_complete.rs
```diff
@@ -0,0 +1,1028 @@
...
+                            None,
+                        );
+                    }
+                }
+
+                count += 1;
+            });
+        });
+
+        (grid_controller.clone(), sheet_id)
```
HactarCE, line 623: Why clone here?

quadratic-core/src/test_util.rs
```diff
@@ -82,12 +90,13 @@ pub fn _assert_cell_format_bold(
...
         x,
-        y
+        y,
+        expect_bold,
+        has_bold
     );
 }
-
 /// Util to print a simple grid to assist in TDD
 pub fn print_table(grid_controller: GridController, sheet_id: SheetId, range: &Rect) {
```
HactarCE, line 101: This should definitely take a reference! We don't want to clone the whole universe just to get some debug output

HactarCE: Also, `Rect` is `Copy` so no need to take a reference

quadratic-core/src/grid/series.rs
```diff
@@ -0,0 +1,637 @@
...
+
+const DAYS_FULL_UPPER: [&str; 7] = [
+    "MONDAY",
+    "TUESDAY",
+    "WEDNESDAY",
+    "THURSDAY",
+    "FRIDAY",
+    "SATURDAY",
+    "SUNDAY",
+];
```
HactarCE, line 76: I think it makes sense to support support lowercase, uppercase, and titlecase for every string series. It might be worth it to find a way to do this without duplicating each series several times, or perhaps not.

quadratic-core/src/grid/series.rs
```diff
@@ -0,0 +1,637 @@
...
+}
+
+pub fn find_string_series(options: SeriesOptions) -> Vec<CellValue> {
+    let mut results: Vec<CellValue> = vec![];
+    let SeriesOptions {
+        ref series,
+        spaces,
+        negative,
+    } = options;
+    let text_series: Vec<Vec<&str>> = vec![
```
HactarCE, line 207: Not need to allocate a new vec for each series. This should work:
```rs
let text_series: &[&[&str]] = &[
    &ALPHABET_LOWER,
    ...
];
```

HactarCE (changes requested):
Overall, looks good! As usual, just a bunch of nitpicks

quadratic-core/src/test_util.rs
ddimaria: Good catch!  This is done.

quadratic-core/src/wasm_bindings/controller/auto_complete.rs
ddimaria: I just changed this var name to match the changes I did in the other file and incorporated your suggestion:

> Extend and/or shrink the contents of selection to range by inferring patterns.

quadratic-core/src/controller/auto_complete.rs
ddimaria: This is a leftover before a refactor into a function.  Removed.

quadratic-core/src/grid/series.rs
ddimaria: I agree.  I think I kept the behavior of the legacy code, but we should be actually ignoring case on a series match IMO.  We could just store the lower case and force lower case on the input when matching?  That would be less code, but slightly less efficient.

quadratic-core/src/test_util.rs
ddimaria: That GridController copy was an oversight, thanks for catching it.  Also, IDK Rect was copy...odd that clippy didn't pick that up. These changes have been made.

quadratic-core/src/grid/series.rs
ddimaria: Thanks, this is done.

quadratic-core/src/controller/auto_complete.rs
davidkircos: Doesn't one Rect represent the starting selection and one represent the ending selection for the expand/autocomplete operation?

ddimaria: The `rect` represents the initial selection (or 1st cursor) and the `range` represents the expansion/shrinkage area _inclusive_ of the `rect`.

davidkircos (approved):

PR #694, opened 2023-09-15
fix: anon users can't double click to edit

jimniels:
This fixes the bug where anonymous users were able to double click a cell and edit it (though changes weren't saved).

It also fixes a bug where double-click/ENTER weren't working for formula/python cells. Now, if you're an anonymous user and you double-click/ENTER on a python cell, the code editor will still open (though you can't edit it).


src/gridGL/interaction/onDoubleClickCanvas.ts
```diff
@@ -1,48 +0,0 @@
...
-    setInteractionState({
-      ...interactionState,
-      ...{
-        showInput: true,
-        inputInitialValue: '',
-      },
-    });
-  }
-  event.preventDefault();
-};
```
jimniels, line 48: I found this file cause I thought it might be where the double click logic was. But nothing was happening. Then I realized `onDoubleClickCanvas` doesn't get imported anywhere so it doesn't look like this file gets used — so i deleted it.

davidfig (approved):

PR #696, opened 2023-09-15
Next Dependency Graph

davidkircos:
Design Reqirements

- Connects a range of cells to a cell as an update dependency
- Can Undo / Redo and retain correctness
- Prevents the addition of cycles
- BFS search (or possibly another type?)
- If rows / columns move the dgraph retains correctness

directed acyclic graph via petgraph

- [x] use Rect instead of Pos?
- [x] Support to modify from Transactions (undo/redo)
- [x] Multi sheets support
- [ ] Load from existing dgraph moved to #700


quadratic-core/src/controller/dependencies.rs
```diff
@@ -0,0 +1,193 @@
...
+                }
+            }
+        }
+
+        seen
+    }
+}
+
+#[cfg(test)]
+mod test {
```
ddimaria, line 49: Nice tests 💯

quadratic-core/src/grid/file/mod.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
```
ddimaria, line 1: If we won't want to pull in the standard library, [hashbrown](https://github.com/rust-lang/hashbrown) is a good alternative.

quadratic-core/src/grid/mod.rs
```diff
@@ -115,10 +126,10 @@ impl Grid {
         self.sort_sheets();
         Ok(id)
     }
-    pub fn remove_sheet(&mut self, sheet_id: SheetId) -> Option<Sheet> {
-        let i = self.sheet_id_to_index(sheet_id)?;
-        let ret = self.sheets.remove(i);
-        Some(ret)
+    pub fn remove_sheet(&mut self, sheet_id: SheetId) -> Sheet {
+        let i = self.sheet_id_to_index(sheet_id);
+        let i = i.expect("sheet_id doesn't exist");
```
ddimaria, line 131: This will panic (not sure that's intentional).  You could return a `Result` instead (the previous `Option` didn't feel like it was the write enum).

ddimaria (approved):
Some non-blocking comments.

quadratic-core/src/grid/file/mod.rs
davidkircos: Since Rust 1.36, hasbrown is the HashMap implementation for the Rust standard library.

quadratic-core/src/grid/mod.rs
davidkircos: I think we want it to panic because this should never happen. Unless we have a bug, the user should not be able to trigger removing a sheet that does not exist.

quadratic-core/src/grid/file/mod.rs
ddimaria: The utility here would be if we didn't want to pull the standard library into the wasm binary.  I think we can punt for now, but we may want to think about file size at some point.

quadratic-core/src/grid/mod.rs
ddimaria: That makes sense.  In that case, you can replace the panic with a more explicit [unreachable](https://doc.rust-lang.org/std/macro.unreachable.html).

PR #700, opened 2023-09-21
File Updating -> Next

davidkircos:
- [x] Collect upgrade js and rust in a single function call `validateAndUpgradeGridFile`


quadratic-core/src/grid/file/v1_5.rs
```diff
@@ -7,44 +7,85 @@ pub use super::current::*; // when creating new version, replace `current` with
...
+                // TODO: validate that sheet name is unique.
+                //       sheet ID is new so it is definitely unique.
+            })
+            .try_collect()?;
+
+        // convert dependencies to Rust format
+        let mut rs_dependencies = HashMap::new();
+        let js_dependencies =
+            serde_json::from_str::<JSDependencySchema>(self.cell_dependency.as_str()).unwrap();
+        for (key, value) in js_dependencies.into_iter() {
```
HactarCE, line 47: `.into_iter()` should not be needed here. `for` loop automatically converts to iterator

quadratic-core/src/grid/file/v1_5.rs
```diff
@@ -7,44 +7,85 @@ pub use super::current::*; // when creating new version, replace `current` with
...
+            let pos = get_value(key).unwrap();
```
HactarCE, line 48: Do not `.unwrap()` here! This will **panic** if the JSON is bad. Instead use `.context("invalid dependency key")?` (or some similar error message

quadratic-core/src/grid/file/v1_5.rs
```diff
@@ -7,44 +7,85 @@ pub use super::current::*; // when creating new version, replace `current` with
...
+            let cell = SheetPos {
+                sheet_id: sheets[0].id,
+                x: pos.0,
+                y: pos.1,
+            };
+            let mut deps: Vec<SheetRect> = vec![];
+            for position in value.into_iter() {
```
HactarCE, line 55: `.into_iter()` is not needed

HactarCE:
```rs
for (x, y) in value {
    let cell = Pos { x, y };
}
```

quadratic-core/src/grid/file/v1_5.rs
```diff
@@ -7,44 +7,85 @@ pub use super::current::*; // when creating new version, replace `current` with
...
 }
 
+fn get_value(key: String) -> Option<(i64, i64)> {
+    let parts: Vec<&str> = key.split(',').collect();
+    if parts.len() != 2 {
+        return None;
+    }
+
+    let x = parts[0].parse::<i64>().ok()?;
+    let y = parts[1].parse::<i64>().ok()?;
```
HactarCE, line 84:
```rs
let [part1, part2] = key.split_once(',')?;

let x = part1.parse().ok()?;
let y = part2.parse().ok()?;
```

HactarCE (changes requested):

davidfig (approved):

PR #701, opened 2023-09-21
fix FloatingContextMenu when switching tabs

davidkircos:
- [x] Fixes FloatingContextMenu for multiple tabs
- [x] Fix: allow using the mouse to select text when editing a tab name.
- [X] Uses text editing cursor when editing Tab Name


jimniels (approved):
Nice good fixes!

PR #707, opened 2023-09-25
fix: proper version comparison

jimniels:
The old version comparison would fail (in the future) for a version comparison like:

"1.10" > "1.8"

This is fixed in this PR with the `compareVersions` function (which includes tests).

### DIFF
--- a/src/schemas/compareVersions.ts
+++ b/src/schemas/compareVersions.ts
@@ -1,0 +1,31 @@
+/**
+ * Takes two version strings and compares whether the first one is greater than,
+ * less than, or equal to the second one.
+ *
+ * Examples:
+ *
+ * ("1.1", "1.2")  ->  -1 (less than)
+ *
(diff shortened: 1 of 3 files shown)


src/schemas/compareVersions.ts
```diff
@@ -0,0 +1,31 @@
...
+    const num2 = parts2[i] || 0;
+
+    if (num1 < num2) {
+      return -1; // less than
+    } else if (num1 > num2) {
+      return 1; // greater than
+    }
+  }
+
+  return 0; // equal
```
davidkircos, line 30: Maybe create an enum type for the result, just to make it more explicit than using numbers.

jimniels: Good call. [Updated](https://github.com/quadratichq/quadratic/pull/707/commits/80649c0f0f16b79de54e12b7fd8ff1d8ae454673)

PR #708, opened 2023-09-25
Next import csv

ddimaria:
- [x] remove `papaparse` from package.json
- [x] Set cursor to positon when inserting CSV so when we undo we get brought back to the location of the CSV
- [x] save is not triggered after import is finished


quadratic-core/src/controller/files.rs
```diff
@@ -0,0 +1,159 @@
...
+                insert_at,
+                Pos {
+                    x: insert_at.x + (width as i64) - 1,
+                    y: insert_at.y + (height as i64) - 1,
+                },
+            );
+            let region = self.region(sheet_id, rect);
+            let ops = vec![Operation::SetCellValues { region, values }];
+
+            return Ok(self.transact_forward(ops, None));
```
davidkircos, line 62: We should set the cursor to something. Maybe

1. Top left cell?
2. Top row?
3. Entire CSV imported?

I am not sure

davidkircos:
remove papaparse from package.json deps

quadratic-core/src/wasm_bindings/controller/mod.rs
```diff
@@ -3,6 +3,7 @@ use crate::{controller::transactions::TransactionSummary, grid::js_types::*};
 use std::str::FromStr;
 
 pub mod clipboard;
+pub mod files;
```
davidkircos, line 6: rename to `csv`? files seems like it would reference Quadratic files

ddimaria: That's a good question.  I named it b/c it looks like there were more formats coming soon.  How about `import` or `imports`?

Screenshot 2023-09-25 at 5 38 16 PM

davidkircos: `import` sounds good!

quadratic-core/src/controller/files.rs
ddimaria: I set the cursor to the upper left corner of the grid.

quadratic-core/src/controller/import.rs
```diff
@@ -0,0 +1,124 @@
...
+
+                // convert the record into a vector of CellValues
+                record
+                    .iter()
+                    .map(|value| {
+                        Ok(if let Ok(number) = BigDecimal::from_str(value) {
+                            CellValue::Number(number)
+                        } else {
+                            CellValue::Text(value.into())
+                        })
```
davidkircos, line 46: CellValue should have a standard function for parsing into the correct types. We want to do the exact same parsing if a user types in text, copy / pastes, etc. 

This is also done [here](https://github.com/quadratichq/quadratic/blob/08721b4fee5e905aefe5345fbf8c8a0afa8ed819/quadratic-core/src/controller/clipboard.rs#L195), [here](https://github.com/quadratichq/quadratic/blob/56457faa551d3532ed003aa15e73a48322ad29e2/quadratic-core/src/grid/file/v1_4.rs#L161) and [here](https://github.com/quadratichq/quadratic/blob/6eb0a6d2215a33e76a7815cd4bd9c908f51351ca/quadratic-core/src/wasm_bindings/mod.rs#L229)

src/grid/actions/insertData/insertCSV.ts
davidkircos, line 1: I think we can remove this file and call `grid.importCsv` directly from `src/ui/components/FileUploadWrapper.tsx`

quadratic-core/src/controller/import.rs
```diff
@@ -0,0 +1,124 @@
...
+    ///
+    /// Returns a [`TransactionSummary`].
+    pub fn import_csv(
+        &mut self,
+        sheet_id: SheetId,
+        file: &str,
+        file_name: &str,
+        insert_at: Pos,
+        cursor: Option<String>,
+    ) -> Result<TransactionSummary> {
```
davidkircos, line 23: @HactarCE or @Asaaj could you review this Rust code?

quadratic-core/examples/smallpop.csv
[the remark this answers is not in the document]
davidkircos: We should create a library of example CSVs outside of the codebase so we can store large files.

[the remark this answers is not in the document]
ddimaria: I'll remove, I was planning on having a suite of file-driven tests.

src/grid/controller/Grid.ts
```diff
@@ -416,6 +418,22 @@ export class Grid {
...
   //#endregion
+
+  //#region Imports
+
+  async importCsv(sheetId: string, file: File, insertAtCellLocation: Coordinate) {
+    const pos = new Pos(insertAtCellLocation.x, insertAtCellLocation.y);
+    const text = await file.text();
+
+    try {
+      const summary = this.gridController.importCsv(sheetId, text, file.name, pos, sheets.getCursorPosition());
```
davidkircos, line 429: we may want to read the file into a `Uint8Array` instead of a string?

```rust
#[wasm_bindgen]
pub fn process_file(data: &[u8]) {
    // Process the file data here
}
```

```typescript
// This async function reads a File object and passes its content to Rust
async function readFileAndPassToRust(file: File): Promise<void> {
    const uint8Array = await readFileAsArrayBuffer(file);
    
    // Call the Rust function
    rustModule.process_file(uint8Array);
}

function readFileAsArrayBuffer(file: File): Promise<Uint8Array> {
    return new Promise((resolve, reject) => {
        const reader = new FileReader();

        reader.onload = function(event) {
            if (event.target?.result instanceof ArrayBuffer) {
                const uint8Array = new Uint8Array(event.target.result);
                resolve(uint8Array);
            } else {
                reject(new Error("Failed to read file as ArrayBuffer"));
            }
        };

        reader.onerror = function() {
            reject(new Error("Error reading file"));
        };

        reader.readAsArrayBuffer(file);
    });
}
```

ddimaria: Nice find!  IDK that it was available: https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html 🚀

quadratic-core/src/controller/import.rs
ddimaria: I'll do a standard conversion function in another PR.  I'll make a TODO in the code for now.

quadratic-core/src/controller/import.rs
ddimaria: This is done.

src/grid/actions/insertData/insertCSV.ts
ddimaria: This is done.

quadratic-core/examples/smallpop.csv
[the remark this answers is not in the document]
ddimaria: This is done.

src/grid/controller/Grid.ts
davidkircos: It looks like you can reference a Uint8Array as a reference from Rust. Maybe this eliminates a copy or two?

ddimaria: This is done.

quadratic-core/src/controller/import.rs
```diff
@@ -0,0 +1,141 @@
...
+Springfield,OH,United States,64325
+Springfield,OR,United States,56032
+Concord,NH,United States,42605
+"#;
+
+    #[test]
+    fn imports_a_simple_csv() {
+        let mut grid_controller = GridController::new();
+        let sheet_id = grid_controller.grid.sheets()[0].id;
+        let pos = Pos { x: 0, y: 0 };
```
Asaaj, line 101: **suggestion**: Testing a non-zero position may be good as well.

Testing insertion in the negative quadrant or possibly spanning the origin would prove a bit more robustness

quadratic-core/src/test_util.rs
```diff
@@ -0,0 +1,143 @@
...
+    assert!(
+        has_bold == expect_bold,
+        "Cell at ({}, {}) is not bold",
+        x,
+        y
+    );
+}
+
+#[cfg(test)]
+pub fn table(grid_controller: GridController, sheet_id: SheetId, range: &Rect) {
```
Asaaj, line 95: **nitpick**: It would help readability if this said it was printing the table, since it's really just for visualization.

```suggestion
pub fn print_table(grid_controller: GridController, sheet_id: SheetId, range: &Rect) {
```

Asaaj (approved):

quadratic-core/src/lib.rs
```diff
@@ -12,6 +12,7 @@ pub mod grid;
 mod position;
 mod rle;
 mod span;
+mod test_util;
```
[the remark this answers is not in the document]
ddimaria: Good call, there won't be any non-test code in this file.

quadratic-core/src/controller/import.rs
ddimaria: I updated this to non-zero, but didn't expect any different results as it just determines the point of insertion.

quadratic-core/src/test_util.rs
ddimaria: This is done

PR #712, opened 2023-09-26
feat: file thumbnails (and shadcn)

jimniels:
## Feature

- Filter a list of files by keyword
- Sort list of files
- Order list of files
- Grid/list view of files
  - Layout option is visible on larger screens, collapses into the dropdown menu on smaller screens.
- Thumbnails generate within the app when it loads _and_ a cell is edited
  - So, for example, creating a new file does not create a thumbnail. Only once you edit it does the thumbnail get created.
  - Thumbnail creates fixed image of 2048px at 16:9 ratio of a fixed space on the grid

**Also: this PR starts using shadcn components instead of MUI**

### Todos

- [x] Test mobile
- [x] "Menu" button on mobile is transparent which shows items behind it when scrolling.
- [x] Last hoverover on iPad stays for files on dashboard.
- [x] Test large files and ensure it doesn't slow down the grid
- [x] Create thumbnail on example create, first load.
- [x] Move thumbnailDirty flag to rust -- handle calculations in rust and send as part of TransactionSummary
- [x] Placeholder for files without thumbnail
- [x] Basic "Examples" tab (without filtering, thumbnails, etc)
- [x] Fix thumbnail generation @davidfig 
- [x] Image size (and format?) optimization
- [x] shadcn-ui integration: `@` imports for URLs
- [x] BUG: type "(" into the search bar, and it will crash the app. 

## Pre-deploy checklist

- [x] Ensure env variables are set in production server
- [x] Same-origin CORS configuration for fetching previews


quadratic-api/src/routes/files/preview.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// if (
+//   process.env.AWS_S3_REGION !== undefined ||
+//   process.env.AWS_S3_ACCESS_KEY_ID !== undefined ||
+//   process.env.AWS_S3_SECRET_ACCESS_KEY !== undefined ||
+//   process.env.AWS_S3_BUCKET_NAME !== undefined
+// ) {
+//   console.log(process.env);
+//   throw new Error('Missing AWS S3 environment variables');
+// }
```
jimniels, line 15: We should remove all this if we don't need it...

quadratic-api/src/routes/files/fileMiddleware.ts
```diff
@@ -27,6 +27,6 @@ export const fileMiddleware = async (req: Request, res: Response, next: NextFunc
     }
   }
 
-  req.file = file;
+  req.document = file;
```
jimniels, line 30: curious why the rename here? if it's `file` in the API, why rename it to document...

quadratic-api/src/types/Request.ts
```diff
@@ -1,15 +1,21 @@
...
+// file is added by the multer middleware
 export interface Request extends JWTRequest {
   auth?: {
     sub: string;
   };
-  user?: User;
-  file?: File;
+  user?: DBUser;
+  document?: DBFile;
+  file?: UploadFile;
```
jimniels, line 20: I found this a bit confusing when I first saw it.

`file` is thing we have in the API and I think reserving that name for one of the primary objects in the system is better than giving up to a file upload. Maybe we can name the file upload something else and keep `file` reserved so it shares semantics with the API, e.g. `/files/:id`

src/api/types.ts
```diff
@@ -22,6 +22,7 @@ export const ApiSchemas = {
   '/v0/files.GET.response': z.array(
     z.object({
       ...fileMeta,
+      preview: z.string().url().nullable(),
```
jimniels, line 25: I'm thinking maybe we should make this `z.string().url().optional()` as that's what we're doing for everything else in the API that may or may not exist.

So far (that I'm aware of) we don't really have strong semantics in the API for differentiating between something be `undefined` (or optional) vs. being `null` so to avoid ambiguity I think it's better to stick to one which, thus far, has been `.optional()`

quadratic-api/prisma/schema.prisma
```diff
@@ -24,6 +24,7 @@ model File {
     id                 Int            @id @default(autoincrement())
     uuid               String         @unique @default(uuid())
     name               String
+    preview            String? // S3FileKey
```
jimniels, line 27: Wondering about a more descriptive name here, e.g. `preview_url`?

src/dashboard/components/DashboardLayoutRoute.tsx
```diff
@@ -145,31 +100,26 @@ function Navbar({ handleDrawerToggle }: { handleDrawerToggle: Function }) {
...
+              My files
+            </Typography>
+          </SidebarNavLink>
+          <SidebarNavLink to={ROUTES.EXAMPLES} style={sidebarLinkStyles}>
+            <MixIcon className="h-5 w-5" />
+            <Typography variant="body2" color="text.primary">
+              Examples
+            </Typography>
+          </SidebarNavLink>
+        </div>
```
jimniels, line 118: @jimniels remove MUI here

src/router.tsx
```diff
@@ -103,6 +104,7 @@ export const router = createBrowserRouter(
             loader={CloudFilesMigration.loader}
           />
 
+          <Route path="/api/files/:uuid" action={filesAction} loader={() => null} />
```
jimniels, line 107: @jimniels consider moving all this to the `/files` route (and `/files/:uuid`) and their actions

src/dashboard/components/FilesListViewControlsDropdown.tsx
```diff
@@ -0,0 +1,184 @@
...
+    </DropdownMenuRadioGroup>
+  );
+
+  return (
+    <>
+      <div className="hidden sm:flex sm:flex-row sm:items-center sm:gap-1">
+        <DropdownMenu>
+          <DropdownMenuTrigger asChild>
+            <Button variant="ghost" size="sm" className="text-muted-foreground">
+              {orderButtonLabel} <ChevronDownIcon className="ml-1" />
```
jimniels, line 116: @jimniels make these like Figma

src/ui/components/FileProvider.tsx
```diff
@@ -35,6 +35,8 @@ const FileContext = createContext<FileContextType>({} as FileContextType);
 export const FileProvider = ({ children }: { children: React.ReactElement }) => {
   // We can guarantee this is in the URL when it runs, so cast as string
   const { uuid } = useParams() as { uuid: string };
+  //@ts-expect-error
+  window.file_uuid = uuid; // TODO: remove this
```
davidkircos, line 39: remove

quadratic-api/src/routes/files/files.ts
```diff
@@ -137,6 +148,42 @@ files_router.post(
...
+    }
+
+    const permissions = getFilePermissions(req.user, req.document);
+
+    if (permissions !== 'OWNER') {
+      return res.status(403).json({ error: { message: 'Permission denied' } });
+    }
+
+    console.log('updated preview');
+    console.log(req.file);
```
jimniels, line 171: @davidkircos we should remove these

quadratic-api/src/types/Request.ts
jimniels: Change `document` to `quadraticFile`

quadratic-api/src/routes/files/preview.ts
jimniels: We should fix this in #650 or #787 but will remove commented code for now

craco.config.js
```diff
@@ -0,0 +1,15 @@
+const path = require('path');
```
ddimaria, line 1: 🙌

quadratic-api/prisma/migrations/20231115212906_add_file_thumbnail/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
```
ddimaria, line 1: Do we have an opinion about down migrations?

quadratic-api/src/routes/files/files.test.ts
```diff
@@ -531,3 +534,46 @@ describe('DELETE - DELETE /v0/files/:uuid with auth and another users file', ()
       .expect(200); // OK
   });
 });
+
+describe('UPDATE - POST /v0/files/:uuid/thumbnail no auth', () => {
```
ddimaria, line 538: 💯

quadratic-api/src/routes/files/files.ts
```diff
@@ -60,16 +71,21 @@ files_router.get(
...
+
     return res.status(200).json({
       file: {
-        uuid: req.file.uuid,
-        name: req.file.name,
-        created_date: req.file.created_date,
-        updated_date: req.file.updated_date,
-        version: req.file.version,
-        contents: req.file.contents.toString('utf8'),
+        uuid: req.quadraticFile.uuid,
```
ddimaria, line 80: You may be able to spread `req.quadraticFile` to reduce some of this code (sans contents).

quadratic-api/src/routes/files/thumbnail.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
+import multer, { StorageEngine } from 'multer';
+import multerS3 from 'multer-s3';
+import { Request } from '../../types/Request';
+
+// Initialize S3 client
+export const s3Client = new S3Client({
+  region: process.env.AWS_S3_REGION,
+  credentials: {
+    accessKeyId: process.env.AWS_S3_ACCESS_KEY_ID,
```
ddimaria, line 11: Not for now, but it would be nice to place all `process.env` vars in a `config.ts` (or whatever) to avoid having any `process.env` outside of that file.

quadratic-core/src/controller/operations.rs
```diff
@@ -48,6 +45,9 @@ impl GridController {
 
                 let old_values = Array::new_row_major(size, old_values)
                     .expect("error constructing array of old values for SetCells operation");
+                CellSheetsModified::add_region(&mut summary.cell_sheets_modified, sheet, &region);
+                summary.generate_thumbnail =
+                    summary.generate_thumbnail || self.thumbnail_dirty_region(region.clone());
```
ddimaria, line 50: If you make the `region` accept a reference, you should be able to avoid these clones.

quadratic-core/src/controller/operations.rs
```diff
@@ -112,6 +112,11 @@ impl GridController {
                     summary.fill_sheets_modified.push(region.sheet);
                 }
 
+                // todo: this is too slow -- perhaps call this again when we have a better way of setting multiple formats within an array
```
ddimaria, line 115: Worthy of an issue?  We should have some reminder to check back after we remove column/row IDs to see if it's still an issue.

quadratic-core/src/controller/update_code_cell_value.rs
```diff
@@ -75,6 +75,43 @@ pub fn update_code_cell_value(
             }
         }
 
+        // updates summary.thumbnail_dirty flag
+        let sheet = grid_controller.grid.sheet_from_id(cell_ref.sheet);
```
ddimaria, line 79: Is this covered in a test?

quadratic-core/src/grid/mod.rs
```diff
@@ -73,9 +74,15 @@ impl Grid {
...
     }
-    // pub fn dependencies_mut(&mut self) -> &mut HashMap<SheetPos, Vec<SheetRect>> {
-    //     &mut self.dependencies
-    // }
+    pub fn first_sheet_id(&self) -> SheetId {
+        self.sheets
+            .iter()
+            .sorted_by(|a, b| b.order.cmp(&a.order))
+            .last()
+            .unwrap()
```
ddimaria, line 82: unwrap here

quadratic-core/src/position.rs
```diff
@@ -131,6 +131,14 @@ impl Rect {
         self.x_range().contains(&pos.x) && self.y_range().contains(&pos.y)
     }
 
+    /// Returns whether a rectangle intersects with the rectangle.
+    pub fn intersects(self, other: Rect) -> bool {
```
ddimaria, line 135: There is identical code in the `SheetRect` impl in the same file.  Consider making this reusable to avoid the duplication.

ddimaria (approved):
I finished reviewing the Rust code.  I just have some non-blocking nits.

quadratic-api/prisma/migrations/20231115212906_add_file_thumbnail/migration.sql
jimniels: Probably a question for @davidkircos 

AFAIK this is all code generated by prisma so — presumably it's ok?

davidkircos: Prisma doesn't generate down migrations automatically.

davidkircos (approved):

quadratic-core/src/controller/operations.rs
davidfig: Issue #825

quadratic-core/src/controller/update_code_cell_value.rs
davidfig: This particular code is not, but the calls to thumbnail_dirty_rect, etc. are

PR #713, opened 2023-09-29
Next address clippy lints

ddimaria:


ddimaria:
@HactarCE I think you might be the interested party here.  Most are just simple lint fixes, but some required a bit more.

quadratic-core/src/formulas/functions/operators.rs
```diff
@@ -104,7 +104,7 @@ mod tests {
         let g = &mut NoGrid;
 
         assert_eq!(
-            (1 * -6 + -2 - 1 * (-3_i32).pow(2_u32.pow(3))).to_string(),
```
HactarCE, line 107: These `1 *`s are intentional, to match the formula string.

quadratic-core/src/grid/mod.rs
```diff
@@ -122,7 +117,7 @@ impl Grid {
             .iter()
             .any(|old_sheet| old_sheet.name == sheet.name)
         {
-            return Err(());
+            return Err(anyhow!("sheet name already in use"));
```
HactarCE, line 120: Use `bail!("sheet name already in use");`

quadratic-core/src/values/array_size.rs
```diff
@@ -68,6 +68,10 @@ impl ArraySize {
     pub fn len(self) -> usize {
         self.w.get() as usize * self.h.get() as usize
     }
+    /// Determine if the array is empty.
+    pub fn is_empty(self) -> bool {
+        self.w.get() == 0 && self.h.get() == 0
```
HactarCE, line 73: This is impossible. `ArraySize` must be nonempty, so having this method is misleading. I'd be ok with it if the doc comment says that it always returns `false` because arrays are always nonempty.

quadratic-core/src/values/mod.rs
```diff
@@ -130,6 +130,7 @@ impl Spanned<Value> {
 
     /// Iterates over an array, converting values to a particular type. If a
     /// value cannot be converted, it is ignored.
+    #[allow(clippy::should_implement_trait)]
```
HactarCE, line 133: Yeahhhh ... technically we _should_ implement the trait but I didn't because then we'd need an explicit type instead of just `impl Iterator<...>`. This is fine.

HactarCE (changes requested):
Almost everything looks good! Good judgement about where to allow lints.

quadratic-core/src/formulas/functions/operators.rs
ddimaria: I reverted and added an override for this function.

quadratic-core/src/grid/mod.rs
ddimaria: This is done, I keep forgetting this macro, very useful...thanks! (I've mostly used thiserror on other projects)

quadratic-core/src/values/array_size.rs
ddimaria: Clippy complains if you implement a `len()` function that it's idiomatic to have an accompanying `is_empty()` function.  You made both w & h a NonZeroU32, so this is just clippy not being smart enough.  I'll just remove the function and add an override as that seems most appropriate.

ddimaria (commented):
All changes have been implemented.

PR #723, opened 2023-10-10
Add sheets to formulas

HactarCE:
This also removes `GridProxy`. Formulas are now callable by parsing a formula, then constructing a `Ctx` using `Ctx::new(&Grid, Pos)` and calling `Formula::eval(&mut Ctx)`. `Ctx` contains a field `cells_accessed` that is populated after evaluating a formula.

Tests are updated, and even slightly improved since now we can test `SUMIF` etc. on blank cells!


HactarCE:
How to run a formula:
```rs
let grid: Grid;
let pos: Pos;
let pos_with_sheet_id: pos = pos.with_sheet(sheet_id);
let formula_str: String;

let mut ctx = Ctx::new(&grid, pos_with_sheet_id);
let result = parse_formula(&formula_str, pos)?.eval(&mut ctx)?;
let cells_accessed = ctx.cells_accessed;
```

davidfig:
Merged into https://github.com/quadratichq/quadratic/pull/733

PR #730, opened 2023-10-13
Next file import export

ddimaria:
- [x] Remove existing 1.4 and 1.5 versions and replace with a new 1.4
- [x] Have a file schema representation in 1.3 and 1.4 that aren't dependent on *any* other grid Rust code
- [x] Remove conversions past 1.3 in TS
- [x] Delegate conversions 1.3+ to Rust
- [x] Add tests to validate happy path conversions on different scenarios
- [x] Validate successful conversions of existing example file 
- [x] Replace existing example files  with 1.4 versions


quadratic-core/Cargo.toml
```diff
@@ -61,6 +61,7 @@ lexicon_fractional_index = "0.0.3"
 htmlescape = "0.3.1"
 bigdecimal = { version = "0.4.1", features = ["serde"] }
 csv = "1.2.2"
+bincode = "1.3.3"
```
davidkircos, line 64: `bincode` is unused?

ddimaria: Yes, I'm removing as we're not implementing binary encoding at a later date.

src/api/apiClient.ts
```diff
@@ -3,26 +3,32 @@ import { v4 as uuid } from 'uuid';
...
 import mixpanel from 'mixpanel-browser';
 import { downloadFileInBrowser } from '../helpers/downloadFileInBrowser';
-import { GridFile, GridFileSchema } from '../schemas';
 import { generateKeyBetween } from '../utils/fractionalIndexing';
 import { fetchFromApi } from './fetchFromApi';
 import { ApiSchemas, ApiTypes } from './types';
 
-const DEFAULT_FILE: GridFile = {
+const DEFAULT_FILE: any = {
+  version: '1.6',
```
davidkircos, line 11: `1.6`?

src/schemas/validateAndUpgradeLegacyGridFile.ts
```diff
@@ -15,7 +15,7 @@ const FILES = [
 ];
 
 // all versions validated by Rust
-const rustFileVersions = ['1.5'];
+const rustFileVersions = ['1.6'];
```
davidkircos, line 18: `1.6`?

ddimaria: These were changed back, but I mistakenly only added files in quadratic-core in my previous commit.  I just committed from the top-level dir and pushed these changes up.

PR #738, opened 2023-10-24
Next delete

davidfig:
- [x] clipboard undo/redo w/code cells
- [x] delete code cells


quadratic-core/src/controller/clipboard.rs
```diff
@@ -384,6 +424,74 @@ mod test {
         );
     }
 
+    #[test]
```
ddimaria, line 427: 💯

quadratic-core/src/grid/sheet.rs
```diff
@@ -194,6 +194,15 @@ impl Sheet {
         }
     }
 
+    pub fn get_cell_value_only(&self, pos: Pos) -> Option<CellValue> {
```
ddimaria, line 197: You could DRY this out by having ` get_cell_value()` call this and invoking `self.get_code_cell_value(pos)` in the `None` case.

ddimaria:
Code looks good. I have one non-blocking comment.

ddimaria (approved):

PR #741, opened 2023-10-24
Next sheet palette permission

davidfig:
Command palette does not have Sheet commands when below editor


src/ui/menus/CommandPalette/ListItems/Sheets.tsx
```diff
@@ -43,7 +47,7 @@ const ListItems = () => {
     sheets.forEach((sheet) => {
       items.push({
         label: `Sheet: Switch to “${sheet.name}”`,
-        isAvailable: () => sheets.current !== sheet.id,
+        isAvailable: (permission) => isEditorOrAbove(permission) && sheets.current !== sheet.id,
```
jimniels, line 50: I don't think you need this here. The ability to switch between sheets should be available for anyone with access to the document (you're not editing the document to switch between sheets). That's all part of viewing the sheet.

jimniels (changes requested):

PR #745, opened 2023-10-25
Next renderer preloader

davidfig:
- [x] separate out preloader from CellsSheet.ts
- [x] add has_render_cells() for use when preloading CellsSheet

### DIFF
--- a/src/gridGL/cells/CellsSheet.ts
+++ b/src/gridGL/cells/CellsSheet.ts
@@ -9,28 +9,25 @@ import { CellsArray } from './CellsArray';
 import { CellsBorders } from './CellsBorders';
 import { CellsFills } from './CellsFills';
 import { CellsMarkers } from './CellsMarker';
+import { CellsSheetPreloader } from './CellsSheetPreloader';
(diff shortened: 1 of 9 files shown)


src/gridGL/cells/CellsSheet.ts
```diff
@@ -9,28 +9,25 @@ import { CellsArray } from './CellsArray';
...
-  // used to draw debug boxes for cellsTextHash
-  private cellsTextDebug: Graphics;
-
-  private cellsTextHashContainer: Container<CellsTextHash>;
-
-  private cellsArray: CellsArray;
+  // friends of CellsSheetPreloader.ts
+  cellsFills: CellsFills;
+  cellsArray: CellsArray;
+  cellsBorders: CellsBorders;
```
davidkircos, line 20: Instead of making these public maybe make a function that calls create on all 3 on CellsSheet?

davidfig: Good catch. I can actually just call the create function in this object instead of preloader.

PR #746, opened 2023-10-26
fix formula highlighting

davidfig:
- [x] formula highlighting only shows on correct sheet
- [x] code cell highlight only shows on correct sheet
- [x] cell highlight fill when cursor is over formula reference
- [ ] fix parse_formula so it properly returns sheet name (Andrew)


ddimaria (approved):
Code looks good 🙌   My only suggestion is to remove commented out code, or provide some info in the code as to why we're keeping it around.

PR #748, opened 2023-10-26
Gird Performance to Sentry

davidkircos:
Sends 1/10 perf events to Sentry. It allows for great visibility into how long things take for the user.

CleanShot 2023-10-30 at 12 05 09@2x

- [x] verify metrics make it to sentry
- [x] verify sentry still works generally


quadratic-core/src/controller/cells.rs
```diff
@@ -88,7 +88,12 @@ impl GridController {
         }
         // todo: include other types here
         else {
-            let values = Array::from(CellValue::Text(value.into()));
+            let values = if value.is_empty() {
+                Array::from(CellValue::Blank)
+            } else {
+                Array::from(CellValue::Text(value.into()))
+            };
```
davidkircos, line 95: @ddimaria should we do this here? Prevents empty strings from being created into cells.

ddimaria: It might be best to have this in convert.rs, as we have string conversions in place that it might be good to enforce centrally: 
```rust
impl From<String> for CellValue {
    fn from(value: String) -> Self {
        CellValue::Text(value)
    }
}
impl From<&str> for CellValue {
    fn from(value: &str) -> Self {
        CellValue::Text(value.to_string())
    }
}
```

davidkircos: Seems like the internal function String -> CellValue should maybe always create a CellValue::Text type? But i am not sure how it's used.

src/index.js
```diff
@@ -13,13 +12,16 @@ import { QuadraticLoading } from './ui/loading/QuadraticLoading';
...
   Sentry.init({
     dsn: process.env.REACT_APP_SENTRY_DSN,
-    integrations: [new BrowserTracing()],
+    environment: 'development',
+    integrations: [new Sentry.BrowserTracing()],
 
     // We recommend adjusting this value in production, or using tracesSampler
     // for finer control
-    tracesSampleRate: 1.0,
+    tracesSampleRate: 0.1,
```
davidkircos, line 20: only sends 1/10 events to Sentry

quadratic-core/src/controller/cells.rs
ddimaria: I was thinking something like:

```rust
impl From<String> for CellValue {
    fn from(value: String) -> Self {
        if value.is_empty() {
            CellValue::Blank
        } else {
            CellValue::Text(value)
        }
    }
}
```

PR #754, opened 2023-10-28
Next perf

davidfig:
- [x] refactor rust's getRenderCells to slightly improve performance
- [x] add scheduling (distance) algorithm to cellSheets so hashes are updated based on distance to middle of viewport 
- [ ] figure out high memory usage


quadratic-core/src/grid/column.rs
```diff
@@ -299,6 +299,14 @@ impl<B: BlockContent> ColumnData<B> {
                 (start..end).filter_map(|y| Some((y, block.get(y)?)))
             })
     }
+    /// Iterates over a range, skipping cells with no data.
+    pub fn values_in_range_all(
+        &self,
+        y_range: Range<i64>,
+    ) -> impl '_ + Iterator<Item = (i64, Option<B::Item>)> {
+        self.blocks_covering_range(y_range.clone())
```
ddimaria, line 307: I think these are all just reads.  You may be able to remove these clones in various ways.  One way could be to send over a read-only range to the functions (e.g. `&Range<i64>`).

quadratic-core/src/grid/sheet/rendering.rs
```diff
@@ -17,107 +17,113 @@ impl Sheet {
...
+        let mut render_cells = vec![];
+        rect.x_range().for_each(|x| {
+            let column = self.get_column(x);
+            rect.y_range().for_each(|y| {
+                if let Some(column) = column {
+                    // first check if there a column.values
+                    if let Some(value) = column.values.get(y) {
+                        let (numeric_format, numeric_decimal) = if value.type_name() == "number" {
+                            let numeric_format = column.numeric_format.get(y);
+                            let is_percentage = if let Some(numeric_format) = numeric_format.clone()
```
ddimaria, line 35: nit: I've been using `to_owned()` to denote an intentional clone that is lightweight (i.e. not expensive).

quadratic-core/src/controller/transaction_summary.rs
```diff
@@ -7,8 +7,8 @@ use crate::{
     Pos,
 };
 
-const CELL_SHEET_WIDTH: u32 = 20;
-const CELL_SHEET_HEIGHT: u32 = 40;
+const CELL_SHEET_WIDTH: u32 = 10;
+const CELL_SHEET_HEIGHT: u32 = 20;
```
ddimaria, line 11: I think we're also using these values in transactions.rs.  Maybe we should make these constants public for the function to use: 
```rust
mpl From<Pos> for CellHash {
    fn from(pos: Pos) -> Self {
        let hash_width = 20_f64;
        let hash_height = 40_f64;
        let cell_hash_x = (pos.x as f64 / hash_width).floor() as i64;
        let cell_hash_y = (pos.y as f64 / hash_height).floor() as i64;
        let cell_hash = format!("{},{}", cell_hash_x, cell_hash_y);

        CellHash(cell_hash)
    }
}
```

ddimaria (approved):
Looks good.  I have some nits and one important comment on cell sheet hash values in transactions.rs.

src/gridGL/cells/CellsSheet.ts
```diff
@@ -278,13 +279,29 @@ export class CellsSheet extends Container {
     return true;
   }
 
+  private dirtyCellsTextHashes(): CellsTextHash[] {
```
davidkircos, line 282: nits
rename dirtyCellTextHashsByDisance ?
does viewport have a topLeft? it's a bit awkward to watch things render up, feels better top to bottom

quadratic-core/src/controller/transaction_summary.rs
davidfig: I think these already exist somewhere. Let me point to that.

src/gridGL/cells/CellsSheet.ts
davidfig: yes. but it'll render stuff outside the viewport before inside (ie, if you use topLeft, then the distance to something above the viewport is smaller than in the middle of the visible viewport)

davidfig: what i could do is sort the resulting list by y so it's top -> down

davidkircos: Ideally, we do what's on the screen first, top to bottom, then everything else by what is closest to the viewport

davidfig: that's how it works now

PR #761, opened 2023-11-01
Next Transaction in Progress / debugShowRustTime

davidfig:
- [x] don't panic when attempting to edit while a transaction is in progress
- [x] add a debugShowRustTime for the proxy rust calls 
- [x] create busy toast for transaction in progress


quadratic-core/src/controller/transaction_summary.rs
```diff
@@ -68,9 +68,19 @@ pub struct TransactionSummary {
 
     // should the grid trigger a save
     pub save: bool,
+
+    // let TS know that the grid is already busy
+    pub transaction_busy: bool,
```
ddimaria, line 73: nit on name, consider transaction_in_progress to match the backend.

ddimaria (approved):
Just one nit.

quadratic-core/src/controller/transaction_summary.rs
davidfig: This is actually different. This only triggers when a transaction is already in progress and the user attempts to start another transaction. The flag does not go on when a transaction is in progress (although that's something I want to do in the future)

Issue #401, opened 2023-04-04
feat: Quadratic Hosted CORS proxy & improved Python API requests overall


davidkircos:
Create a cors proxy and Python Lib to access it that mimics 'requests'. 

This will make it much easier to fetch data

- [ ] Secrets added at proxy level?

lucasew:
Well, CORS has a reason to exist.

That would allow some abuses, and bandwidth for arbitrary data is not so cacheable and cheap to host.

Here is an alternative:
https://github.com/Rob--W/cors-anywhere

davidkircos:
Nice, thanks for sharing cors-anywhere.

We will add user auth and rate limiting on the CORs proxy to combat abuse.

davidkircos:
https://github.com/koenvo/pyodide-http

We can also document this in the meantime and maybe install it by default

ddimaria:
We now load pyodide-http by default.  Is there more work to do on this @luke-quadratic @davidkircos?

PR #766, opened 2023-11-04
feat: Data Connectors (SQL)

davidkircos:
## Goals

- Add a SQL connection
- Support Postgres, MySQL
- Connection Code Editor
- Run code to get data in sheet
- Basic schema viewer

Closes: https://github.com/quadratichq/quadratic/issues/401

## TODOs

- [x] Add mix panel events where it makes sense for connection-related interactions @jimniels @davidkircos @luke-quadratic 
- [ ] Test Connection endpoint has no timeout (should be like 15s). Testing a connection where the connection doesn't resolve just hangs forever.

### Deployment

Remaining TODOs:

- [x] Infra ready for production @davidkircos 
- [ ] Create/publish docs for connections @luke-quadratic 

Release process:

- [x] Ensure env variables are configured properly
- [x] Set IAM Permissions in AWS for `infra`
- [ ] Merge PR
- [ ] Run migration script 1 & 2
`npm run script:migrate-1 --workspace=quadratic-api`
`npm run script:migrate-1 --workspace=quadratic-api -- --commit`
`npm run script:migrate-2 --workspace=quadratic-api`
`npm run script:migrate-2 --workspace=quadratic-api -- --commit`

Note: the app will break for anyone who doesn't refresh.

### Followups

- [ ] `TODO: (team-schema)` comments are things to fix once we merge & run the migration script (e.g. `ownerTeamId` can now be a required field in the schema)
- [x] "Share" dialog for private file vs. team file (#1562)
- [x] Create, duplicate actions from within the app (#1561)
- [ ] Edit button in the command palette to edit an action directly?
- [ ] Store which file references which connections. Maybe do this in file service on checkpoints?
- [ ] Enhancement: split up the team dashboard API calls on a per-route basis (less data over the wire)
- [ ] Better experience around handling the lifecycle of a connection, e.g. somebody adds a connection to a file, runs a query, then later deletes that connection. So what's a good UX for working with a connection in a sheet that's been deleted?
- [ ] Fix syntax highlighting for formulas (should be the same in the inline cell editor, the code editor, and the ai assistant code block)
- [ ] Option with SQL to not include the header when returning data to the sheet. For example when returning a count(*) you might just want the value and not also the header `count`

## UX To Test

General:

- [ ] First-timers: when you come into the app for the first time:
  - [ ]  **On your own**: a team is created for you
  - [ ] **From a team invite**: you're assigned to the team and a team _is not_ created for you
- [ ] Permissions (both in dashboard & app)
  - [ ] View-only team members can create or edit their own drafts, but have no edit access to public team files. They can view team members. They cannot view team connections.
  - [ ] Editor team members can create & edit files, connections, members, etc.
- [ ] Sharing works as expected, e.g.
  - [ ] Can share a public team file with someone not on a team
  - [ ] Can share a private team file with someone on and not on a team

Dashboard: everything that's a button in the UI , but a few things as an example:

- [ ] Can move a file from "private" team files to "public" team files
- [ ] Creating/duplicating files does so in the right "spot" (e.g. as a private file, team file, etc.)
- [ ] Switching between teams works as expected
- [ ] Editing team name works as expected
- [ ] Leave a team, see that you're redirected to another team (or a new one is created for you)
- [ ] Duplicating an example file works (and ends up in private files)
- [ ] Teams can't (yet) be deleted or archived.

In-app:

- [ ] Duplicating a file duplicates to your private files in the team
- [ ] Creating a file creates a private file in the team you're in
- [ ] In-app label shows correctly whether you're looking at a team file or a private draft (by you or someone else)

## Notable breaking changes

- You can no longer duplicate files across teams (#1423)
- Cloud files migration no longer works (#1423)
- API breaking changes so the old client will break (#1423)


src/gridGL/UI/Cursor.ts
```diff
@@ -91,7 +91,7 @@ export class Cursor extends Graphics {
     this.lineTo(x, y + height);
     this.lineTo(x, y);
 
-    if (showInput && cellEdit) {
+    if (showInput && cellEdit && sheets.sheet.id === editorInteractionState.selectedCellSheet) {
```
davidkircos, line 94: incorrectly merged?

quadratic-api/prisma/schema.prisma
```diff
@@ -119,6 +120,52 @@ enum TeamRole {
...
+    uuid             String     @unique @default(uuid()) // used in URLs
+    connectionId     Int
+    connection       Connection @relation(fields: [connectionId], references: [id])
+    success          Boolean
+    std_out          String?
+    std_err          String?
+    result_meta_data Json?
+    time_start       DateTime   @default(now())
+    time_end         DateTime?
+}
```
davidkircos, line 167: Not stored on API

quadratic-api/src/routes/v0/internal.connections.$uuid.GET.ts
```diff
@@ -0,0 +1,34 @@
+import { Response } from 'express';
+import z from 'zod';
+import { validateM2MAuth } from '../../internal/validateM2MAuth';
+import { getConnection } from '../../middleware/getConnection';
+import { validateAccessToken } from '../../middleware/validateAccessToken';
+import { parseRequest } from '../../middleware/validateRequestSchema';
+import { RequestWithUser } from '../../types/Request';
+
+export default [validateM2MAuth, validateAccessToken, handler];
```
davidkircos, line 9: Validating both M2MAuth and the users accessToken seems like the right path long term, but feel free to comment out as we are testing

quadratic-client/src/app/quadratic-core-types/index.d.ts
```diff
@@ -1,7 +1,8 @@
 // This file is automatically generated by quadratic-core/src/bin/export_types.rs
 // Do not modify it manually.
 
-export type CodeCellLanguage = "Python" | "Formula";
+export type CodeCellLanguage = "Python" | "Formula" | { "Connection": { kind: ConnectionKind, id: string, } };
```
jimniels, line 4: @ddimaria minor, but can we rename `id` to `uuid`? It is a UUID string and that's how it's represented in the API, so it makes the client code a bit more uniform when we're working with this, e.g.

```
connectionsFromApi.find(connection => connection.uuid === editorInteractionState.mode?.Connection?.uuid)
```

jimniels: Actually, we refer to the `kind` as `type` in the API as well, might be worth aligning on this language as wel...

ddimaria: Rustland used kind instead of type b/c type is a reserved word.  I can look into rewriting during serialization.

I have a small opinion about not changing id to uuid (seems like a database antipattern to have two unique identifiers...a UUID type in Postgres is idiomatic for the PK, I think we use the id as an index or ordering mechanism?).

jimniels:
Noting this for future reference (if ever needed).

---

There are a few different ways we can load the connections in the sheet.

### One

Let the root of the file route `/file/:id` return the connection data (along with all the other file data) and re-validate.

PRO:
- Don't have to fetch the connection data when the "manage connections" dialog opens
- Server figures out connections based on permissions [^1]
- Revalidation happens automatically and just for the data we want

CON:
- Have to run conditional revalidation on root, pull out what we do and don't want
- Will revalidate when share  modal opens, for example

```
// /file/:uuid
loader() {
  if (!hasMounted) {
    await fetch(/file/uuid)
  }
  await fetch(connections)
  return { connections, staticIps, file }
}
```


### Two

Use a pathless route under `/file/:uuid`, but use a new endpoint in the API to fetch the data `/files/:uuid/connections`

PRO:
- All of no. 1 

CON:
- A new endpoint in the API `/files/:uuid/connections` because loaders run in parallel and all we know is the file UUID
- Would revalidate on everything, like share modal

```
// /file/:uuid
loader() {
  fetch(fileData)
}

// /file/:uui/[pathless-route]
loader() {
  fetch(file/uuid/connections)
  return { connections, staticIps }
}
```

### Three

Fetch at runtime. This won't block the app from loading _before_ we have the list of connections. This is probably ok since it's unlikely somebody opens the sheet and opens the cell type menu immediately.

This allows us to fetch the data in one place and re-use it elsewhere. It also allows us to use info we get from the file route (like what team the file is in) and then make the call (rather than running different API calls, as in no. 2)

```
// <ConnectionsMenu>
fetcher({ key: 'connections' })

// <CellTypeMenu>
// Only fetch if teamPermissions = TEAM_EDIT
useFecther({ key: 'connections' })
```

Going with option 3 because it's the best set of trade-offs without significantly re-factoring the routes. Should have little to no end-user implication as they shouldn't be able to really load the app fast enough to see connections missing.



[1]: If you fetch the data at runtime, you have to figure out if they should be able to fetch them based on permissions

jimniels (approved):
I have reviewed the client/api files in this code (js/ts/tsx) and LGTM 👍 

image

PR #772, opened 2023-11-07
Next commas

davidfig:
- [x] add Toggle Commas in number top menu
- [x] add Toggle Commas in palette
- [x] add comma formatting for numbers:
   * if commas = true, show commas
   * if currency && commas = undefined, show commas
   * otherwise no commas  

Note: I did not add to the floating tool bar, as it's getting too big


public/examples/airports_distance.grid
```diff
@@ -5599,6 +5599,7 @@
             "wrap": {},
             "numeric_format": {},
             "numeric_decimals": {},
+            "numeric_commas": {},
```
ddimaria, line 5602: Are there defaults if we leave this out?

quadratic-core/src/controller/formatting.rs
```diff
@@ -119,6 +123,30 @@ impl GridController {
...
+        cursor: Option<String>,
+    ) -> TransactionSummary {
+        let sheet = self.sheet(sheet_id);
+        let commas = if let Some(commas) = sheet.get_formatting_value::<NumericCommas>(source) {
+            !commas
+        } else {
+            true
+        };
+        let region = self.region(sheet_id, rect);
+        let ops = vec![Operation::SetCellFormats {
```
ddimaria, line 140: Not related to this PR, but why do we need `region` in both places here?  That is, can't we derive the `len()` from the region in the first invocation?

quadratic-core/src/values/cellvalue.rs
```diff
@@ -88,16 +88,55 @@ impl CellValue {
         }
     }
 
+    fn add_commas(s: &str) -> String {
```
ddimaria, line 91: This works, but you could try [chunks](https://doc.rust-lang.org/std/slice/struct.Chunks.html) for brevity and the extra allocations.

quadratic-core/src/values/cellvalue.rs
```diff
@@ -88,16 +88,55 @@ impl CellValue {
...
+    }
+
+    /// converts a BigDecimal to a String w/commas
+    fn with_commas(bd: BigDecimal) -> String {
+        let mut s = bd.to_string();
+        let negative = s.starts_with('-');
+        s = s.trim_start_matches('-').to_string();
+        let n = if s.contains('.') {
+            let mut parts = s.split('.');
+            let left = parts.next().unwrap();
```
ddimaria, line 113: These unwraps are unsafe.  If you do the split and collect before the `contains()` conditional, you replace the conditional with one that checks for a size of 2 of the resulting vec and avoid the unwraps.

quadratic-core/src/controller/formatting.rs
davidfig: Do you mean that we could avoid one of the clones by calculating length before using region?

quadratic-core/src/values/cellvalue.rs
```diff
@@ -88,16 +88,55 @@ impl CellValue {
...
         numeric_decimals: Option<i16>,
+        numeric_commas: Option<bool>,
     ) -> String {
         match self {
             CellValue::Blank => String::new(),
             CellValue::Text(s) => s.to_string(),
             CellValue::Number(n) => {
                 let numeric_format = numeric_format.unwrap_or_default();
+                let use_commas = numeric_commas.is_some_and(|c| c)
+                    || (numeric_commas.is_none()
```
ddimaria, line 138: Is the `is_none()` check necessary?
```rust
let use_commas = numeric_comma.is_some() || numeric_format.kind == NumericFormatKind::Currency;
```

quadratic-core/src/values/cellvalue.rs
davidfig: Technically these unwraps are safe since BigDecimal guarantees a valid number string "a.b" if a '.' is included. But there's probably a better way of doing this. I relied on AI for both of these functions.

ddimaria (approved):
Looks good!  I left mostly non-blocking comments (removing the unwraps could be considered blocking though).

quadratic-core/src/controller/formatting.rs
ddimaria: That's an option, but I was just saying if we pass in the region and the len if the RLE is always the regions length, then we don't really need that field collected (and can remove all of the clones everywhere).  We can discuss this more if we want to refactor the code at some point.

quadratic-core/src/values/cellvalue.rs
ddimaria: Agreed they are technically safe, but it's generally a good idea to remove them in non-test code.  I guess you can add code comments about them being safe.

quadratic-core/src/values/cellvalue.rs
davidfig: It's necessary. Otherwise you couldn't remove commas from currency

PR #777, opened 2023-11-08
Next python running indicator

davidfig:


ddimaria (approved):
Code looks good and it behaves as expected.

PR #782, opened 2023-11-08
Add sheet info to `parse_formula()`

HactarCE:
The params & return value stay the same. `parse_formula()` just does parsing, no semantics, so it only knows sheet names given by the user, not sheet IDs or the current sheet. E.g., `SUM(A1:A10)` would have `sheet: null` for the cell reference `A1:A10` but `SUM(MySheet!A1:A10)` would have `sheet: "MySheet"` for the cell reference `MySheet!A1:A10`.

To review:
- Look over the new tests and see if anything's missing.


ddimaria:
I did a code pass and all looks great 💯

davidfig (approved):

davidfig:
Looks great. I pushed the changes to make it work with highlightedCells and works perfectly. Thanks!

PR #787, opened 2023-11-11
chore: replace CRA with vite

jimniels:
### Todos

- [x] Move all client in to `quadratic-client`
- [x] Get python working in the app
- [x] Get any rust stuff working
- [x] Get tests passing
- [x] Get previews working on vercel
- [x] Get front-end compiling with typescript
- [x] Get `npm run build` working
- [x] Get linting working as part of typescript

We're going to hold off on switching to vitest until later. For now we'll keep working with jest.

### Pre-deploy checklist

- [x] Rename all env variables from `REACT_APP_` to `VITE_`
- [x] Confirm build command works on Amplify


.env_example
```diff
@@ -1,10 +1,10 @@
...
+VITE_GOOGLE_ANALYTICS_GTAG=G-0000000000
+VITE_AMPLITUDE_ANALYTICS_API_KEY=
+VITE_MIXPANEL_ANALYTICS_KEY=
+VITE_SENTRY_DSN=https://xxxxxxxxxxxxxxxxxx@xxxxxxxxxxxx.ingest.sentry.io/xxxxxxxxxxxx
+VITE_DEBUG=0 // use =1 to enable debug flags
+VITE_AUTH0_DOMAIN=auth-dev.quadratic.to
+VITE_AUTH0_CLIENT_ID=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+VITE_AUTH0_AUDIENCE=https://localhost:8000
+VITE_AUTH0_ISSUER=https://auth-dev.quadratic.to/
+VITE_QUADRATIC_API_URL=http://localhost:8000
```
jimniels, line 10: Once this PR merges, update [the Notion doc](https://www.notion.so/Env-Variables-78b1a1da19d0421993abe8c449e51496?pvs=4) where we store these for the team

quadratic-core/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Andrew Farkas <andrew.farkas@quadratic.to>"]
 edition = "2021"
 description = "Infinite data grid with Python, JavaScript, and SQL built-in"
 repository = "https://github.com/quadratichq/quadratic"
+license-file = "LICENSE"
```
jimniels, line 8: Not sure why this is part of this PR?

src/gridGL/cells/cellsLabel/LabelMeshEntry.ts
```diff
@@ -12,8 +12,10 @@ export class LabelMeshEntry extends Mesh {
   vertexCount = 0;
   uvsCount = 0;
 
+  // @ts-expect-error
   indices!: Uint16Array;
   vertices!: Float32Array;
+  // @ts-expect-error
```
jimniels, line 18: This was one thing I ran into when setting up vite.

Idk why Typescript is yelling at us for these in this branch and not main? But I had to add these directives to expect an error otherwise typescript throws an error like:

```
Property 'indices' will overwrite the base property in 'Mesh<MeshMaterial>'. If this is intentional, add an initializer. Otherwise, add a 'declare' modifier or remove the redundant declaration.ts(2612)
```

quadratic-core/Cargo.toml
jimniels, line 8: Not sure why this is part of this PR?

src/index.tsx
```diff
@@ -37,8 +36,3 @@ root.render(
     />
   </React.StrictMode>
 );
-
-// If you want to start measuring performance in your app, pass a function
-// to log results (for example: reportWebVitals(console.log))
-// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
-reportWebVitals();
```
jimniels, line 44: This was [a CRA-speficic thing](https://create-react-app.dev/docs/measuring-performance/) that, AFAIK, we're not using or taking advantage of. So I removed it.

quadratic-core/Cargo.toml
ddimaria: Git says that @davidfig made this addition.  I think this can be removed as it expects a file name and that's not a file name that exists in the repo (it exists at the root of the repo, not in this dir).

PR #788, opened 2023-11-13
Next last 2

davidfig:
- [x] deleting cells properly triggers computation cycle
- [x] formula highlight works across sheets with the correct color


ddimaria (approved):
Code looks good!

PR #795, opened 2023-11-14
sheet name properly renames on blur

### LINKED ISSUE #790: Rename sheet doesn't confirm when clicking away
When renaming sheet if you click away without confirming via pressing enter the sheet rename should still be confirmed. 

Not really a bug, moreso incorrect UI behavior (leaving in bug category for now).

davidfig:
Fixes #790


ddimaria (approved):
Code looks good.

PR #796, opened 2023-11-14
input arrow left/right moves cells

davidfig:
- Arrows while editing a cell will move the cell left and right if at the start or end
- Shift Tab submits and goes to cell 1 to the left, tab goes 1 to the right


ddimaria (approved):
Code looks good on this one.  This still needs a functionality review.

jimniels (commented):
LGTM!

PR #800, opened 2023-11-14
console indicates when it has std out or std err

davidkircos:


src/ui/menus/CodeEditor/Console.tsx
```diff
@@ -49,6 +50,8 @@ export function Console({ consoleInput, editorMode, editorContent, evaluationRes
             label="Console"
             id="console-tab-0"
             aria-controls="console-tabpanel-0"
+            icon={hasOutput ? <Circle sx={{ fontSize: 10 }}></Circle> : undefined}
```
jimniels, line 53: I'd say bump this down in size just a tiny bit more, like `fontSize: 8`

jimniels (approved):
I'd say bump the size down on the indicator a tiny bit more, but otherwise I think this is a change in the right direction!

PR #803, opened 2023-11-14
0 properly outputs as string for python

### LINKED ISSUE #802: bug: 0 doesn't show up in array output from Python
CleanShot 2023-11-14 at 12 27 40@2x

davidfig:
Fixes #802

### DIFF
--- a/src/web-workers/pythonWebWorker/python.ts
+++ b/src/web-workers/pythonWebWorker/python.ts
@@ -2,6 +2,10 @@ import { grid, pointsToRect } from '../../grid/controller/Grid';
 import { JsCodeResult } from '../../quadratic-core/quadratic_core';
 import { PythonMessage, PythonReturnType } from './pythonTypes';
 
+const stringOrNumber = (input: string | number): string => {
+  return typeof input === 'string' ? input : typeof input === 'undefined' ? '' : input.toString();
+};
+
 class PythonWebWorker {
   private worker?: Worker;
   private loaded = false;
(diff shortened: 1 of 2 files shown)


ddimaria (approved):
Code looks good!

src/web-workers/pythonWebWorker/python.ts
```diff
@@ -2,6 +2,10 @@ import { grid, pointsToRect } from '../../grid/controller/Grid';
 import { JsCodeResult } from '../../quadratic-core/quadratic_core';
 import { PythonMessage, PythonReturnType } from './pythonTypes';
 
+const stringOrNumber = (input: string | number): string => {
+  return typeof input === 'string' ? input : typeof input === 'undefined' ? '' : input.toString();
```
davidkircos, line 6: Why do we have to check if it's undefined if the typescript says it's always either `(input: string | number)`?

davidkircos (changes requested):

src/web-workers/pythonWebWorker/python.ts
davidkircos: Could you use `if` instead of nested ternary? difficult to read

src/web-workers/pythonWebWorker/python.ts
```diff
@@ -18,11 +22,11 @@ class PythonWebWorker {
         if (pythonResult.array_output) {
           if (!Array.isArray(pythonResult.array_output[0])) {
             pythonResult.array_output = pythonResult.array_output.flatMap((entry: string | number) => [
-              [entry ? entry.toString() : ''],
+              [stringOrNumber(entry)],
             ]);
           } else {
             pythonResult.array_output = pythonResult.array_output.map((entry: (string | number)[]) =>
-              entry.map((entry: String | number) => (entry ? entry.toString() : ''))
```
davidkircos, line 25: if it can't be undefined, we can always call entry.toString()?

src/web-workers/pythonWebWorker/python.ts
davidfig: The typescript is wrong. It can be undefined. I'll add it as an option, and remove the ternary (although I do think it's cool).

ddimaria (approved):

davidkircos (approved):

PR #809, opened 2023-11-15
FloatingContextMenu Icon Consistency

davidkircos:
Before

CleanShot 2023-11-14 at 21 46 34@2x

After

CleanShot 2023-11-14 at 21 47 08@2x


jimniels (approved):
Looks good to me!

There is one linting issue you could fix before merging.

ddimaria (approved):
Code looks good.  UI works as expected.

PR #810, opened 2023-11-15
Readability Theme Updates

davidkircos:
- make grid lines more readable
- make headers more readable


jimniels (approved):
I tweaked this to use less colors (and color variants). 

I pulled the hue color from our theme and adjusted it for the grid so there's a unifying hue through the app (even more so when we get the app using shadcn styles more coherently).

Left is before my commits, right is after. Subtle but more unified with the overarching theme of the app.

image

PR #811, opened 2023-11-15
Double click headers mouse move

davidfig:
- [x] fixes bug where you can click the bottom-right corner of a the header, move the mouse, and still complete the double click to resize the column
- [x] fixed a nasty bug where automatically resizing the column would result in bad rendering of columns. also fixed the same bug with undoing that change 
- [x] changed the minimum column size from 20px (which is the minimum a user can shrink the heading) to 200px (which is the default column width)
- [x] read-only can no longer resize or auto resize headers

### DIFF
--- a/quadratic-core/src/controller/sheet_offsets.rs
+++ b/quadratic-core/src/controller/sheet_offsets.rs
@@ -16,8 +16,8 @@ impl GridController {
         transient_resize: Option<TransientResize>,
         cursor: Option<String>,
     ) -> TransactionSummary {
-        let sheet = self.grid.sheet_mut_from_id(sheet_id);
         if let Some(transient_resize) = transient_resize {
+            let sheet = self.grid.sheet_mut_from_id(sheet_id);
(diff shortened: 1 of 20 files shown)


quadratic-core/src/controller/sheet_offsets.rs
```diff
@@ -39,6 +39,29 @@ impl GridController {
             TransactionSummary::default()
         }
     }
+
+    pub fn commit_single_resize(
```
ddimaria, line 43: Can you add a TODO here to test this function?

ddimaria (approved):
Code looks good.

quadratic-core/src/controller/sheet_offsets.rs
davidkircos: Is now a good time to just add some tests?

davidfig: added some tests

PR #819, opened 2023-11-16
Fix blank grid loading on mobile

### LINKED ISSUE #815: bug: When loading files on mobile they appear blank until moving the viewport
Tested on `main` / iPhone / Chrome

davidkircos:
Fixes #815

When loading documents on mobile, they previously show up blank until moving the viewport.


src/ui/menus/TopBar/SubMenus/QuadraticMenu.tsx
```diff
@@ -51,6 +52,7 @@ export const QuadraticMenu = () => {
   useEffect(() => {
     if (isMobile) {
```
davidkircos, line 53: For mobile, Headers are set to not visible by default. For some reason this resulted in a partial render of the grid by default for all files loaded on mobile.

davidfig (approved):

PR #820, opened 2023-11-16
Fix blank string cells from copy/paste/import csv

### LINKED ISSUE #818: feat: Strip Whitespace on Paste/CSV Import
- leading and trailing whitespace when importing data
- if this results in an empty string for a cell, keep the cell blank

davidkircos:
fixes #818


quadratic-core/src/controller/cells.rs
```diff
@@ -36,13 +36,16 @@ impl GridController {
         &mut self,
         sheet_id: SheetId,
         pos: Pos,
-        value: &str,
+        raw_value: &str,
```
ddimaria, line 39: Shadowing is OK in Rust, so you don't have to change this name.  It's a complete 180 degree shift from JS.  You'd just need to update line 47: 
```rust
let value = value.trim();
```

ddimaria (approved):
Easy to review this one!  I have one comment, but it's not important just a note.

quadratic-core/src/controller/cells.rs
davidkircos: nice!

davidfig (approved):

Issue #604, opened 2023-07-11
feat: export to CSV


waveminded:
The ability to export the entire grid or selected cells of the grid to CSV. Potentially 2 options. 

1. The ability to 'right click' a selected cell area and export to CSV
2. A secondary option to dump the whole grid.

PR #826, opened 2023-11-16
refactor: remove icons from UI

jimniels:
IMO we have too many icons. It's diluting the value/efficacy of the ones we do have. I think we should use them a lot more sparingly. My impulse is to remove a bunch, and then we can always add them back where we can justify their value.

The new command palette 

CleanShot 2023-11-16 at 16 31 28@2x

which is a lot more akin to Figma's: graphical elements are only used for 1) toggles, to indicate the on/off state of an item, or 2) external links, indicating these items will take you out of the app.

(Figma's UI)

CleanShot 2023-11-16 at 16 28 28@2x

I also removed them from dropdown menus

CleanShot 2023-11-16 at 20 34 34@2x

CleanShot 2023-11-16 at 20 35 44@2x

For now, I think it's ok to keep them in the formatting menu.

## Todos

- [ ] Once #604 is done, remove the icons from the floating flyout menu and then merge this

### DIFF
--- a/src/ui/menus/CommandPalette/ListItems/Borders.tsx
+++ b/src/ui/menus/CommandPalette/ListItems/Borders.tsx
@@ -1,15 +1,3 @@
-import {
-  BorderAll,
-  BorderBottom,
-  BorderClear,
-  BorderHorizontal,
-  BorderInner,
-  BorderLeft,
-  BorderOuter,
-  BorderRight,
-  BorderTop,
-  BorderVertical,
-} from '@mui/icons-material';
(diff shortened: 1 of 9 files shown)


ddimaria:
You've been cursed by the single commit "feature" in Heroku

davidkircos:
For me it's easier to skim for symbols than it is to read each item. Esp for formatting items icons make a lot of sense to me.

Why do you want to remove them?

jimniels:
This is going to be superceeded by the work in #840

PR #827, opened 2023-11-17
Fixes imports for both 1.3 and 1.4+ files

davidfig:


src/dashboard/components/CreateFileButton.tsx
```diff
@@ -29,12 +29,23 @@ export default function CreateFileButton() {
...
+      };
+    }
+    // versions > 1.3
+    else {
+      data = {
+        name: file.name ? file.name.replace('.grid', '') : 'Untitled',
+        version: validFile.version,
+        contents: validFile.contents, //JSON.stringify(validFile),
+      };
+    }
```
jimniels, line 48: After reading through this, it seems the only difference is the `contents` so I think the code would be easier to read if it was more condensed like this:

```js
const data: CreateActionRequest = {
  name: file.name ? file.name.replace('.grid', '') : 'Untitled',
  version: validFile.version,
  contents: validFile.version === '1.3' ? JSON.stringify(validFile) : validFile.contents
};
```

jimniels (changes requested):

src/dashboard/components/CreateFileButton.tsx
davidfig: Sure

jimniels (approved):

ddimaria (approved):

PR #831, opened 2023-11-17
cell sheet may not be defined if deleted

davidfig:
Removed the exception in cellSheet.ts as we do expect cellsHash to be undefined if all the data from that has was deleted.


src/gridGL/cells/CellsSheet.ts
```diff
@@ -371,12 +371,9 @@ export class CellsSheet extends Container {
   modified(modified: CellSheetsModified[]): void {
     for (const update of modified) {
       const cellsHash = this.getCellsHash(Number(update.x) * sheetHashWidth, Number(update.y) * sheetHashHeight, true);
-
-      // cellsHash should always exists since we createAsNeeded = true
```
ddimaria, line 375: Curious why this statement is no longer valid?

ddimaria (approved):

PR #836, opened 2023-11-17
feat: export grid selection as csv

ddimaria:
- [x] Select rectangle on grid, select Download Selection as CSV, downloads CSV with grid file name and the file name.
- [x] CSV export in Command Palette
- [x] CSV export in DataMenu
- [x] Add Shortcut Command + Shift + E (for export, follows Figma export command)
- [x] Add Rust tests


src/grid/controller/Grid.ts
```diff
@@ -604,6 +604,24 @@ export class Grid {
 
   //#endregion
 
+  //#region Exports
+
+  exportCsvSelection(): string {
+    if (!this.gridController) throw new Error('Expected grid to be defined in Grid');
+
+    debugTimeReset();
+    console.log(rectangleToRect(sheets.sheet.cursor.getRectangle()));
```
davidkircos, line 613: remove log

davidkircos (approved):

src/grid/actions/downloadSelectionAsCSV.ts
```diff
@@ -0,0 +1,7 @@
+import { downloadFile } from '@/helpers/downloadFileInBrowser';
+import { grid } from '../controller/Grid';
+
+export const downloadSelectionAsCSV = (fileName: string) => {
+  const csv = grid.exportCsvSelection();
+  downloadFile(fileName, csv, 'text/plain', 'csv');
+};
```
jimniels, line 7: We have been slowly working on a centralized place for defining a bunch of actions/labels that get re-used like this.  (We haven't done it yet for "Copy selection as PNG" but should...)

It's probably worth sticking this in there and re-using it?

`src/actions.ts` is the file

Something like

```js
// src/actions.ts
export const downloadSelectionAsCSV = {
  label: 'Download selection as CSV",
  run: (fileName: string) => {
    const csv = grid.exportCsvSelection();
    downloadFile(fileName, csv, 'text/plain', 'csv');
  }
};
```

jimniels: Also minor casing thing:

We have it named `exportCsvSelection` in another file but `downloadSelectionAsCSV` here. We should either capitalize "CSV" everywhere, or stick to camelCase.

I would say we stick to camelCase, as we'll do things like `fileId` elsewhere, so this should be `downloadSelectionAsCsv`

src/ui/menus/CommandPalette/ListItems/Edit.tsx
```diff
@@ -109,6 +111,24 @@ const ListItems = [
...
+          action={() => {
+            downloadSelectionAsCSV(name);
+          }}
+          icon={<Download />}
+          shortcut="E"
+          shortcutModifiers={[KeyboardSymbols.Command, KeyboardSymbols.Shift]}
+        />
+      );
+    },
+  },
```
jimniels, line 131: Per the comment about centralizing things in `src/actions.ts` here you could import the label/action and re-use it.

src/ui/menus/TopBar/SubMenus/DataMenu.tsx
```diff
@@ -32,13 +30,19 @@ export const DataMenu = () => {
...
+        <MenuItem
+          onClick={() => {
+            downloadSelectionAsCSV(fileName);
+          }}
+        >
+          <MenuLineItem
+            primary="Selection to CSV"
+            Icon={UploadFile}
+            secondary={KeyboardSymbols.Command + KeyboardSymbols.Shift + 'E'}
+          />
```
jimniels, line 43: We're using a different icon in this menu than the other menus.

CleanShot 2023-11-20 at 15 44 30@2x

We should use the same icon we're using in the other menus

CleanShot 2023-11-20 at 15 44 36@2x

CleanShot 2023-11-20 at 15 44 45@2x

jimniels (changes requested):
Looks pretty good. Love this feature! Just a few things to cleanup then it LGTM

src/ui/menus/TopBar/SubMenus/DataMenu.tsx
jimniels: @davidkircos can you update the icon on this one to use the same one we're using in the command palette? Looks like it's `<Download />`

jimniels (approved):
One thing (update icon) and then LGTM 👍

PR #838, opened 2023-11-17
feat: better file errors

jimniels:
When you navigate to a file and it fails to load, today there's really only one generic message: a 404.

But it's not always a file not found error. Sometimes there are other errors. Here are some examples:

- Forbidden: the file exists but the user doesn't have the right to access it (do we want to actually want to communicate this to the user? is it a security/privacy concern? e.g. "you found a file that does in fact exist, but you can't access it")
- Broken file: the file exists but it couldn't be loaded into the app (e.g. it was returned from the server but the app failed to load it)
- Other...?

### DIFF
--- a/src/api/fetchFromApi.ts
+++ b/src/api/fetchFromApi.ts
@@ -23,13 +23,13 @@ export async function fetchFromApi<T>(path: string, init: RequestInit, schema: z
 
   // Handle response if a server error is returned
   if (!response.ok) {
-    const error = await newHTTPError('Unsuccessful response', response, init.method);
+    const error = await createAPIError('Unsuccessful response', response, init.method);
     throw error;
   }
 
   // Handle if the response is not JSON
   const json = await response.json().catch(async () => {
-    const error = await newHTTPError('Not a JSON body', response, init.method);
+    const error = await createAPIError('Not a JSON body', response, init.method);
     throw error;
   });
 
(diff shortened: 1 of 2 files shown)


quadratic-api/src/routes/files/fileMiddleware.ts
```diff
@@ -3,6 +3,7 @@ import dbClient from '../../dbClient';
 import { Request } from '../../types/Request';
 
 export const fileMiddleware = async (req: Request, res: Response, next: NextFunction) => {
+  // TODO: validate UUID
```
jimniels, line 6: This is an improvement: we can validate the query param is a UUID and say so if it isn't.

This differentiates between a bad request `/file/foo` and a valid request that's a 404 `/file/:valid-uuid`

quadratic-client/src/api/fetchFromApi.ts
```diff
@@ -19,35 +33,35 @@ export async function fetchFromApi<T>(path: string, init: RequestInit, schema: z
...
+  const json = await response.json().catch((error) => {
+    Sentry.captureException(error);
+    throw new ApiError('An unknown error occurred: response is not JSON.', response.status, init.method);
+  });
 
   // Handle response if a server error is returned
   if (!response.ok) {
-    const error = await newHTTPError('Unsuccessful response', response, init.method);
-    throw error;
+    // TODO: ensure API only ever returns uniform error response, e.g. `json.errors` or `json.error.message`
```
jimniels, line 47: We have places where, when an error happens, the server might respond with `{ errors: [] }` or `{ error: { message: ""}}` and I'm thinking maybe we should have one standard error response but we can do that separately. Here we check for both.

quadratic-client/src/constants/urls.ts
```diff
@@ -5,3 +5,4 @@ export const DOCUMENTATION_FILES_URL = `${DOCUMENTATION_URL}/files`;
 export const BUG_REPORT_URL = 'https://github.com/quadratichq/quadratic/issues';
 export const DISCORD = 'https://discord.gg/quadratic';
 export const TWITTER = 'https://twitter.com/quadratichq';
+export const SUPPORT_URL = 'https://quadratichq.com/support';
```
jimniels, line 8: Where do we point people for support? We should have one standard place for this

jimniels: Change this to `/contact` and it will depend on https://github.com/quadratichq/quadratic-website/pull/39

quadratic-api/src/routes/files/fileMiddleware.ts
jimniels: Leaving this as a TODO because the teams branch has more helper utils around making this better. Will pick up then.

quadratic-client/src/api/fetchFromApi.ts
jimniels: Will leave this for later in the teams work as we standardize response formats

quadratic-client/src/dashboard/FileRoute.tsx
```diff
@@ -109,38 +92,62 @@ export const Component = () => {
...
     );
   }
 
-  // Maybe we log this to Sentry someday...
+  if (title && description) {
+    return <Empty title={title} description={description} Icon={ExclamationTriangleIcon} actions={actions} />;
+  }
+
+  // If we reach here, it's an error we don't know how to handle.
+  // TODO: probably log this to Sentry...
```
davidkircos, line 143: Is this logged to sentry with the console.error below?

quadratic-client/src/api/fetchFromApi.ts
```diff
@@ -19,35 +33,35 @@ export async function fetchFromApi<T>(path: string, init: RequestInit, schema: z
...
-  });
-
   // Compare the response to the expected schema
   const result = schema.safeParse(json);
   if (!result.success) {
-    console.error(result.error);
-    const error = await newHTTPError('Unexpected response schema', response, init.method);
-    throw error;
+    console.log('Schema validation failed.');
+    console.log(result.error);
```
davidkircos, line 61: remove console logs?

quadratic-client/src/dashboard/FileRoute.tsx
jimniels: I don't know? I'm still not clear how Sentry works. Does it wrap the app and send off anything that calls`console.error`?

quadratic-client/src/api/fetchFromApi.ts
jimniels: Personally I find these handy because they allow us to debug this stuff better in prod. If a file doesn't open, you can open the console and zod will tell you where it's failing.

ddimaria: Maybe move to a log package (e.g. Winston) so you can conditionally see these as debugs, logs or traces?  This will also help with the noise in the new API output in local development.

davidkircos (approved):

PR #840, opened 2023-11-17
feat: implement a command palette that uses components from shadcn/ui

jimniels:
- Implement the command palette using [the one from shadcn/ui](https://ui.shadcn.com/docs/components/command)
- #1149 
  - Removed "Command palette" from the "Quadratic Menu" dropdown and gave it its own spot in the top-level header bar

CleanShot 2024-03-11 at 16 47 22@2x

CleanShot 2024-03-11 at 16 47 49@2x

**Notes**

- When you filter to an item that's at the bottom of the list, then clear your query, the scroll position in the list doesn't reset. This seems to be a feature-and-or-bug with the cmdk lib itself. [I filed an issue](https://github.com/pacocoursey/cmdk/issues/234) and we can track it there. Shouldn't stop us from shipping this as-is IMO.


jimniels:
Closing this. Will have to revisit later.

jimniels:
Every single command doesn't have an icon. For example: in the "more" menu of the formatting bar

CleanShot 2024-03-08 at 14 39 54@2x

It can be difficult to meaningfully differentiate between similar commands like "Paste", "Paste with values", "Paste with formatting" etc.

Similarly, in the command palette there are actions like "Run code", "Run code in sheet" and "Run code in all sheets"

CleanShot 2024-03-08 at 14 13 39@2x

I explored some options for doing this in radix-icon style

CleanShot 2024-03-08 at 14 41 03@2x

But ultimately decided, for now, to just not put icons in places where commands are very similar but different.

davidkircos (approved):

PR #844, opened 2023-11-17
don't overwrite code changes

### LINKED ISSUE #842: bug: transaction finishing shifts focus out of code editor
**Description**

I was writing code in the code editor that was accessing the network and took a moment to complete.

I hit `Cmd` + `Enter`, the code started to run and while it was running I began writing more code. 

In the middle of writing new lines of code, the transaction completed and it redrew the output on the grid and completely lost the state of where I was in the code editor.

https://github.com/quadratichq/quadratic/assets/1316441/3174e521-879d-467d-9cd1-e7f1a1fe33e9

**Desired behavior**

Focus/state of the code editor should never be lost, even while transactions run.

davidfig:
Code Editor will not overwrite code changes after a transaction completes. (It only overwrites the saved code, so it will still pop a message on exit if the code is different from the one stored in Rust).

Fixes #842


src/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -35,6 +35,7 @@ export const CodeEditor = () => {
   const [codeString, setCodeString] = useState('');
   const [out, setOut] = useState<{ stdOut?: string; stdErr?: string } | undefined>(undefined);
   const [evaluationResult, setEvaluationResult] = useState<any>(undefined);
+
   const updateCodeCell = useCallback(() => {
```
davidkircos, line 39: what does this do now?

davidkircos (approved):

PR #846, opened 2023-11-19
feat: Chart Support

davidfig:
- [x] html w/h should be f32 so it better matches with gridLines
- [x] escape key cancels resizing of html
- [x] bug: HTML does not properly remove when no longer returning an HTML value
- [x] bug when changing html -- zoom events are no longer caught properly
- [x] bug: code editor / terminal can't be resized
- [x] bug with spills :(
- [x] snap to grid
- [x] change "HTML" label to "CHART"
- [x] try changing border to same as code cell output
- [x] panning and zooming the viewport when hovering over html and div
- [x] thumbnail placeholders
- [x] html support for python cells
- [x] resizing html
- [x] div support for python cells (sort of working)
- [x] resizing div
- [x] update html for changes
- [x] duplicate sheets w/html
- [x] rename OutputSize to RenderSize
- [x] returning a plotly object should call to_html()
- [x] fix pinch zooming the entire document
- [x] add highlight color when hovering over resize edges
- [x] code comments and tests 

later
- [ ] bug: when resizing a column that would move a chart. the chart doesn't move until the page is reloaded
- [ ] default to no icon bar, just pointer hovers for charts { 'displaylogo': False, 'displayModeBar': True }
- [ ] heading is behind HTML -- may need another renderer just for column/row headings
- [ ] .show() support
- [ ] snapping (removed in build since it was not exact)

launch
- [x] Update examples
- [x] Update docs
- [ ] Update marketing website


ddimaria:
It looks like coverage went down:
Screenshot 2023-11-22 at 10 42 01 AM

quadratic-core/src/grid/sheet/rendering.rs
```diff
@@ -86,6 +87,20 @@ impl Sheet {
                         italic: Some(true),
                         text_color: Some(String::from("red")),
                     }
+                } else if value.type_name() == "html" {
```
HactarCE, line 90: I rewrote this section to use a `match` instead of comparing based on `.type_name()`

quadratic-core/src/grid/sheet/rendering.rs
```diff
@@ -124,6 +139,31 @@ impl Sheet {
             .collect()
     }
 
+    pub fn get_html_output(&self) -> Vec<JsHtmlOutput> {
+        self.code_cells
+            .iter()
+            .filter_map(|(cell_ref, code_cell_value)| {
+                if let Some(output) = code_cell_value.get_output_value(0, 0) {
+                    if output.type_name() != "html" {
```
HactarCE, line 147: use `matches!()` instead of `.type_name() ==`

quadratic-core/src/grid/sheet/rendering.rs
```diff
@@ -124,6 +139,31 @@ impl Sheet {
             .collect()
     }
 
+    pub fn get_html_output(&self) -> Vec<JsHtmlOutput> {
+        self.code_cells
+            .iter()
+            .filter_map(|(cell_ref, code_cell_value)| {
+                if let Some(output) = code_cell_value.get_output_value(0, 0) {
```
HactarCE, line 146: use `?` to propogate the `None`; no indent necessary. otherwise I have to read all the way to the bottom of the closure to see what you've done with the `None` case.

quadratic-core/src/values/cellvalue.rs
```diff
@@ -85,6 +88,7 @@ impl CellValue {
             CellValue::Instant(_) => todo!("repr of Instant"),
             CellValue::Duration(_) => todo!("repr of Duration"),
             CellValue::Error(_) => "[error]".to_string(),
+            CellValue::Html(s) => s.clone(),
```
HactarCE, line 91: Is it possible to return HTML from a formula? Probably not, but if it is then what's the syntax for that?

quadratic-core/src/values/cellvalue.rs
```diff
@@ -290,6 +296,8 @@ impl CellValue {
             | (CellValue::Instant(_), _)
             | (CellValue::Duration(_), _)
             | (CellValue::Blank, _) => return Ok(None),
+
+            (CellValue::Html(_), _) => return Ok(None),
```
HactarCE, line 300: This can be lumped in with the previous case. Just add another `| (CellValue::Html(_), _)` before the arrow and let rustfmt make it pretty

quadratic-core/src/values/cellvalue.rs
```diff
@@ -427,13 +436,24 @@ impl CellValue {
...
             value = CellValue::Text(s.to_string());
         }
         (value, ops)
     }
+
+    pub fn is_html(&self) -> bool {
+        match self {
+            CellValue::Html(_) => true,
+            _ => false,
+        }
```
HactarCE, line 455: `matches!(self, CellValue::Html(_))`

quadratic-core/src/wasm_bindings/controller/formatting.rs
```diff
@@ -223,6 +224,29 @@ impl GridController {
...
+        w: i64,
+        h: i64,
+        cursor: Option<String>,
+    ) -> Result<JsValue, JsValue> {
+        let sheet_id = SheetId::from_str(&sheet_id).unwrap();
+        let value = if w == 0 && h == 0 {
+            None
+        } else {
+            Some(RenderSize { w, h })
+        };
```
HactarCE, line 244: If `RenderSize` and `ArraySize` are equivalent, then this is simply `ArraySize::new()`.

In general, is `RenderSize { w: 0, h: 0 }` allowed to exist anywhere?

HactarCE (dismissed):

quadratic-client/src/gridGL/htmlCells/HTMLResizeControl.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <div
+      className={`resize-control resize-control--position-${position}`}
+      style={{
+        // @ts-expect-error typescript doesn't like us setting CSS custom properties
+        '--resize-control-highlight': colors.quadraticPrimary,
+        '--resize-control-background': colors.mediumGray,
+        pointerEvents: 'none',
+      }}
+      // onPointerDown={(e) => {
```
ddimaria, line 21: Remove commented-out code?

quadratic-core/src/controller/operations.rs
```diff
@@ -94,12 +97,28 @@ impl GridController {
...
                 summary.code_cells_modified.insert(cell_ref.sheet);
+
+                // track changes for html
+                if let Some(code_cell_value) = code_cell_value.as_ref() {
+                    if let Some(code_cell_value) = code_cell_value.get_output_value(0, 0) {
+                        if code_cell_value.is_html() {
+                            summary.html.insert(cell_ref.sheet);
+                        }
+                    }
+                } else if let Some(old_code_cell_value) = &old_code_cell_value {
```
ddimaria, line 114: Some duplication here with code above, consider refactoring.

quadratic-core/src/controller/operations.rs
```diff
@@ -181,6 +200,15 @@ impl GridController {
                             self.set_cell_formats_for_type::<FillColor>(&region, fill_color, None),
                         )
                     }
+                    CellFmtArray::RenderSize(output_size) => {
+                        // todo: this should send the updated html to the client
+                        // summary.html.push(region.sheet);
```
ddimaria, line 205: Does line [242](https://github.com/quadratichq/quadratic/pull/846/files#diff-569e8c04577f8c41e69ec3972938bd703f2d8a375fc14f8878cfca3853c2eda8R242) now cover this?

quadratic-core/src/grid/sheet.rs
```diff
@@ -111,27 +111,22 @@ impl Sheet {
...
-        // }
-
-        // TODO: check for new spills, if the cell was deleted
-        let spill = None;
+        // returns if there's a change in html (html cell is added or removed from sheet)
+        let html = if let Some(value) = &value {
+            old_value.is_html() != value.is_html()
+        } else {
+            old_value.is_html()
+        };
```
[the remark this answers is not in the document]
davidfig: i slightly prefer the if statements since it's quicker for me to understand

quadratic-core/src/values/cellvalue.rs
davidfig: Sure. if you return a string starting with <div> or <html> it should turn it into CellValue::Html(). i'm not sure if that's implemented yet. Similarly, you can also output the value of a cell that has HTML to propagate the HTML value

quadratic-core/src/grid/sheet/code.rs
```diff
@@ -94,7 +94,36 @@ impl Sheet {
         self.code_cells.keys().copied()
     }
 
-    // fn unspill(&mut self, source: CellRef) {
-    //     todo!("unspill cells from {source:?}");
-    // }
+    /// returns the output_size for a html-like cell; returns 0, 0 if not set
+    pub fn html_output_size(&self, pos: Pos) -> (i64, i64) {
```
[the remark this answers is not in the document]
davidfig: Replaced this with render_size(), which just returns Option<RenderSize>

davidkircos (approved):

PR #853, opened 2023-11-21
Improve Performance When Updates Large Regions of CellValues

davidkircos:
- [x] Reduce SetCellValues to a single call on:
- [x] - CSV
- [x] - Paste
- [x] - Autocomplete I think only calls SetCellValues 3 times in total so it is fast.
- [x] - Delete Cells

When we were searching for code cells by looping through x, y and checking each pos if there was a code cell there. I changed the search to looping through each code cell and seeing if it was in the updated rect.


quadratic-core/src/computation/mod.rs
```diff
@@ -16,6 +16,7 @@ use crate::controller::{
 #[derive(Debug, Default, Clone)]
 pub struct TransactionInProgress {
     reverse_operations: Vec<Operation>,
+    cells_updated: IndexSet<RegionRef>,
     cells_to_compute: IndexSet<CellRef>,
```
davidkircos, line 20: I am wondering if we can convert cells_to_compute to an IndexSet of only CodeCells

Every time we update values we can add to the `cells_updated` by region, and when we find code cells to update we can add them to `cells_to_compute`

ddimaria: Maybe add a TODO or create an issue for this work?

quadratic-core/src/computation/transaction_in_progress.rs
```diff
@@ -52,6 +53,11 @@ impl TransactionInProgress {
         // apply operations
         transaction.transact(grid_controller, operations, compute);
 
+        // crate::util::dbgjs(&format!(
```
ddimaria, line 56: We should add in a logging crate so we can just have these message conditionally display.

quadratic-core/src/computation/transaction_in_progress.rs
```diff
@@ -258,19 +248,25 @@ impl TransactionInProgress {
...
+                self.cells_to_compute.extend(dependent_cells);
+            }
+        }
+
         if let Some(cell_ref) = self.cells_to_compute.shift_remove_index(0) {
             // todo: this would be a good place to check for cycles
             // add all dependent cells to the cells_to_compute
             if let Some(dependent_cells) = grid_controller.get_dependent_cells(cell_ref) {
-                self.cells_to_compute.extend(dependent_cells);
+                #[cfg(feature = "show-operations")]
```
ddimaria, line 261: Ditto on this `show-operations` feature...could all just be conditional on logging-levels.

quadratic-core/src/computation/transaction_in_progress.rs
```diff
@@ -258,19 +248,25 @@ impl TransactionInProgress {
...
-                    if cfg!(feature = "show-operations") {
-                        if let Some(pos) = sheet.cell_ref_to_pos(*cell_ref) {
-                            crate::util::dbgjs(format!("[Adding Dependent Cell] {:?}", pos));
-                        }
+                    if let Some(pos) = sheet.cell_ref_to_pos(*cell_ref) {
+                        crate::util::dbgjs(format!("[Adding Dependent Cell] {:?}", pos));
                     }
                 });
+
+                self.cells_to_compute.extend(dependent_cells);
```
ddimaria, line 269: Do we care about duplicates here?

quadratic-core/src/controller/auto_complete.rs
```diff
@@ -570,10 +570,21 @@ impl GridController {
             negative,
         });
 
-        let array = Array::new_row_major(range.size(), series.to_owned().into())
+        let values = Array::new_row_major(range.size(), series.to_owned().into())
             .map_err(|e| anyhow!("Could not create array of size {:?}: {:?}", range.size(), e))?;
 
-        let ops = self.set_cells_operations(sheet_id, range.min, array);
+        let start_pos = range.min;
```
ddimaria, line 576: Just a note to self, but we should implement [add](https://doc.rust-lang.org/rust-by-example/trait/ops.html) for adding an `Array` to a `Pos` that results in a `Rect`.  I've seen this done in other places.

quadratic-core/src/controller/import.rs
```diff
@@ -28,29 +28,44 @@ impl GridController {
...
                     })
-                    .collect::<Result<Vec<Vec<Operation>>>>()
+                    .collect::<Result<Vec<CellValue>>>()
             })
-            .collect::<Result<Vec<Vec<Vec<Operation>>>>>()?
-            .into_iter()
-            .flatten()
-            .flatten()
-            .collect::<Vec<Operation>>();
+            .collect::<Vec<Vec<CellValue>>>();
```
ddimaria, line 52: Some duplication here and in other places (e.g. clipboard).  Can you add a TODO to refactor into a separate function?

quadratic-core/src/controller/operations.rs
```diff
@@ -36,12 +37,13 @@ impl GridController {
...
                     .map(|(cell_ref, value)| {
                         let pos = sheet.cell_ref_to_pos(cell_ref)?;
-                        cells_to_compute.insert(cell_ref);
                         let response = sheet.set_cell_value(pos, value)?;
                         Some(response.old_value)
                     })
                     .map(|old_value| old_value.unwrap_or(CellValue::Blank))
                     .collect();
+                cells_updated.insert(region.clone());
+                // crate::util::dbgjs(&format!("[region updated] {}", region.clone())); TODO: REMOVE
```
ddimaria, line 46: Remove?

quadratic-core/src/grid/ids.rs
```diff
@@ -79,6 +79,12 @@ impl From<CellRef> for RegionRef {
     }
 }
 
+impl Display for RegionRef {
```
ddimaria, line 82: 💯

quadratic-core/src/grid/ids.rs
```diff
@@ -104,6 +110,13 @@ impl RegionRef {
     pub fn is_empty(&self) -> bool {
         self.columns.len() == 0 && self.rows.len() == 0
     }
+
+    // returns if a cell is in the region
+    pub fn contains(&self, cell: &CellRef) -> bool {
```
ddimaria, line 115: Add a test for this?

ddimaria (approved):
Looks great!  I made some comments, but none are blocking.

quadratic-core/src/computation/transaction_in_progress.rs
davidkircos: It's an IndexSet so it shouldn't add the duplicate?

ddimaria: Thanks, I forgot about that (thought it was a `vec` because of the `extend`).  I should start reviewing these in VS Code instead of the GitHub UI.

quadratic-core/src/computation/mod.rs
davidfig: I did this in the id branch.

PR #854, opened 2023-11-21
remove extra px

davidkircos:
Before:
CleanShot 2023-11-20 at 20 39 08@2x

After:
CleanShot 2023-11-20 at 20 41 45@2x


davidfig (approved):

davidfig:
I can't really see a difference. Not sure why I put the +1 in there. But if you like it better, go for it.

PR #861, opened 2023-11-21
feat: improve bottom bar stats functionality

jimniels:
The current implementation of the bottom bar outputs a number that is _really_ big.

CleanShot 2023-11-21 at 14 00 58@2x

Excel caps this number at 9 digits beyond the decimal

CleanShot 2023-11-21 at 14 01 25@2x

So we do the same thing:

CleanShot 2023-11-21 at 14 01 08@2x

Also: excel provides the ability for you to click and copy the values in the bottom bar. This allows you to do the same thing in Quadratic.

CleanShot 2023-11-21 at 14 08 22

### DIFF
--- a/src/ui/menus/BottomBar/SelectionSummary.tsx
+++ b/src/ui/menus/BottomBar/SelectionSummary.tsx
@@ -2,6 +2,7 @@ import { useMediaQuery } from '@mui/material';
 import { useEffect, useState } from 'react';
 import { sheets } from '../../../grid/controller/Sheets';
 // import { getColumnA1Notation, getRowA1Notation } from '../../../gridGL/UI/gridHeadings/getA1Notation';
+import { TooltipHint } from '@/ui/components/TooltipHint';
 import { grid } from '../../../grid/controller/Grid';
 import BottomBarItem from './BottomBarItem';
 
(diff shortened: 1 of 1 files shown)


src/ui/menus/BottomBar/SelectionSummary.tsx
```diff
@@ -40,12 +42,33 @@ export const SelectionSummary = () => {
     cursor.terminalPosition.y,
   ]);
 
+  const handleOnClick = (valueToCopy: string) => {
+    navigator.clipboard.writeText(valueToCopy).then(() => {
```
davidkircos, line 46: I think this causes problems in firefox? I think we have to check the `fullClipboardSupport` flag first.

davidkircos (changes requested):

src/ui/menus/BottomBar/SelectionSummary.tsx
jimniels: `navigator.clipboard.writeText` works in all browsers. It's [the `navigator.clipboard.write` you gotta watch out for](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard).

This code works for me in Firefox for mac



https://github.com/quadratichq/quadratic/assets/1316441/afce00c4-ffd8-463b-8adb-70cbd165d283

davidkircos (approved):

PR #872, opened 2023-11-23
upgrade to pyodide 0.24.1

### LINKED ISSUE #855: bug: some pyodide supported Python libraries not working in-app
Currently a few major versions of pyodide behind - upgrading is good place to start and then see if issues persist after upgrade. 

examples that don't work (will log as I run into them) - scipy stats, scikit-learn (presumably many others in the supported list on pyodide)

image
image
image

davidkircos:
CleanShot 2023-11-22 at 18 20 36@2x

fixes #855


ddimaria (approved):
No code changes to review other than `package.json`.  The Python example file works as expected.  Should we add any newly supported packages to that file?

PR #892, opened 2023-11-28
Improvements to npm scripts

davidfig:
note: this removes `npm run watch:wasm:types` from `npm start` to avoid double rust compiles on rust changes. `npm run build:wasm:types` is run only once when the script starts. To rebuild the rust types, you need to manually call `npm run build:wasm:types`, or restart `npm start`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -49,7 +49,6 @@ jobs:
 
       - name: Run Clippy
         run: |
-          cd quadratic-client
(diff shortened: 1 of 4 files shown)


ddimaria:
@davidfig can you add the "To rebuild the rust types, you need to manually call npm run build:wasm:types, or restart npm start" instructions to the CONTRIBUTING.md file?

ddimaria (approved):

Issue #877, opened 2023-11-24
bug: too many digits by default with formulas (default to 9)


davidkircos:
- [x] on formula return
- [x] in bottombar

![CleanShot 2023-11-23 at 19 59 38@2x](https://github.com/quadratichq/quadratic/assets/3479421/e6939261-6e01-434c-a6ef-a45e45ef58d9)

PR #893, opened 2023-11-28
Round to 9 decimal places for formulas w/o a user-set decimal places

ddimaria:
Addresses https://github.com/quadratichq/quadratic/issues/877

- [x] Round to 9 decimal places when user hasn't set decimals

### DIFF
--- a/quadratic-core/src/grid/sheet.rs
+++ b/quadratic-core/src/grid/sheet.rs
@@ -460,10 +460,13 @@ impl Sheet {
             match value {
                 CellValue::Number(n) => {
                     let (_, exponent) = n.as_bigint_and_exponent();
+                    let max_decimals = 9;
+                    let decimals = exponent.min(max_decimals) as i16;
+
                     if is_percentage {
-                        Some(exponent as i16 - 2)
+                        Some(decimals - 2)
                     } else {
-                        Some(exponent as i16)
+                        Some(decimals)
                     }
                 }
                 _ => None,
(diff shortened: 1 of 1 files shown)


davidkircos:
Bottom bar still shows too many decimals for sum. Everything else appears to work as expected!

CleanShot 2023-11-27 at 20 56 54@2x

quadratic-client/src/ui/menus/BottomBar/SelectionSummary.tsx
```diff
@@ -16,12 +16,12 @@ export const SelectionSummary = () => {
   const [copied, setCopied] = useState(false);
 
   const runCalculationOnActiveSelection = () => {
-    let result = grid.summarizeSelection();
+    let result = grid.summarizeSelection(9);
```
davidfig, line 19: Can you make the 9 a constant

ddimaria:
> Bottom bar still shows too many decimals for sum. Everything else appears to work as expected!
> 
> CleanShot 2023-11-27 at 20 56 54@2x

This is resolved.  The frontend was forcing a decimal size, now rounding happens in Rust, though significant digits are supplied by TS.

davidfig (approved):

quadratic-client/src/ui/menus/BottomBar/SelectionSummary.tsx
ddimaria: This is done.

davidkircos (changes requested):
sum and average show more than 9 decimals

CleanShot 2023-11-28 at 11 58 34@2x

ddimaria:
There was an issue in the summary where we convert to a f64 after rounding that introduced unwanted behavior.....because [floats](https://0.30000000000000004.com/).  I added a round function for working directly f64s to mitigate this.

davidkircos (approved):

PR #897, opened 2023-11-28
fix fills so it properly renders large numbers

davidfig:


ddimaria (approved):
The code looks good.  I don't have a Heroku login, so can't trigger an API server build due to the 1-commit issue.  I ran this locally and it works as expected.

PR #899, opened 2023-11-28
Update GitHub Actions

davidkircos:
- [x] Moves all GH actions to a single easy to maintain file
- [x] Speed up CI!
- [x] Verify new tests do everything the previous ones do
- [x] report coverage
- [x] backup in case `self-hosted` is down
- [x] Verify optimization happens when it should

With our self-hosted runners all online tests run fast:
CleanShot 2023-12-03 at 21 43 34@2x

All our self-hosted runners are tagged `ubuntu-latest-8-cores` if our self-hosted runners are not available or if more jobs are run than the eight available. GitHub will spin up hosted runners with the same tag `ubuntu-latest-8-cores`. 

With our self-hosted runners all offline. Tests still pass but take longer:
CleanShot 2023-12-03 at 17 17 20@2x

This is the current speed of actions on `main`
CleanShot 2023-12-04 at 10 07 06@2x


vercel.json
```diff
@@ -1,4 +0,0 @@
-{
-  "buildCommand": "cd quadratic-client && npm run build:ci",
-  "installCommand": "cd quadratic-client && npm install"
-}
```
davidkircos, line 4: Pretty sure this is unused as it is defined in the vercel ui

quadratic-core/Cargo.toml
```diff
@@ -83,6 +83,8 @@ harness = false
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
 
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = ['-Os', '-g']  # TODO: -g seems to fix the name mangling problem
```
davidkircos, line 87: see 🧵 https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1701655393730929

quadratic-api/package.json
```diff
@@ -13,7 +13,7 @@
     "lint:prettier:write": "prettier --write \"src/**/*.ts\"",
     "prisma:migrate": "npx prisma migrate dev",
     "prisma:migrate:create": "npx prisma migrate dev --create-only",
-    "docker:up": "docker-compose up -d && sleep 1",
+    "docker:up": "docker-compose up -d && sleep 3",
```
ddimaria, line 16: Curious, what's the need for this `sleep`?

amplify.yml
```diff
@@ -11,7 +11,8 @@ frontend:
 
     build:
       commands:
-        - npm run build:wasm
```
ddimaria, line 14: Is this already built in GHA?

ddimaria (approved):
Looks great!  I just had a couple of questions, but nothing blocking.

quadratic-api/package.json
davidkircos: It takes a sec for Postgres to be available. Commands that actually wait for Postgres to be up are really complicated. This works 🤷‍♂️

amplify.yml
davidkircos: Amplify is a separate pipeline

PR #903, opened 2023-11-29
Basic ws server

ddimaria:


quadratic-multiplayer/src/message.rs
```diff
@@ -0,0 +1,200 @@
...
+use uuid::Uuid;
+
+use crate::state::{Room, State, User};
+
+// NOTE: needs to be kept in sync with multiplayerTypes.ts
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(tag = "type")]
+pub(crate) enum MessageRequest {
+    EnterRoom {
+        user_id: String,
```
ddimaria, line 25: Why string, I think this was a UUID?

quadratic-multiplayer/src/message.rs
```diff
@@ -0,0 +1,200 @@
...
+                y,
+            };
+
+            broadcast(user_id, file_id, Arc::clone(&state), response.clone())?;
+
+            Ok(response)
+        }
+
+        // User changes their selection
+        MessageRequest::ChangeSelection {
```
ddimaria, line 122: Can you add a test for this in server.rs?

quadratic-api/src/middleware/team.ts
```diff
@@ -19,6 +21,9 @@ export const teamMiddleware = async (
   res: Response<ResponseError>,
   next: NextFunction
 ) => {
+  await validateAccessToken(req, res, () => {});
+  await userMiddleware(req, res, () => {});
```
davidkircos, line 25: Why add these here?

davidfig: these are from Jim's branch

davidfig: Although not sure why there's a diff for them showing up

davidkircos: This is a PR to jims branch?

quadratic-client/src/gridGL/interaction/pointer/Pointer.ts
```diff
@@ -23,14 +24,20 @@ export class Pointer {
...
+    multiplayer.sendMouseMove();
+  };
+
   destroy() {
     const viewport = pixiApp.viewport;
     viewport.off('pointerdown', this.handlePointerDown);
     viewport.off('pointermove', this.pointerMove);
     viewport.off('pointerup', this.pointerUp);
     viewport.off('pointerupoutside', this.pointerUp);
+    pixiApp.canvas.removeEventListener('pointerleave', this.pointerLeave);
```
davidkircos, line 40: Better to attach leave events to the `window`?

davidfig: this is specific for the canvas. mouse should only work over the viewport

quadratic-multiplayer/.env.example
```diff
@@ -0,0 +1,2 @@
+HOST=127.0.0.1
+PORT=3001
```
davidkircos, line 2: I think port `443` is the one people use most of the time for WS

ddimaria: This is for local dev, so don't want ssl here.  We should have ssl/wss for any real usage thought.

quadratic-api/src/middleware/team.ts
ddimaria: I think this was by mistake when I was pairing with Jim, I'll remove.

quadratic-multiplayer/.env.example
davidkircos: Ok. sounds good. I'll add an item to a-team so we remember before shipping

PR #912, opened 2023-12-04
MP transactions

davidfig:


quadratic-client/src/multiplayer/multiplayer.ts
```diff
@@ -36,32 +44,41 @@ export class Multiplayer {
...
+      if (this.state === 'connecting') {
+        this.waitingForConnection.push(resolve);
+        return;
+      }
+
+      this.state = 'connecting';
       this.websocket = new WebSocket(import.meta.env.VITE_QUADRATIC_MULTIPLAYER_URL);
       this.websocket.addEventListener('message', this.handleMessage);
+
+      // todo: this is not ideal
```
ddimaria, line 58: What is not ideal about this?  Also, what about the `error` event for websockets?

quadratic-client/src/multiplayer/multiplayer.ts
```diff
@@ -104,19 +123,27 @@ export class Multiplayer {
...
-  update() {
-    if (!this.ready || !this.room || !this.uuid) return;
-    if (!this.websocket) {
-      throw new Error('[Multiplayer] Websocket not initialized.');
-    }
+  async sendTransaction(operations: string) {
+    await this.init();
+    const message: MessageTransaction = {
+      type: 'Transaction',
+      user_id: this.uuid!,
```
ddimaria, line 130: Not a UUID anymore?

quadratic-core/src/computation/mod.rs
```diff
@@ -33,4 +33,7 @@ pub struct TransactionInProgress {
     waiting_for_async: Option<CodeCellLanguage>,
     // true when transaction completes
     pub complete: bool,
+
+    // operations for multiplayer
+    pub multiplayer_operations: Vec<Operation>,
```
ddimaria, line 38: Why not just `operations`?

quadratic-core/src/controller/transactions.rs
```diff
@@ -32,6 +33,7 @@ impl GridController {
             TransactionType::Redo => {
                 self.undo_stack.push(transaction);
             }
+            TransactionType::Multiplayer => (),
```
ddimaria, line 36: This needed?  If so, is `MultiPlayer` the right name?  That is, would we just need a type that allows us to apply transactions?

quadratic-core/src/controller/transactions.rs
```diff
@@ -157,6 +162,21 @@ impl GridController {
             self.transaction_in_progress = Some(transaction.to_owned());
         }
     }
+
+    pub fn multiplayer_transaction(&mut self, transaction: String) -> TransactionSummary {
+        let operations: Vec<Operation> = serde_json::from_str(&transaction).unwrap();
```
ddimaria, line 167: Unsafe unwrap

quadratic-core/src/controller/transactions.rs
ddimaria: Maybe `RawTransaction` or `Operations`?

quadratic-core/src/wasm_bindings/controller/mod.rs
```diff
@@ -95,6 +95,13 @@ impl GridController {
         self.calculation_get_cells(get_cells)
     }
 
+    #[wasm_bindgen(js_name = "multiplayerTransaction")]
+    pub fn js_multiplayer_transaction(&mut self, operations: String) -> Result<JsValue, JsValue> {
```
ddimaria, line 99: Can you add a TODO to change to a `Vec<Operation>` at some point?

quadratic-multiplayer/src/message.rs
```diff
@@ -113,6 +117,24 @@ pub(crate) async fn handle_message(
             Ok(response)
         }
 
+        MessageRequest::LeaveRoom { user_id, file_id } => {
+            if let Ok(room) = state.get_room(&file_id).await {
+                if state.leave_room(file_id, &user_id).await {
```
ddimaria, line 122: Use the question mark operator to avoid the `anyhow` invocations and have the anyhow's from state propagate to the result.

quadratic-multiplayer/src/state.rs
```diff
@@ -82,6 +82,30 @@ impl State {
 
         room.users.insert(user_id, user.clone()).is_none()
     }
+
+    /// Removes a user from a room. If the room is empty, it deletes the room.
+    /// Returns true if the room still exists after the user leaves.
+    pub(crate) async fn leave_room(&self, file_id: Uuid, user_id: &String) -> bool {
```
ddimaria, line 88: Can you add a test for this?

quadratic-multiplayer/src/state.rs
```diff
@@ -82,6 +82,30 @@ impl State {
...
+        let mut rooms = self.rooms.lock().await;
+
+        // todo: there's probably a better way of handling the case where the room does not exist
+        let room = rooms.entry(file_id).or_insert_with(|| Room::new(file_id));
```
ddimaria, line 92: If the room does not exist, the function should return.  Changing the return type to `Result<bool>` to allow error responses.  They should be propagated to the logs and in the client response.

quadratic-multiplayer/src/state.rs
```diff
@@ -82,6 +82,30 @@ impl State {
...
+        room.users.remove(user_id);
+
+        // remove the room if it's empty
+        if room.users.len() == 0 {
```
ddimaria, line 96: Maybe add a `remove_room` function in `state`

quadratic-client/src/multiplayer/multiplayer.ts
davidfig: updated the comment

quadratic-client/src/multiplayer/multiplayer.ts
davidfig: True. but it's always a string either way

quadratic-core/src/controller/transactions.rs
davidfig: I like MP as it indicates where the transaction originated from. If anything, TransactionType::Normal should probably be TransactionType::User.

PR #915, opened 2023-12-05
creating rowId/columnId adds operations

davidfig:


ddimaria (approved):
Will review as part of the parent branch

PR #916, opened 2023-12-05
Chart updates

### LINKED ISSUE #929: bug: charts can be resized in read only mode

davidfig:
- [x] interactivity is controlled by an .env variable: `VITE_ALLOW_CHART_INTERACTIVITY=1` to turn it on
- [x] chart resizing is disabled if read-only access 
- [x] chart thumbnail

Fixes #929


jimniels:
### Bugs?

Idk if this is part of this PR, or more generally something with charts, but when you interact with the chart (e.g. click on one of the buttons) it steals focus from the grid. So then when you go back to trying to zoom in/out of the grid, it doesn't zoom.

To reproduce:

- Click on one of the icons in the chart
- Now try to CMD + Wheel to zoom. It doesn't zoom (doesn't matter if your cursor is over the chart or not)

Also: the resizing controls seem to get stuck when you click and drag. When I let up on the mouse button it doesn't stop dragging. It keeps acting like the mouse is down. I have to move my mouse outside the graph (fast enough) and click on the grid to get it to reset. (See video)

https://github.com/quadratichq/quadratic/assets/1316441/06d6e6fd-8c76-43ef-9a18-0be2bd283938

### Static vs. Interactive

My two cents on David's comment:

> it's better without interactivity and waiting until we can have better chart support with a custom renderer.

I'm on the fence about this. Personally, I think the interactivity of these charts feels half-baked inside Quadratic and, for the sake of a more cohesive product, I would prefer to keep the charts as static. I think it's always easier to start constrained and see if people ask for more (and ask why). So in this case, give people charts but just keep them static and see how much we hear people asking for the default "interactivity" of a plotly chart.

That said, I do see this feature as "plotly charts" which means let people have default plotly charts, however seemingly subpar because that's what they expect when they embed a plotly chart.

Will they see any incoherence or half-baked-ness as a reflection of Quadratic the product, or of Plotly itself?

Personally, I'm always a fan of starting constrained and adding more if/when people really want it. In this case, that would mean keep charts as static and seeing if people really clamor for "give me those plotly charts interactions, i really need them!" and we can ask more about why it's so important.

davidkircos (approved):

PR #922, opened 2023-12-07
Update AI Tab

davidkircos:
- [x] Use a larger context window
- [x] Update docs in prompt


jimniels (approved):
This makes sense to me 👍 

Idk where/ how you pulled the text version of the docs, but there's gotta be an easy way we can just have gitbook publish all our docs as plain text files at an endpoint and we just slurp those in — that way they stay up to date?

PR #928, opened 2023-12-08
/support URL changed to /contact

waveminded:
looks like we changed the support URL to /contact


davidkircos (approved):

jimniels:
If we change this on the website we need to change it in the app to because we have a "support" URL in the app that points here

PR #931, opened 2023-12-11
Mp rust workspace

ddimaria:
- [x] Place Rust crates in a workspace
- [x] Import quadratic-core into quadratic-multiplayer
- [x] PoC of load and update grid files
- [x] Update files in state for each Transaction message


quadratic-multiplayer/examples/data/v1_4_simple.grid
davidfig, line 1: Centralize these so we don't have multiple example files

quadratic-multiplayer/src/file.rs
```diff
@@ -0,0 +1,76 @@
...
+    operations: String,
+) -> Result<TransactionSummary> {
+    tracing::info!("Applying operations: {}", operations);
+
+    let operations: Vec<Operation> = serde_json::from_str(&operations)?;
+
+    apply_operations(grid, operations)
+}
+
+pub(crate) fn apply_operations(
```
davidfig, line 22: Maybe move this to one of the transaction files so local mp uses the same function.

davidfig (approved):

quadratic-multiplayer/src/file.rs
ddimaria: This is a temp structure until we move out of stringified transactions.  I made a corresponding function in quadratic-core.  Is that what you meant?

PR #938, opened 2023-12-13
Mp transaction queue

ddimaria:
- [x] Build out basic transaction queue
- [x] Add to the transaction queue for each transaction in ANY room
- [x] Accept UUIDs for incoming Transaction messages
- [x] Broadcast UUIDs for all outgoing Transaction responses
- [x] Broadcast current sequence number to all users in the room
- [x] Allow the client to retrieve missing transactions

### DIFF
--- a/quadratic-multiplayer/src/background_worker.rs
+++ b/quadratic-multiplayer/src/background_worker.rs
@@ -1,11 +1,11 @@
+use anyhow::{anyhow, Result};
 use std::{sync::Arc, time::Duration};
-
-use tokio::time;
+use tokio::{task::JoinHandle, time};
 use uuid::Uuid;
 
 use crate::{
     message::{broadcast, response::MessageResponse},
-    state::State,
+    state::{room::Room, State},
 };
 
 /// In a separate thread:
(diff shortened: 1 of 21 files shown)


quadratic-multiplayer/src/background_worker.rs
```diff
@@ -0,0 +1,141 @@
...
+///   * Broadcast sequence number to all users in the room
+///   * Check for stale users in rooms and remove them.
+pub(crate) async fn work(state: Arc<State>, heartbeat_check_s: i64, heartbeat_timeout_s: i64) {
+    let state = Arc::clone(&state);
+
+    tokio::spawn(async move {
+        let mut interval = time::interval(Duration::from_millis(heartbeat_check_s as u64 * 1000));
+
+        loop {
+            // let rooms = state.rooms.lock().await.clone();
```
davidfig, line 21: remove

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -0,0 +1,141 @@
...
+    let (num_removed, num_remaining) = state
+        .remove_stale_users_in_room(file_id.to_owned(), heartbeat_timeout_s)
+        .await?;
+
+    tracing::info!("Checking heartbeats in room {file_id} ({num_remaining} remaining in room)");
+
+    if num_removed > 0 {
+        return Ok(broadcast(
+            // TODO(ddimaria): use a real session_id here
+            Uuid::new_v4(),
```
davidfig, line 84: maybe make an Option since we may want to send messages to all users sometimes.

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -102,44 +102,63 @@ pub(crate) async fn handle_message(
...
+            );
 
             let response = MessageResponse::Transaction {
+                id,
                 file_id,
                 operations,
+                sequence_num,
             };
 
             broadcast(session_id, file_id, Arc::clone(&state), response);
```
davidfig, line 127: won't this not send this message to the person who sent the transaction?

davidfig: the session_id should be None

quadratic-multiplayer/src/state/transaction_queue.rs
```diff
@@ -0,0 +1,141 @@
...
+            .queue
+            .get(&file_id)
+            .ok_or_else(|| anyhow!("file_id {file_id} not found in transaction queue"))?
+            .1
+            .to_owned();
+
+        Ok(transactions)
+    }
+
+    pub(crate) fn get_transactions_min_sequence_num(
```
davidfig, line 63: I don't think we want the min_sequence transaction. That's the last one the user has. Should be >, not >=

davidfig (approved):

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -0,0 +1,141 @@
...
+
+use crate::{
+    message::{broadcast, response::MessageResponse},
+    state::{room::Room, State},
+};
+
+/// In a separate thread:
+///   * Broadcast sequence number to all users in the room
+///   * Check for stale users in rooms and remove them.
+pub(crate) async fn work(state: Arc<State>, heartbeat_check_s: i64, heartbeat_timeout_s: i64) {
```
davidkircos, line 15: More descriptive function name?

quadratic-multiplayer/src/background_worker.rs
davidkircos, line 22: remove?

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -0,0 +1,141 @@
...
+                    file_id,
+                    room,
+                    heartbeat_timeout_s,
+                )
+                .await
+                {
+                    tracing::warn!(
+                        "Error removing stale users from room {file_id}: {:?}",
+                        error
+                    );
```
davidkircos, line 44: What will we do with these errors? Log service / sentry?

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -0,0 +1,141 @@
...
+) -> Result<JoinHandle<()>> {
+    let (num_removed, num_remaining) = state
+        .remove_stale_users_in_room(file_id.to_owned(), heartbeat_timeout_s)
+        .await?;
+
+    tracing::info!("Checking heartbeats in room {file_id} ({num_remaining} remaining in room)");
+
+    if num_removed > 0 {
+        return Ok(broadcast(
+            // TODO(ddimaria): use a real session_id here
```
davidkircos, line 85: still todo?

quadratic-multiplayer/src/background_worker.rs
ddimaria: Thanks for catching this!  Removed.

quadratic-multiplayer/src/message/handle.rs
ddimaria: Good point, I guess we want this now.  I need to update that param to be a vec anyhow, so will make than change now.

quadratic-multiplayer/src/state/transaction_queue.rs
ddimaria: The client should be asking for transactions starting with a given sequence number.  So, if the user's last sequence number is 5, they should ask for all transactions starting at 6.

quadratic-multiplayer/src/message/response.rs
```diff
@@ -15,15 +16,23 @@ pub(crate) enum MessageResponse {
...
     },
+    UserUpdate {
+        session_id: Uuid,
+        file_id: Uuid,
+        update: UserStateUpdate,
+    },
     Transaction {
+        id: Uuid,
         file_id: Uuid,
         // todo: this is a stringified Vec<Operation>. Eventually, Operation should be a shared type.
```
davidkircos, line 27: for all systems that are not actually reading ops data, we should just leave it in JOSN (eventually binary) format.

quadratic-multiplayer/src/state/transaction_queue.rs
davidfig: Ah. That makes sense. That's why you used min_. I was thinking the user would send their last transaction number. Both ways work.

davidkircos: Maybe this is a clearer name, `get_transactions_from_sequency_number`?

quadratic-multiplayer/src/background_worker.rs
ddimaria: This is being converted to a vec to support mixed use cases: 
```rust
pub(crate) fn broadcast(
    exclude: Vec<Uuid>,
```

ddimaria: Implemented

quadratic-multiplayer/src/background_worker.rs
ddimaria: Renamed

quadratic-multiplayer/src/background_worker.rs
ddimaria: This has been resolved by the recent broadcast updates.  TODO removed.

quadratic-multiplayer/src/message/response.rs
ddimaria: Agreed, we'll need to implement bincode encoding for transactions sent to the queue.

quadratic-multiplayer/src/message/handle.rs
ddimaria: The user who submitted the transaction now receives this message.

PR #941, opened 2023-12-14
Move FileCheckpoint Store to S3 and Remove from API

davidkircos:
- [x] Create a new model for `FileCheckpoint`
- [x] Return link to S3 instead of file contents from Express API
- [x] Refactor how example files are created
- [x] Create file works
- [x] Duplicate file works
- [ ] Script to move files from Postgres to S3 when we launch

Future
- [ ] Script to thin out checkpoints on S3
- [ ] API endpoint to return all checkpoints
- [ ] Client that allows a user to revert to a checkpoint (this is complicated)


quadratic-client/src/api/apiClient.ts
```diff
@@ -93,7 +93,10 @@ export const apiClient = {
 
   async downloadFile(uuid: string) {
     mixpanel.track('[Files].downloadFile', { id: uuid });
-    return this.getFile(uuid).then((json) => downloadQuadraticFile(json.file.name, json.file.contents));
+    const { file } = await this.getFile(uuid);
+    const checkpointUrl = file.lastCheckpointDataUrl;
+    const checkpointData = await fetch(checkpointUrl).then((res) => res.text());
```
jimniels, line 98: This seems like a pretty common thing we're going to be doing. I almost wonder if we should create something in the api client for this, like `getFileData` that does this all for you, e.g.

```js
const apiClient = {
  async getFile(uuid: string) => {...},
  async getFileData(uuid: string) => {
    const { file: { lastCheckpointDataUrl } } = await this.getFile(uuid);
    const fileData = await fetch(lastCheckpointDataUrl).then(res => res.text());
    return fileData;
  }
}
```

PR #944, opened 2023-12-14
Mp transaction dumb ops

davidfig:
- [ ] ensure all operations are "dumb"
   - [x] cell values
   - [x] formatting
   - [x] clipboard
   - [x] autocomplete
   - [x] borders
   - [x] import
   - [ ] update code cells
   - [ ] spills
   - [ ] sheets
   - [ ] offsets
- [ ] add spill changes as an Operation
- [x] clean up Controller directory structure: 
        * user_actions - functions called by wasm_bindgen to change the grid via a user's action
        * operations - functions to create operations (but not execute them)
        * execution - execute operations


quadratic-core/src/controller/execution/execute_operation.rs
```diff
@@ -1,13 +1,17 @@
+use crate::grid::formatting::CellFmtArray;
 use crate::{grid::*, Array, CellValue, SheetPos};
 
-use super::{formatting::CellFmtArray, operation::Operation, GridController};
+use super::super::operations::operation::Operation;
```
ddimaria, line 4: Nit, but consider `crate::` over multiple super sections.

quadratic-core/src/controller/execution/mod.rs
```diff
@@ -1,6 +1,9 @@
 pub mod compute;
 pub mod control_transaction;
 pub mod eval_formula;
+/// This module handles the application of operations to the Grid.
```
ddimaria, line 4: Move to the top?

quadratic-core/src/controller/operations/autocomplete.rs
```diff
@@ -0,0 +1,649 @@
+use anyhow::{anyhow, Result};
```
ddimaria, line 1: Odd this file shows as completely new (no diff).  I've been able to avoid this in the past by making file moves a separate commit so git picks up the rename/move.  You may be able to create a proper diff by moving this back to it's original location and moving the files in a separate commit.

ddimaria:
Test coverage went down slightly

ddimaria (approved):

PR #949, opened 2023-12-15
feat: update gtag analytics

jimniels:
Per a request from David:

> Hey we want to push GTag email so google Ads knows when someone is new. We just need this code snippet...And we have to fill out the email with Auth0 data

```
dataLayer.push({
 'event': 'Pageview',
  'userData': {
    'email': '{{ email }}'
  }
}); 
```

### DIFF
--- a/quadratic-client/src/utils/analytics.ts
+++ b/quadratic-client/src/utils/analytics.ts
@@ -12,16 +12,17 @@ type Options = {
 
 // This runs in the root loader, so analytics calls can run inside loaders.
 export function initializeAnalytics({ isAuthenticated, user }: Options) {
-  loadGoogleAnalytics();
+  loadGoogleAnalytics(user);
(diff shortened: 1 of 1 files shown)


quadratic-client/src/utils/analytics.ts
```diff
@@ -31,7 +32,18 @@ function loadGoogleAnalytics() {
   const script_2 = document.createElement('script');
   script_2.innerText = `
         window.dataLayer = window.dataLayer || [];
-        function gtag(){dataLayer.push(arguments);}
+        function gtag(){
+          dataLayer.push(arguments);
```
jimniels, line 36: This seemed like the right place to call this code? Right after the existing `dataLayer.push(arguments)` code.

If checks first if there's an email, because there might not be one, and only calls the `.push` if there is one. Does that sound right?

davidkircos (approved):

PR #955, opened 2023-12-27
Mp file service + Mp dumb operations

davidfig:
Combination of mp-naive-file-service with mp-dumb-ops + ActiveTransaction refactor

- [x] separate out transactions so we can have separate transactions running (this enables the next two todos)
- [x] push async transactions to stack to allow multiplayer operations to be applied while waiting for async to complete
- [x] need to rollback async transactions as well
- [x] add tests for adding multiplayer transactions while a user async transaction is pending


quadratic-multiplayer/src/message/handle.rs
```diff
@@ -120,30 +142,51 @@ pub(crate) async fn handle_message(
...
-            broadcast(vec![], file_id, Arc::clone(&state), response);
+            // I believe room_sequence_num is always correct. Let's broadcast after unpacking but before we push it to the transaction_queue.
+            // that way we don't need to wait for the transaction_queue to update the sequence_num
+            // TODO[ddimaria]: Please check my reworking of the order of operations below.
+
+            // unpack the operations or return an error
+            let operations_unpacked: Vec<Operation> = serde_json::from_str(&operations)?;
+
+            // lock the room
+            if let Ok(mut room) = get_mut_room!(state, file_id) {
```
ddimaria, line 159: You can remove the conditional and explicit error return with just adding the question mark operator:

```rust
let mut room = get_mut_room!(state, file_id)?;
```

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -120,30 +142,51 @@ pub(crate) async fn handle_message(
...
+                // add the transaction to the transaction queue
+                let calculated_sequence_num = state.transaction_queue.lock().await.push_pending(
+                    id,
+                    file_id,
+                    operations_unpacked,
+                    sequence_num,
+                );
+
+                // this should never happen if my logic above is correct
+                assert_eq!(
```
ddimaria, line 182: Can you change this to return an error rather than panic?  e.g.

```rust
if calculated_sequence_num > sequence_num {
    return Err(MpError::Something("Something"));
}
```

quadratic-multiplayer/src/message/handle.rs
davidfig: Makes sense. I was originally thinking we'd want to do something with the error, like send it back to the client. But not sure what to do with that.

infra/index.ts
```diff
@@ -4,7 +4,7 @@
 import {
   multiplayerInstanceDns,
   multiplayerPublicDns,
-} from "./multiplayer/multiplayer";
+} from "./multiplayer/multiplayer-service";
```
ddimaria, line 7: @davidkircos will this rename affect anything?

quadratic-api/src/app.ts
```diff
@@ -70,7 +71,7 @@ app.use('/v0/internal', internal_router);
 
 if (SENTRY_DSN) {
   // test route
-  app.get('/debug-sentry', function mainHandler(req, res) {
+  app.get('/debug-sentry', function mainHandler(/*req, res*/) {
```
ddimaria, line 74: Is this commenting out intentional?

quadratic-client/src/grid/controller/Grid.ts
```diff
@@ -69,14 +69,15 @@ export const rectToPoint = (rect: Rect): Point => {
...
+  grid: GridFile,
+  lastSequenceNum: number
 ): Promise<{
   contents: string;
   version: string;
 } | null> => {
   await init();
   try {
-    const gc = GridController.newFromFile(JSON.stringify(grid));
+    const gc = GridController.newFromFile(JSON.stringify(grid), lastSequenceNum);
```
ddimaria, line 80: I'll prob understand this better later in the code, but not sure why we would need the sequence number here?

quadratic-client/src/grid/controller/Grid.ts
```diff
@@ -749,11 +750,36 @@ export class Grid {
...
+  receiveSequenceNum(sequenceNum: number) {
+    if (debugShowMultiplayer) console.log(`[Multiplayer] Server is at sequence_num ${sequenceNum}.`);
+    const summary = this.gridController.receiveSequenceNum(sequenceNum);
+    this.transactionResponse(summary);
+  }
+
+  receiveMultiplayerTransactions(transactions: string) {
+    if (debugShowMultiplayer) console.log('[Multiplayer] Received catch-up transactions.');
+    const summaryResponse = this.gridController.receiveMultiplayerTransactions(transactions);
+    if (summaryResponse.Ok) {
```
ddimaria, line 776: Duplicate pattern, consider refactoring if you end up adding more of the same.

quadratic-client/src/multiplayer/multiplayer.ts
```diff
@@ -407,23 +409,35 @@ export class Multiplayer {
...
+  // Receives a collection of transactions to catch us up based on our sequenceNum
   private receiveTransactions(data: ReceiveTransactions) {
-    console.log(data.transactions);
+    grid.receiveMultiplayerTransactions(data.transactions);
+  }
+
+  // Receives the current transaction number from the server when entering a room.
+  // Note: this may be different than the one provided by the api as there may be unsaved Transactions.
+  private receiveEnterRoom(data: ReceiveEnterRoom) {
+    if (data.file_id !== this.room) {
```
ddimaria, line 428: Duplicate code, it might be good to add a `canAccess(file_id)` function on `this` to remove the duplication.

quadratic-client/src/multiplayer/multiplayerTypes.ts
```diff
@@ -77,12 +91,12 @@ export interface SendGetTransactions {
...
   session_id: string;
   file_id: string;
-  min_sequence_num: number;
+  min_sequence_num: bigint;
 }
 
 export interface ReceiveTransactions {
   type: 'Transactions';
-  transactions: Transaction[];
+  transactions: string;
```
ddimaria, line 99: Why the string here?

quadratic-client/src/ui/components/FileProvider.tsx
```diff
@@ -6,12 +6,9 @@ import { isOwner as isOwnerTest } from '../../actions';
 import { apiClient } from '../../api/apiClient';
 import { editorInteractionStateAtom } from '../../atoms/editorInteractionStateAtom';
 import { useFileRouteLoaderData } from '../../dashboard/FileRoute';
-import { debugShowFileIO } from '../../debugFlags';
-import { grid } from '../../grid/controller/Grid';
-import { useInterval } from '../../hooks/useInterval';
 
-const syncInterval = 500;
-const syncErrorInterval = 1000;
+// const syncInterval = 500;
```
ddimaria, line 10: Remove?

quadratic-client/src/ui/components/FileProvider.tsx
```diff
@@ -37,7 +34,7 @@ export const FileProvider = ({ children }: { children: React.ReactElement }) =>
   const { uuid } = useParams() as { uuid: string };
   const initialFileData = useFileRouteLoaderData();
   const [name, setName] = useState<FileContextType['name']>(initialFileData.name);
-  const [publicLinkAccess] = useState(initialFileData.sharing.public_link_access);
+  // const [publicLinkAccess] = useState(initialFileData.sharing.public_link_access);
```
ddimaria, line 37: Remove?

quadratic-core/src/controller/active_transactions/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+use uuid::Uuid;
+pub mod pending_transaction;
+
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct ActiveTransactions {
+    // async user transactions that are awaiting a response.
+    async_transactions: Vec<PendingTransaction>,
+
+    // Completed and async user Transactions that do not yet have a sequence number from the server.
+    // Vec<(forward_transaction, reverse_transaction)>
```
ddimaria, line 21: Nice documentation in this file 💯

quadratic-core/src/controller/active_transactions/pending_transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+            cells_accessed: HashSet::new(),
+            current_sheet_pos: None,
+            waiting_for_async: None,
+            complete: false,
+        }
+    }
+}
+
+impl PendingTransaction {
+    pub fn to_transaction(&self, sequence_num: Option<u64>) -> Transaction {
```
ddimaria, line 79: It might be more idiomatic to `impl From<>` or `Into` for this type of conversion

quadratic-core/src/controller/active_transactions/pending_transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+        Transaction {
+            id: self.id,
+            sequence_num,
+            operations: self.operations.clone().into(),
```
ddimaria, line 83: This `.into()` might not be needed.

quadratic-core/src/controller/execution/control_transaction.rs
```diff
@@ -1,173 +1,129 @@
...
-    fn transact(&mut self, operations: Vec<Operation>, compute: bool) {
-        for op in operations.iter() {
-            if cfg!(feature = "show-operations") {
-                crate::util::dbgjs(&format!("[Operation] {:?}", op.to_string()));
-            }
+            self.execute_operation(transaction);
 
-            self.execute_operation(op.clone(), compute);
+            if transaction.waiting_for_async.is_some() {
+                self.transactions.add_async_transaction(transaction.clone());
```
ddimaria, line 31: You may be able to avoid this clone by passing it by reference.

quadratic-core/src/controller/execution/control_transaction.rs
```diff
@@ -1,173 +1,129 @@
...
-                self.undo_stack.push(self.to_transaction());
+                let undo = transaction.to_undo_transaction();
+                self.undo_stack.push(undo.clone());
+                self.transactions
+                    .unsaved_transactions
+                    .push((transaction.to_forward_transaction(), undo));
             }
             TransactionType::Multiplayer => (),
+            TransactionType::MultiplayerKeepSummary => (),
             TransactionType::Unset => panic!("Expected a transaction type"),
```
ddimaria, line 75: Is this TransactionType still needed?

quadratic-core/src/controller/execution/execute_operation/execute_set_borders.rs
```diff
@@ -0,0 +1,43 @@
+use crate::controller::{
+    active_transactions::pending_transaction::PendingTransaction, operations::operation::Operation,
+    GridController,
+};
+
+impl GridController {
+    pub fn execute_set_borders(&mut self, transaction: &mut PendingTransaction, op: &Operation) {
+        match op.clone() {
```
ddimaria, line 8: I've seen this pattern in a few places.  You can remove the clone and just update the calls below: 
```rust
// should be removed
transaction.forward_operations.push(Operation::SetBorders {
    sheet_rect: *sheet_rect,
    borders: *borders,
});
transaction.reverse_operations.insert(
    0,
    Operation::SetBorders {
        sheet_rect: *sheet_rect,
        borders: old_borders,
    },
);
```

quadratic-core/src/controller/execution/execute_operation/execute_set_cell_values.rs
```diff
@@ -0,0 +1,215 @@
...
+    Array, CellValue,
+};
+
+impl GridController {
+    pub(crate) fn execute_set_cell_values(
+        &mut self,
+        transaction: &mut PendingTransaction,
+        op: &Operation,
+    ) {
+        match op {
```
ddimaria, line 15: For all of these single matches, you can just use `if let Operation::SetCellValues { sheet_rect, values } = op {}` syntax to remove the clone and avoid the unnecessary unreachable match.

quadratic-core/src/controller/execution/mod.rs
```diff
@@ -19,43 +16,32 @@ pub enum TransactionType {
...
-            self.summary.forward_operations = Some(
-                serde_json::to_string(&self.forward_operations)
-                    .expect("Failed to serialize forward operations"),
-            );
-        }
-        let summary = self.summary.clone();
-        self.summary.clear(self.complete);
-        summary
+    /// Sets the last_sequence_num for multiplayer. This should only be called when receiving the sequence_num.
+    pub fn set_last_sequence_num(&mut self, last_sequence_num: u64) {
```
ddimaria, line 38: Why does the grid controller need to know about the last sequence number?

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -0,0 +1,846 @@
...
+    pub fn receive_sequence_num(&mut self, sequence_num: u64) -> TransactionSummary {
+        let mut summary = TransactionSummary::default();
+        if sequence_num != self.transactions.last_sequence_num {
+            let now = Utc::now();
+            if match self.transactions.last_get_transactions_time {
+                None => true,
+                Some(last_request_transaction_time) => {
+                    last_request_transaction_time
+                        .checked_add_signed(Duration::seconds(SECONDS_TO_WAIT_FOR_GET_TRANSACTIONS))
+                        .unwrap()
```
ddimaria, line 114: unsafe unwrap

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -0,0 +1,846 @@
...
+        if self.transactions.out_of_order_transactions.is_empty() {
+            self.transactions.last_sequence_num = sequence_num;
+            return;
+        }
+
+        // combines all out of order transactions into a single vec of operations
+        let mut operations = VecDeque::new();
+        self.transactions.out_of_order_transactions.retain(|t| {
+            // while the out of order transaction is next in sequence, we apply it and remove it from the list
+            if t.sequence_num.unwrap() == sequence_num + 1 {
```
ddimaria, line 144: unsafe unwrap

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -0,0 +1,846 @@
...
+    }
+
+    /// Check the out_of_order_transactions to see if they are next in order. If so, we remove them from
+    /// out_of_order_transactions and apply their operations.
+    fn apply_out_of_order_transactions(
+        &mut self,
+        transaction: &mut PendingTransaction,
+        sequence_num: u64,
+    ) {
+        let mut sequence_num = sequence_num;
```
ddimaria, line 132: I think you can avoid this by just adding `mut` to the param.

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -0,0 +1,846 @@
...
+                self.transactions.last_get_transactions_time = None;
+                summary.request_transactions = Some(self.transactions.last_sequence_num + 1);
+            }
+        }
+        summary
+    }
+
+    /// Check the out_of_order_transactions to see if they are next in order. If so, we remove them from
+    /// out_of_order_transactions and apply their operations.
+    fn apply_out_of_order_transactions(
```
ddimaria, line 127: I now see why there is a sequence number used, but on the fence about this being in core 🤔

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -0,0 +1,846 @@
...
+                .extend(transaction.sheets_with_dirty_bounds);
+        });
+        self.reapply_unsaved_transactions(&mut results);
+        self.finalize_transaction(&mut results);
+        Ok(results.prepare_summary(false))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ddimaria, line 254: Great tests 💯

quadratic-core/src/controller/execution/run_code/get_cells.rs
```diff
@@ -0,0 +1,71 @@
...
+        let sheet_name = get_cells.sheet_name();
+
+        // if sheet_name is None, use the sheet_id from the pos
+        let sheet = sheet_name.clone().map_or_else(
+            || Some(self.grid().sheet_from_id(current_sheet)),
+            |sheet_name| self.grid().sheet_from_name(sheet_name),
+        );
+
+        let transaction_type = transaction.transaction_type.clone();
+        assert_eq!(transaction_type, TransactionType::User);
```
ddimaria, line 35: This will panic, leaving things in a bad state.  You can just wrap the code below in a conditional, or just return early if it's not a `User` `TransactionType`.

quadratic-core/src/controller/execution/run_code/run_python.rs
```diff
@@ -0,0 +1,574 @@
...
+        // stop the computation cycle until async returns
+        transaction.current_sheet_pos = Some(sheet_pos);
+        transaction.waiting_for_async = Some(CodeCellLanguage::Python);
+        transaction.has_async = true;
+        true
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ddimaria, line 69: Great tests in this and in the previous file 🙌

quadratic-core/src/core_error.rs
```diff
@@ -0,0 +1,38 @@
+//! quadratic-core error handling
```
ddimaria, line 1: Nice addition 🎉

quadratic-core/src/grid/file/v1_5/schema.rs
```diff
@@ -10,6 +11,20 @@ pub struct GridSchema {
...
+
+impl From<SheetId> for Id {
+    fn from(id: SheetId) -> Self {
+        Self { id: id.to_string() }
+    }
+}
+
+impl From<Id> for SheetId {
+    fn from(id: Id) -> Self {
+        SheetId::from_str(&id.id).unwrap()
```
ddimaria, line 24: Unsafe unwrap

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -37,8 +37,9 @@ pub(crate) async fn start(state: Arc<State>, heartbeat_check_s: i64, heartbeat_t
                 let state = Arc::clone(&state);
 
                 tokio::spawn(async move {
-                    tracing::info!("Processing room {}", file_id);
+                    // tracing::info!("Processing room {}", file_id);
```
ddimaria, line 40: Can you remove this comment or convert to a `tracing::trace!`?

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -115,12 +116,17 @@ async fn remove_stale_users_in_room(
         .remove_stale_users_in_room(file_id.to_owned(), heartbeat_timeout_s)
         .await?;
 
-    tracing::info!("Checking heartbeats in room {file_id} ({num_remaining} remaining in room)");
+    // tracing::info!("Checking heartbeats in room {file_id} ({num_remaining} remaining in room)");
```
ddimaria, line 119: Can you remove this comment or convert to a tracing::trace!?

quadratic-multiplayer/src/background_worker.rs
```diff
@@ -139,7 +145,7 @@ mod tests {
     use super::*;
 
     #[tokio::test]
-    async fn broadcast_sequence_num() {
+    async fn test_broadcast_sequence_num() {
```
ddimaria, line 148: This is a test, it's a little redundant prepend with test?

quadratic-multiplayer/src/error.rs
```diff
@@ -57,6 +57,12 @@ pub(crate) enum MpError {
 
     #[error("User error: {0}")]
     User(String),
+
+    #[error("User does not exist: {0}")]
+    UserDoesNotExist(String),
+
+    #[error("Sequence number mismatch")]
+    SequenceNumberMismatch,
```
ddimaria, line 65: Can you alpha sort this?

quadratic-multiplayer/src/file.rs
```diff
@@ -137,14 +131,14 @@ pub(crate) async fn process_queue_for_room(
...
-    // update the checkpoint in quatratic-api
+    // update the checkpoint in quadratic-api
     let key = &key(*file_id, last_sequence_num);
     set_file_checkpoint(
         base_url,
         jwt,
         file_id,
         last_sequence_num,
-        "1.4".into(),
+        CURRENT_VERSION.into(),
```
ddimaria, line 141: 🙌

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -83,12 +101,32 @@ pub(crate) async fn handle_message(
...
+                session_id,
+                file_id,
+                Arc::clone(&state),
+                MessageResponse::EnterRoom {
+                    file_id,
+                    sequence_num,
+                },
+            )
+            .await
+            .map_err(|e| MpError::SendingMessage(e.to_string()))?;
```
ddimaria, line 115: Can you just use the ? operator here rather than map the error?

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -83,12 +101,32 @@ pub(crate) async fn handle_message(
...
             // only broadcast if the user is new to the room
             if is_new {
                 let room = state.get_room(&file_id).await?;
                 let response = MessageResponse::from(room.to_owned());
 
-                broadcast(vec![], file_id, Arc::clone(&state), response);
+                broadcast(vec![], file_id, Arc::clone(&state), response)
+                    .await
+                    .map_err(|e| {
+                        MpError::SendingMessage(format!(
```
ddimaria, line 125: Can you just use the ? operator here rather than map the error?

quadratic-multiplayer/src/message/mod.rs
```diff
@@ -49,3 +49,35 @@ pub(crate) fn broadcast(
         }
     })
 }
+
+/// Send a message to a specific user in a room.
+/// All messages are sent in a separate thread.
+#[tracing::instrument(level = "trace")]
+pub(crate) fn direct_message(
```
ddimaria, line 56: Can you rename this to `send_user_message()` and then call this function from within `broadcast()` and remove the duplicated code in `broadcast()`?

quadratic-multiplayer/src/message/response.rs
```diff
@@ -26,12 +24,15 @@ pub(crate) enum MessageResponse {
...
         id: Uuid,
         file_id: Uuid,
-        // todo: this is a stringified Vec<Operation>. Eventually, Operation should be a shared type.
-        operations: String,
         sequence_num: u64,
+        operations: String,
     },
     Transactions {
-        transactions: Vec<Transaction>,
+        transactions: String,
```
ddimaria, line 31: Why revert to strings?

quadratic-multiplayer/src/server.rs
```diff
@@ -337,14 +341,15 @@ pub(crate) mod tests {
...
 
-        let response = integration_test_send_and_receive(socket, request, true).await;
+        let response = integration_test_send_and_receive(&socket, request, true).await;
 
         assert_eq!(response, Some(serde_json::to_string(&expected).unwrap()));
     }
 
     #[tokio::test]
-    async fn user_enters_a_room() {
+    async fn test_user_enters_a_room() {
```
ddimaria, line 350: Can you remove the `test_` prepend here?

quadratic-multiplayer/src/state/room.rs
```diff
@@ -80,6 +85,7 @@ impl State {
         );
 
         if num_in_room == 0 {
+            tracing::info!("Empty room.");
```
ddimaria, line 88: Can you add the file_id to this message?

quadratic-multiplayer/src/test_util.rs
```diff
@@ -135,3 +135,14 @@ pub(crate) async fn integration_test_send_and_receive(
 
     Some(response)
 }
+
+pub(crate) async fn integration_test_receive(
```
ddimaria, line 139: This is duplicated code.  You can call this function in the previous function and remove the duplications in lines 131 - 134.

infra/index.ts
davidfig: I renamed it because I kept opening the wrong file when opening multiplayer.ts. It seemed to be clean, though.

quadratic-client/src/grid/controller/Grid.ts
davidfig: The plan is to more elegantly handle these errors. Just a placeholder for now.

quadratic-client/src/multiplayer/multiplayer.ts
davidfig: This is not really a canAccess issue; more like a sanity check to ensure the socket server sent the expected file_id (ie, we're in the correct room). This should not happen and most messages don't send the file_id.

quadratic-client/src/multiplayer/multiplayerTypes.ts
davidfig: The types get messed up if we allow TS to parse the objects (I think). So I just keep it stringified until it's sent to the server.

quadratic-client/src/ui/components/FileProvider.tsx
davidfig: These are all the local file save deletion.

quadratic-core/src/controller/active_transactions/pending_transaction.rs
davidfig: I thought you can't do that if you pass an additional parameter like sequence_num?

quadratic-core/src/controller/active_transactions/pending_transaction.rs
davidfig: Into is needed since it goes from Vec to VecDeque

quadratic-core/src/controller/execution/control_transaction.rs
davidfig: It does not. I meant to delete it.

quadratic-core/src/controller/execution/execute_operation/execute_set_borders.rs
davidfig: I tried, but it's more difficult than just cloning on top since it's used in a bunch of places throughout that functions.

quadratic-core/src/controller/execution/execute_operation/execute_set_cell_values.rs
davidfig: Awesome. I was looking for a better pattern. I'll clean all these files up.

quadratic-core/src/controller/execution/execute_operation/execute_set_borders.rs
davidfig: Nevermind. I found a much better way of just consuming op.

quadratic-core/src/controller/execution/mod.rs
davidfig: ActiveTransactions tracks the sequence_num to properly request updates to transactions. This is the logic that is used within `receive_multiplayer.rs`. Happy to talk you through how it works in a call.

quadratic-api/src/app.ts
ddimaria: It might be best to remove the commented out code or add a comment above it regarding the reason.

quadratic-core/src/controller/execution/receive_multiplayer.rs
davidfig: I'm not sure how else to do this. There's a bunch of logic in core that handles properly ordering transactions and requesting additional transactions. We can't do the proper ordering without knowing what the current sequence_num is. This conversation is better in a call. It's complimicated.

quadratic-client/src/grid/controller/Grid.ts
ddimaria: I should have added to this yesterday.  I'm wondering if managing sequence numbers is the domain of quadratic core?  It seems to be a concern outside of it.

quadratic-core/src/core_error.rs
davidfig: I saw you do it and thought it was awesome.

quadratic-multiplayer/src/background_worker.rs
davidfig: useful for debugging down the road. just created lots of noise

quadratic-multiplayer/src/background_worker.rs
davidfig: I like having the test_ prepend to search for them more easily. I just search test_br and it pops up. Otherwise, I get all the calls to broadcast_sequence_num.

quadratic-multiplayer/src/error.rs
davidfig: heh. didn't realize it was sorted

quadratic-multiplayer/src/message/handle.rs
davidfig: It's possible, but then we'd have to map all the JoinErrors to MpErrors. Not sure that properly handles all the cases.

quadratic-multiplayer/src/message/handle.rs
davidfig: Same as above. We'd have to map the JoinErrors to MpErrors. I guess I'm lazy.

quadratic-multiplayer/src/message/response.rs
davidfig: Becuase of TS type conversion issues (I think?)

quadratic-multiplayer/src/message/mod.rs
davidfig: Renamed. Broadcast steps through all users. Not sure there's a way of combining the two without searching the user list twice.

quadratic-client/src/multiplayer/multiplayer.ts
ddimaria: The canAccess naming is less important than the duplication.  Adding an abstraction here not only makes the code more DRY but also ads explicitness.

quadratic-core/src/controller/active_transactions/pending_transaction.rs
ddimaria: Good point, I overlooked that in the signature, please ignore this.

quadratic-core/src/controller/execution/receive_multiplayer.rs
ddimaria: Yea, let's discuss.  We can't add this type of code to the shared crate since it has some dependences on core, but still seems like it should be outside of core.

quadratic-multiplayer/src/message/handle.rs
ddimaria: Oh, on these, just remove the await and the map_err, these run in a separate thread and we intentionally don't want to block on them.  Same comment for the one below.

quadratic-multiplayer/src/message/mod.rs
ddimaria: Looking back, what I wrote was confusing (not the renaming, but the reuse).  Both functions ultimately send a message to a single user, so breaking out that piece into a separate function and having both functions invoke it will reduce duplication.

ddimaria:
It looks like coverage didn't happen 🤔

PR #968, opened 2024-01-06
Add basic support for `.show()`

Asaaj:


quadratic-client/public/run_python.py
```diff
@@ -147,6 +148,82 @@ def ensure_not_cell(item):
...
+
+    plotly.io._base_renderers.open_html_in_browser = prev_open_html
+
+class _ResultHolder:
+    def __init__(self):
+        self._result = None
+        self._result_set_from_line = None
+
+    def set_result(self, figure):
+        user_call_frame = _get_frame_calling_into_file("/site-packages/plotly/")
```
Asaaj, line 174: This is a bit hacky to be sure, but since we're patching a lower-level function in plotly (mostly to ensure users who want to show plots in a different way can't bypass it), it's not easy to show useful information about where the user already called `.show()`.

I think it's a valuable error message, but I'm certainly unhappy with this implementation

Asaaj:
The Python version in Quadratic right now is 3.113, but CI is running 3.9 (thus the failure). Which are we supporting?

davidkircos:
> The Python version in Quadratic right now is 3.113, but CI is running 3.9 (thus the failure). Which are we supporting?

We should support Python 3.11.3 (the version in the app). I think you can change it in CI here https://github.com/quadratichq/quadratic/blob/support-plotly-show/.github/workflows/ci.yml#L85

quadratic-client/public/run_python.py
```diff
@@ -147,6 +149,106 @@ def ensure_not_cell(item):
         return [ensure_not_cell(x) for x in item]
     return not_cell(item)
 
+def _get_user_frames() -> list[inspect.FrameInfo] | None:
```
Asaaj, line 152: Tests are still needed, especially for this function

Asaaj:
Examples of the error messages:

image

image

The second one will be changed, I just need to check whether people think it's safe to `rstrip` the code of newlines and then just count the number of lines to get the return line. That seems okay to me since (I believe) there can only be one valid return line per script, but I just want to make sure

Asaaj:
Possible wording for example 2: "you must call .show() on the last line of code"

davidkircos:
> > image
> 
> @Asaaj, what do you think about changing the error message to "Plot `.show()` must be on this cell's last line of code. Cells can either return data OR a plot, not both."

Reading it again. I think Jacob's OG message is clear.

Asaaj:
> Reading it again. I think Jacob's OG message is clear.

I am happy to do either one, up to you. I think your proposed wording is slightly clearer, but it does impose a false limitation (though I don't know who would care about that)

quadratic-client/public/run_python.py
```diff
@@ -147,6 +149,111 @@ def ensure_not_cell(item):
...
+        return None
+
+def _line_number_from_traceback() -> int:
+    cl, exc, tb = sys.exc_info()
+    line_number = traceback.extract_tb(tb)[-1].lineno
+    return line_number
+
+def _get_return_line(code: str) -> int:
+    code = code.rstrip()
+    return code.count("\n") + 1
```
Asaaj, line 177: Python supports several line separators in `str.splitlines`, but that's about 5-8x slower than counting `\n` (making it take about 15 seconds on my M3 for 1000 lines). 

Since Linux, Windows, and MacOS (since at least 2001) all have one `\n` in their newline, I assumed this would work fine.

[Full list of `splitlines` line separators](https://docs.python.org/3/library/stdtypes.html#str.splitlines)

quadratic-client/public/run_python.py
```diff
@@ -147,6 +149,111 @@ def ensure_not_cell(item):
...
+    def result(self):
+        return self._result
+
+    @property
+    def result_set_from_line(self) -> int | None:
+        return self._result_set_from_line
+
+async def _intercept_plotly_html(code) -> _FigureHolder | None:
+    if "plotly" not in code:
+        return None
```
Asaaj, line 238: Obviously this isn't perfect, but it should save most people from encountering an unnecessary `micropip.install`. Let me know if anyone has any other ideas for how to handle this

quadratic-client/public/run_python.py
```diff
@@ -147,6 +149,111 @@ def ensure_not_cell(item):
...
+            lambda frame: frame.filename == "<exec>", rev_frames
+        )
+
+        return list(reversed(list(rev_frames)))
+    except:
+        return None
+
+def _line_number_from_traceback() -> int:
+    cl, exc, tb = sys.exc_info()
+    line_number = traceback.extract_tb(tb)[-1].lineno
```
Asaaj, line 172: Note: I changed this line from `traceback.extract_tb(tb)[-1][1]` because it seems clearer. I assume the previous version predates the named fields or something

quadratic-client/public/run_python.py
davidkircos: Pyodide has a function for finding imports. We could use that.
https://pyodide.org/en/stable/usage/api/python-api/code.html#pyodide.code.find_imports

davidkircos (approved):
Left one comment.

@Asaaj up to you if you think it's worth doing. Otherwise ready to merge.

quadratic-client/public/run_python.py
Asaaj: Oh very nice, I'll make that change. Thanks!

PR #978, opened 2024-01-12
feat: new user association

jimniels:
When a new user comes in to the system:

- Create them in the User table of the database
- Find any outstanding invites that they have to a team or a file
- Make them members of the team or file
- Delete the old invites

Tested this locally by:

- Deleting my quadratichq.com user from my local database and then from auth0 (staging)
- Use my personal gmail.com to invite my quadartichq.com user to 1 file and 1 team
- Log in to the app using my quadrtichq.com user and see that immediately belong to 1 team and have 1 file in the "Share with me" tab
- Look in my local database and see that the old invites were removed and associations made to the team/file

### DIFF
--- a/quadratic-api/src/auth0/profile.ts
+++ b/quadratic-api/src/auth0/profile.ts
@@ -26,6 +26,7 @@ export const getAuth0Users = async (auth0Ids: string[]) => {
 
 /**
  * Given a list of users from our system, we lookup their info in Auth0.
+ * If we don't find a user, we throw.
  *
  * Example in:
  *   [
@@ -33,7 +34,7 @@ export const getAuth0Users = async (auth0Ids: string[]) => {
  *   ]
  * Example out:
  *   {
- *     'google-oauth2|112233': {
+ *     10: {
  *       id: 10,
  *       auth0Id: 'google-oauth2|112233',
  *       email: 'john_doe@example.com',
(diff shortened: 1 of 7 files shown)


quadratic-api/src/auth0/profile.ts
```diff
@@ -26,14 +26,15 @@ export const getAuth0Users = async (auth0Ids: string[]) => {
...
+ * If we don't find a user, we throw.
  *
  * Example in:
  *   [
  *     { id: 10, auth0Id: 'google-oauth2|112233', ... }
  *   ]
  * Example out:
  *   {
- *     'google-oauth2|112233': {
+ *     10: {
```
jimniels, line 37: I had a bad comment here, fixed this.

quadratic-api/src/middleware/user.ts
```diff
@@ -1,20 +1,76 @@
...
+
+  // If they don't exist yet, create them
+  const newUser = await dbClient.user.create({
+    data: {
       auth0Id,
     },
   });
 
-  return user;
+  // And now we have some extra work to do if it's their first time logging in
```
davidkircos, line 24: Move this all to a function for actions to take when we see a user in our system for the first time.

quadratic-api/src/middleware/user.ts
```diff
@@ -1,20 +1,76 @@
...
+  const { email } = usersById[newUser.id];
+
+  // See if they've been invited to any teams and make them team members
+  const teamInvites = await dbClient.teamInvite.findMany({
+    where: {
+      email,
+    },
+  });
+  if (teamInvites.length) {
+    await dbClient.teamInvite.deleteMany({
```
davidkircos, line 37: delete after create. Shouldn't matter here, but good practice.

PR #979, opened 2024-01-12
A team fix nits

davidfig:
- [x] Charts not displaying on first render
- [x] HTML charts no longer overlap the grid headings
- [x] file loading speed is way faster (I skip the upgrade check if file >= v1.4)
- [x] multiplayer cursors out of bounds works better
- [x] fix bug with code editor not updating to latest code after a code run
- [x] polish multiplayer user and avatar
- [x] add follow UI treatment  
- [x] copy portions of a code_run to the clipboard (if the code cell itself is copied, then only the code cell is included on the quadratic clipboard; otherwise the code cell's overlapping output values are included)
- [x] bug where paste doesn't select all pasted cells
- [x] bug where you cannot undo a code_cell paste
- [x] fix multiplayer cursor scale bug


quadratic-core/src/controller/active_transactions/unsaved_transactions.rs
```diff
@@ -64,10 +64,13 @@ impl UnsavedTransactions {
                 };
                 if !cfg!(test) && !cfg!(feature = "multiplayer") && send {
                     if let Ok(stringified) = serde_json::to_string(&transaction) {
-                        crate::wasm_bindings::js::addUnsentTransaction(
-                            transaction.forward.id.to_string(),
-                            stringified,
-                        );
+                        #[allow(unused_unsafe)]
+                        unsafe {
```
ddimaria, line 68: Did the compiler suggest this use of unsafe?  It might be best to isolate unsafe code to the `wasm_bindings` directory.

quadratic-core/src/controller/execution/execute_operation/execute_cursor.rs
```diff
@@ -0,0 +1,89 @@
+use crate::controller::{
+    active_transactions::pending_transaction::PendingTransaction, operations::operation::Operation,
+    GridController,
+};
+use serde::Serialize;
+
+#[derive(Serialize)]
```
ddimaria, line 7: I'm not familiar with the new code organization.  Should these structs be placed closer to the bindings since they only relate to binding-level messaging?

ddimaria (approved):

quadratic-core/src/controller/user_actions/clipboard.rs
```diff
@@ -70,6 +71,55 @@ impl GridController {
...
+            .iter_code_output_in_rect(clipboard_rect)
+            .filter(|(_, code_cell)| !code_cell.spill_error)
+            .for_each(|(output_rect, code_cell)| {
+                // only change the cells if the CellValue::Code is not in the selection box
+                let code_pos = Pos {
+                    x: output_rect.min.x,
+                    y: output_rect.min.y,
+                };
+                if !clipboard_rect.contains(code_pos) {
+                    let x_start = if output_rect.min.x > clipboard_rect.min.x {
```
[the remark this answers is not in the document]
davidfig: at one point my IDE was complaining about this. not anymore. weird.

Issue #53, opened 2022-04-20
Python Errors: Stack Trace is Wrong


davidkircos:
Line errors are not helpful when they come from other packages. Says a line number that isn’t in your code, without the full stack trace.

Example formula:
```python
import pandas as pd

a = pd.DataFrame()
a[0][0]

```

<img width="294" alt="Screen Shot 2022-04-20 at 3 08 44 PM" src="https://user-images.githubusercontent.com/3479421/164323348-655c4a0a-aa0d-4a0c-84f0-4c3aee5ca1e8.png">

davidkircos:
Upgrading Pyodide and using the CodeRunner should help

https://pyodide.org/en/stable/usage/api/python-api/code.html#pyodide.code.CodeRunner

ddimaria:
This is fixed in https://github.com/quadratichq/quadratic/tree/staging-2024-01

PR #982, opened 2024-01-14
Split python into modules

Asaaj:
- [x] Move Python code to a library
- [x] Build wheel and place in `quadratic-client/public`
- [x] Test Python code 
- [x] Add package.json + scripts for dev, build, and test 
- [x] Add a README.md 
- [x] Simple auto-complete for `cell()` and `cells()` 
- [x] Fixes https://github.com/quadratichq/quadratic/issues/53
- [x] Addresses https://github.com/quadratichq/quadratic/issues/627


quadratic-client/public/quadratic_py/plotly_patch.py
```diff
@@ -0,0 +1,68 @@
...
+    def result(self):
+        return self._result
+
+    @property
+    def result_set_from_line(self) -> int | None:
+        return self._result_set_from_line
+
+
+async def intercept_plotly_html(code) -> _FigureHolder | None:
+    import pyodide.code
```
Asaaj, line 48: **issue**: Needs to be moved after the imports check

quadratic-core/build.rs
Asaaj, line 1: `cargo` is certainly not the right place to do this, but it's what I found simplest. If someone else has a better way of generating the wheel every time a python file changes, please let me know

davidfig (approved):

PR #989, opened 2024-01-18
Add gallery examples to app 

luke-quadratic:
PR adds a bunch of the gallery examples to the app, makes descriptions of all files more succinct, and removes an unused file.


davidkircos (changes requested):
CleanShot 2024-01-18 at 10 55 45@2x
some examples don't work

PR #993, opened 2024-01-19
feat: support sending emails

jimniels:
- Creates an "internal" route at `/v0/internal/emails/:id` which allows _anyone_ to hit that API and get the template for an email. Response is JSON with the keys `subject` and `html` and placeholder values for the emails. This is meant to help with quickly testing our email templating system.
- Add a note to `a-team` that we need to add a `SENDGRID_API_KEY` for prod and other environments where we want to support sending emails


quadratic-api/src/email/sendEmail.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  // Don't log anything if we're testing
+  if (process.env.NODE_ENV === 'test') {
+    return;
+  }
+
+  // Don't try to send an email if we don't have the API key
+  if (!apiKey) {
+    console.log('Skip sending email');
```
davidkircos, line 25: "Development, console logging email:"

davidkircos (approved):

PR #996, opened 2024-01-20
helper: Press '=' to code

davidfig:
An empty grid will show the `Press '=' to enter code` message at the cursor.

Screenshot 2024-01-20 at 5 15 28 AM


jimniels:
I like the cell-level implementation.

But I feel like only ever showing this when the grid is empty is less than ideal.

Like if somebody enters the app for the first time, via a sheet somebody shared with them (or even the default sheet), they're going to see a grid with a bunch of data in it and they won't see this nice little hint.

Or if you type something in to a cell, but never actually use the `=` functionality, you also won't see the placeholder text.

I think ideally we would show this once we know the user has actually clicked `=` and used the cell type menu. Then we could drop something in localstorage that turns this off. 

This way, we _know_ people have used this feature and triggered that menu at least once before we hide it.

Also, FWIW: in the current implementation, if you type '=', choose python, then immediately close the code editor, the grid is empty but placeholder is no longer there

luke-quadratic:
reasonable to get this merged and if we want to do deeper thinking later we can? 

lowest hanging fruit of what the PR does will have meaningful increase in users that get to writing their first line of code

PR #998, opened 2024-01-21
Support Image cell value for plots

Asaaj:


quadratic-kernels/python-wasm/quadratic_py/run_python.py
```diff
@@ -145,6 +146,11 @@ async def run_python(code: str):
         except Exception:
             pass
 
+        bytes_output = None
+        if isinstance(output_value, bytes):
+            bytes_output = output_value
+            output_value = ''
```
Asaaj, line 152: This block should probably be removed before merging. It allows any user to just return a `bytes` object, which will then be put into an image type later if possible. Currently, it just always tries to show `bytes` objects as `IMAGE` cells, which is obviously bad.

image

quadratic-core/src/controller/execution/run_code/mod.rs
```diff
@@ -274,6 +280,9 @@ impl GridController {
                 } else {
                     Value::Single("".into())
                 }
+            } else if let Some(bytes_output) = js_code_result.bytes_output() {
+                // TODO(jrice): Parse and tag the output here with the proper image type
+                Value::Single(CellValue::Image(bytes_output))
```
Asaaj, line 285: I believe this is where we need to insert the parsing logic to validate that a `bytes` object returned from Python is indeed an image. This needs to be done before merging.

Asaaj:
Note: That PR does not contain all of the changes in this one, such as returning bytes from Python. That will be added back in a subsequent PR.

Issue #862, opened 2023-11-21
Paste not working in firefox


luke-quadratic:
#693 did not solve this issue. Paste still not working on firefox.

luke-quadratic:
works

PR #1001, opened 2024-01-22
fix firefox paste - issue-862

PawelJastrzebski:
https://github.com/quadratichq/quadratic/issues/862


quadratic-client/src/grid/actions/clipboard/clipboard.ts
```diff
@@ -69,6 +69,10 @@ export const pasteFromClipboardEvent = (e: ClipboardEvent) => {
   if (e.clipboardData.types.includes('text/plain')) {
     plainText = e.clipboardData.getData('text/plain');
   }
+  if (!html && !fullClipboardSupport()) { // try firefox storage
```
PawelJastrzebski, line 72: This fixes the issue

Issue #851, opened 2023-11-20
bug: getCells works where c fails


davidkircos:
![CleanShot 2023-11-20 at 13 53 49@2x](https://github.com/quadratichq/quadratic/assets/3479421/d5193ce5-8af7-4ae8-9a36-d0bf6cae72f9)
![CleanShot 2023-11-20 at 13 54 00@2x](https://github.com/quadratichq/quadratic/assets/3479421/42afea2a-cc55-4ae5-914b-5c6fb2cfc6e4)

PR #1002, opened 2024-01-22
fix issue 851

PawelJastrzebski:
https://github.com/quadratichq/quadratic/issues/851
updated python "c()" function to handle cell range


davidkircos:
fix in another branch

PR #1016, opened 2024-01-24
issue 325 - support text resize in code editor 

PawelJastrzebski:


jimniels (approved):
Updated this with the latest from `main` and added a shortcut for `Cmd` + `0` to reset font size. Otherwise, LGTM 👍

Issue #383, opened 2023-03-28
bug: arrow key navigation pans late when the code editor is open


jimniels:
When you move the cursor position with the arrow keys, the grid will re-render when the cursor moves off screen. 

However, when you have the code editor open, the cursor _passes under_ the code editor but doesn't pan the grid until it hits the edge of the screen.

![CleanShot 2023-03-28 at 14 44 44@2x](https://user-images.githubusercontent.com/1316441/228363033-9119c71d-8ec3-4af7-8b43-b715a081307c.png)

.gif example

![Mar-28-2023 14-44-26](https://user-images.githubusercontent.com/1316441/228363191-02295942-4647-4a0c-9650-dea545764875.gif)

This leaves you in a position where navigating the grid with the keyboard is difficult because you can't see the cursor. The cursor should always remain visible on screen.

(Discovered watching Francisco in a user session)

luke-quadratic:
This is fixed.

PR #1017, opened 2024-01-24
fix cell cursor visibility when the code editor is open (#383)

PawelJastrzebski:


davidkircos:
Planned fixed by adding a layout to the full application. Closing this

PR #1027, opened 2024-01-26
Follow markers

davidfig:
- [x] don't change order of users when following
- [x] add follow information to User and UserUpdate in Rust/TS
- [x] UI to indicate if users are following
- [x] track a user's index on the server to ensure all users see the same user number/color
- [x] multiplayer users are sorted by index so everyone sees the same user ordering
- [x] bug: anonymous user avatar doesn't display properly


quadratic-client/src/multiplayer/multiplayer.ts
```diff
@@ -40,28 +40,29 @@ export type MultiplayerState =
   | 'syncing';
 
 export class Multiplayer {
+  sessionId;
```
ddimaria, line 43: String?

quadratic-client/src/multiplayer/multiplayer.ts
```diff
@@ -354,7 +361,11 @@ export class Multiplayer {
   private receiveUsersInRoom(room: ReceiveRoom) {
     const remaining = new Set(this.users.keys());
     for (const user of room.users) {
-      if (user.session_id !== this.sessionId) {
+      if (user.session_id === this.sessionId) {
+        this.index = user.index;
+        this.colorString = MULTIPLAYER_COLORS[user.index % MULTIPLAYER_COLORS.length];
+        console.log(this.index);
```
ddimaria, line 367: Remove?

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -89,6 +91,13 @@ pub(crate) async fn handle_message(
 
             validate_can_edit_or_view_file(&permissions)?;
 
+            let follow = follow
+                .map(|follow| match Uuid::parse_str(&follow) {
```
ddimaria, line 95: You can remove the `flatten` and replace `map` with `flat_map`.  Is this code just converting a vec of string Uuids to Uuid?  If so, you can likely replace `|follow| match Uuid::parse_str(&follow)` with just `Uuid::parse_str`.

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -321,14 +333,11 @@ pub(crate) mod tests {
             session_id: user_1.session_id,
             file_id,
             update: UserStateUpdate {
-                sheet_id: None,
                 selection: Some("selection".to_string()),
```
ddimaria, line 336: All of these changes are going to have conflicts with [my branch](https://github.com/quadratichq/quadratic/pull/1013).

quadratic-multiplayer/src/message/handle.rs
```diff
@@ -509,4 +482,52 @@ pub(crate) mod tests {
 
         // TODO(ddimaria): mock the splitsink sender to test the actual sending
     }
+
+    #[tokio::test]
+    async fn enter_room_follow() {
```
ddimaria, line 487: I have utilities for this in my current branch: https://github.com/quadratichq/quadratic/pull/1013

quadratic-multiplayer/src/state/user.rs
```diff
@@ -181,6 +186,17 @@ impl State {
                 if let Some(viewport) = user_state.viewport.to_owned() {
                     user.state.viewport = viewport;
                 }
+                if let Some(follow) = user_state.follow.to_owned() {
+                    if follow.is_empty() {
+                        user.state.follow = None;
+                    } else {
+                        user.state.follow = if let Ok(uuid) = Uuid::parse_str(&follow) {
```
ddimaria, line 193: This is a duplicate of a pattern else where in your code.  Consider abstracting uuid parsing.  Also, you can avoid these chained conditionals and rewrite this in a more Rusty way.

davidfig: I'm not sure how to make this more rusty. Is there another way to check for an empty string? This is doing something slightly different than the first use case.

davidfig:
Converted into `staging-2024-01` branch

PR #1038, opened 2024-01-28
Multiplayer broken connection

### LINKED ISSUE #1012: Alert user trying to leave before syncing is finished
Description: alert the user with a popup box in the top with stay and leave options when they try to leave any time the sheet is syncing. 

Would also likely solve #891 and similar scenarios where user can leave without their work being saved 

Plenty of first hand experience thinking my work is synced and leaving.

davidfig:
- [x] improve sync status message by making the status red w/different tooltip text when reconnecting after a loss of connection
- [x] alert user that tries to navigate away from the page while syncing to the server
 * this provides a generic (non-configurable) alert dialog if the user is syncing to the multiplayer server while attempting to navigate away from the page
 * this will not work well on mobile devices, as there's not a great way to do it (see https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/ for a further discussion)
 * this only covers syncing w/the multiplayer server. For example, it does not cover the case where you are importing a large CSV file and attempt to navigate away while it is processing. Once that large file is processed, however, you will be alerted if you navigate away before it finishes uploading to the server.

fixes #1012


davidkircos:
CleanShot 2024-01-29 at 19 55 54@2x

davidfig:
Good catch. It works properly when there is stuff in the cell, but for empty cells, it's conflicting with the '=' to open the cell type window. I'll make sure we check for ctrl before opening the cell type window.

davidfig:
Should be fixed. You can try again.

PR #1040, opened 2024-01-29
Fix cell edit zoom

### LINKED ISSUE #900: Giving focus to a cell allow browser-level zooming
If you give a cell focus, then press cmd +, the browser zooms rather than the grid zooming.

davidfig:
- [x] fix ctrl+/- when editing a cell (fixes #900)
- [x] fix visual glitch where the auto-expand box (bottom-right) of a cell does not render when editing a cell - this glitch was caused by the grid cursor being unable to measure with the width of CellInput since the cursor draws before react renders the CellIput


davidkircos (changes requested):
Does not appear to fix 
CleanShot 2024-01-29 at 19 32 03

davidfig:
Good catch. It works properly when there is stuff in the cell, but for empty cells, it's conflicting with the '=' to open the cell type window. I fixed it by checking if ctrl or meta key is pressed before opening the cell-type menu. Please try again.

PR #1044, opened 2024-01-29
tests: for the files/:uuid/sharing endpoint

jimniels:


quadratic-api/src/routes/v0/files.$uuid.sharing.PATCH.ts
```diff
@@ -40,10 +40,10 @@ async function handler(req: Request, res: Response) {
...
 
   // Then edit!
-  await dbClient.file.update({
+  const newFile = await dbClient.file.update({
     where: { uuid },
     data: { publicLinkAccess },
   });
 
-  return res.status(200).json({ message: 'File updated.' });
+  return res.status(200).json({ publicLinkAccess: newFile.publicLinkAccess });
```
jimniels, line 48: We don't actually use this anywhere on the client, but seems like good practice to respond with the newly-updated information. Plus now we can use it in the tests to ensure the data changed as expected.

davidkircos (approved):

davidkircos (changes requested):
CleanShot 2024-01-29 at 20 02 11@2x

Issue #1051, opened 2024-01-30
Python append to array output instead of replacing it


davidfig:

luke-quadratic:
Exists but could be improved - closing, if someone wants to reopen or someone asks for it we can reopen, otherwise is possible in prod tday

PR #1068, opened 2024-02-01
remove protections against self referencing

davidfig:
This removes the protections around cell references within python's `cell` and `cells` commands. This will unlock self updating data within a code run

Addresses https://github.com/quadratichq/quadratic/issues/1051


quadratic-core/src/controller/execution/run_code/get_cells.rs
```diff
@@ -41,8 +41,8 @@ impl GridController {
             )));
         };
 
-        let (current_sheet, pos) = if let Some(current_sheet_pos) = transaction.current_sheet_pos {
-            (current_sheet_pos.sheet_id, current_sheet_pos.into())
+        let current_sheet = if let Some(current_sheet_pos) = transaction.current_sheet_pos {
```
ddimaria, line 44: A more idiomatic approach would be https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

quadratic-core/src/controller/execution/run_code/get_cells.rs
```diff
@@ -247,20 +225,20 @@ mod test {
...
-        let sheet = gc.sheet(sheet_id);
-        let error = sheet
-            .code_run(Pos { x: 0, y: 1 })
-            .unwrap()
-            .clone()
-            .std_err
-            .unwrap();
-        assert!(error.contains("cell cannot reference"));
+        assert_eq!(code.len(), 0);
+        // assert_eq!(code[0].special, Some(JsRenderCellSpecial::RunError));
```
ddimaria, line 233: Remove?

ddimaria (approved):

PR #1075, opened 2024-02-02
Replace array for set cell values

davidfig:
Replace Array to both simplify the code, reduce the size, and squeeze out a bit more performance when merging between `Sheet.column[].values` and new struct for use in `SetCellValues`.

- [x] create `CellValues` struct, which contains a `Vec<BtreeMap<u32, CellValue>>` to replace Array for use within `SetCellValues`
- [x] replace `Array` with `CellValues` where appropriate


quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -32,56 +33,35 @@ impl GridController {
...
-                    let _ = array.set(x, y, value);
-
-                    x += 1;
-
-                    if x == clipboard.w {
-                        x = 0;
-                        y += 1;
-                    }
-                });
+                let code = clipboard
```
ddimaria, line 47: Way more rusty 🦀

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -90,24 +70,20 @@ impl GridController {
...
         match special {
             PasteSpecial::None => {
-                let (values, code) = GridController::array_from_clipboard_cells(clipboard, special);
+                let (values, code) =
+                    GridController::cell_values_from_clipboard_cells(&clipboard, special);
                 if let Some(values) = values {
-                    ops.push(Operation::SetCellValues { sheet_rect, values });
+                    ops.push(Operation::SetCellValues {
+                        sheet_pos: start_pos,
+                        values: values.clone(),
```
ddimaria, line 85: Is this clone avoidable?

ddimaria (approved):

quadratic-core/src/controller/operations/clipboard.rs
davidfig: Regrettably, yes. It's passed as a reference, so we have to clone it to send it to ops.

PR #1083, opened 2024-02-05
Fix safari clipboard

davidfig:
To repro in main:

1. open quadratic in safari
2. add text content to your clipboard
3. double click to edit a cell
4. click somewhere else
5. try to copy a cell or paste the clipboard content

Expected: copy and paste works
Safari: copy will beep to let you know it was not successful; paste will not work


jimniels (approved):
Yup, works for me!

PR #1085, opened 2024-02-05
Search box

davidfig:
- [x] rust function to search a sheet
- [x] rust function to search across all sheet
- [x] exported rust function for TS
- [x] search dialog and implementation
- [x] search in command palette (both search in sheet and search in all sheets)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -26,6 +26,7 @@
     "pulumi",
     "rects",
     "reqwest",
+    "shadcn",
     "Signin",
     "smallpop",
     "Southborough",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -30,6 +30,7 @@
     "Signin",
     "smallpop",
     "Southborough",
+    "szhsin",
     "thiserror",
     "unspill",
     "vals",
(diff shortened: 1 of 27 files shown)


jimniels:
I pulled this down and added some style/design enhancements to the functionality (including responsive functionality for devices less than 400px wide)

After my tweaks, there are three things I still see with this:

### 1. Call the feature "Find" not "Search"

This feature feels much more akin to the native "Find" functionality so many apps have. In fact, it's even triggered by the browser's native "Find" keyboard shortcut (CMD + F).

As such, I tend to think we should call this "Find" and not "Search". That includes putting it under the "Edit" menu in the upper left, e.g. "Edit > Find" 



CleanShot 2024-02-07 at 09 43 34@2x

Accordingly, we would also rename it to "Find" in the command palette (and anywhere else it appears).

### 2. CMD + F should trigger _anywhere_

Right now, it seems the grid has to have focus in order to be able to trigger "CMD + F", otherwise the browser's "Find" functionality will be triggered.

For example: click on header then cmd + f will trigger the browser find UI

CleanShot 2024-02-07 at 09 44 32


### 3. Find box re-renders as it's closing

This is minor, but there appears to be an issue with the animation + closing the find box.

Closing triggers re-render as the box is closing (with the animation).

CleanShot 2024-02-07 at 09 38 24

Note how in the GIF the results re-render to "0 of 0" as the popover is closing. Not sure if there's an easy solution to this, but would be a nice thing to fix.

ddimaria (approved):

Issue #1056, opened 2024-01-30
Python typing 


luke-quadratic:
**Return types, io to/from Python**
#627 addresses return types from Python to sheet, this issue addresses types as they are ingested from sheet to Python 

**current behavior:** everything is just ingested as an object from sheet to Python, additionally single column ranges are ingested as dataframe when they should just be a series 

**outline:** 
c(x,y) where the value of c(x,y) is an int should be int when brought into python 
c(x,y) where the value of c(x,y) is a float should be float when brought into python 

cells((x,y),(x,y)) where the range is one column should be brought in as a series in Python 
cells((x,y),(x,y)) where the range is multiple columns should be brought in as a dataframe in Python 
additionally
if a column in the range is all ints, the series type should be int64
if a column in the range has any floats alongside ints or exclusively floats, the series type should be float64
if a column in the range has any strings, the series type should be object 
if a column in the range is all bools, then the series type should be bool 
if a column in the range is all datetimes, then the series type should be datetime 
(take note of the distinctions above when mixing types, straightforward when all same, but also needs to take note of best approach when mixing)

![image](https://github.com/quadratichq/quadratic/assets/146771258/3ea2d018-249e-4ff5-9491-af85d31e8de2)
![image](https://github.com/quadratichq/quadratic/assets/146771258/173ba6f4-b9a6-4c3a-bf90-805cf50ba11e)

Type mapping 
https://app.quadratichq.com/file/1ad99fbb-f42b-44ee-b261-588c82136e2e

davidkircos:
done

Issue #567, opened 2023-06-20
feat: python cell highlights


juanbercoff:
![image](https://github.com/quadratichq/quadratic/assets/15874591/e7e1c691-88ca-40e3-8a1f-5b059f03fdfb)

I worked on it in 5af393b280c8a734312d50870daa7d774caf8dc3 
Ideally should be implemented in rust to avoid code duplication.

AH-Hammi:
Would really love to see that, very good idea

ddimaria:
This issue is resolved and in https://github.com/quadratichq/quadratic/tree/staging-2024-01

PR #1087, opened 2024-02-07
Python run inspect

ddimaria:
Addresses: 
* https://github.com/quadratichq/quadratic/issues/1056
* https://github.com/quadratichq/quadratic/issues/627
* https://github.com/quadratichq/quadratic/issues/53
* https://github.com/quadratichq/quadratic/issues/567

- [x] Inspect return types in the console (add icon next to return line, hide icon with unsaved changes)
- [x] Support rust types -> python types and python types -> rust types
- [x] Output series when appropriate
- [x] Fix cases where the editor hangs if a TS error occurs when calculating python
- [x] Support the Pyright LSP (in JS in a web worker) for auto completion, hover and signature
- [x] Autogenerate python typshed stubs of Quadratic code and supported third-party libraries
- [x] Fix deltaDecorations bug in useEditorCellHighlights.ts
- [x] Enhance regex python parser to handle the various forms of cell/cells/c/getCell/getCells
- [x] Utilize Pyright diagnostics in Monaco
- [x] Handle NoneType returns in the UI 
- [x] Pyodide within a web worker now working in tests
- [x] Fixs stop-python-execution regression

Known Issues
- [x] AST parser sometimes returns the incorrect return line number
- [ ] Top-level async/await supported in Pyodide, but Pyright complains
- [ ] Hover overlay is not respecting markdown (see autocomplete for correct markdown support)
- [ ] Regex parser for Quadratic function highlights is naive and should be replaced with an AST parsed version


davidkircos:
Python stopping and restarting is broken 
CleanShot 2024-03-02 at 20 22 05@2x

This same bug exists in `main` btw

ddimaria:
> Python stopping and restarting is broken CleanShot 2024-03-02 at 20 22 05@2x
> 
> This same bug exists in `main` btw

I noticed this as well.  The python worker restarts but the UI remains in a loading state.  This is a regression.  I can look into it in my this PR.

PR #1094, opened 2024-02-14
Billing with Stripe

davidkircos:
Todo:
- [x] Add env vars to API
- [x] Associate Subscription ID with Team after Checkout, update on every load of Team that is invalid or outdated
- [x] Get Prices dynamically from API on the Team Create page


quadratic-api/src/internal/addUserToTeam.ts
```diff
@@ -0,0 +1,21 @@
...
+      userId,
+      teamId,
+      role,
+    },
+  });
+
+  // Update the seat quantity on the team's stripe subscription
+  await updateSeatQuantity(teamId);
+
+  // TODO: send them an email
```
jimniels, line 18: I don't think we'll ever send an email here. We'll do it in each specific place where this function is called. (For example: this gets called when a new user comes into the app, in which case they're coming from an invite and already received an email, so we shouldn't send another). 

So you can remove this.

quadratic-api/src/internal/removeUserFromTeam.ts
```diff
@@ -0,0 +1,19 @@
...
+        userId,
+        teamId,
+      },
+    },
+  });
+
+  // Update the seat quantity on the team's stripe subscription
+  await updateSeatQuantity(teamId);
+
+  // TODO: send them an email?
```
jimniels, line 18: remove

quadratic-api/src/internal/removeUserFromTeam.ts
```diff
@@ -0,0 +1,19 @@
+import dbClient from '../dbClient';
+import { updateSeatQuantity } from '../stripe/stripe';
+
+export const removeUserFromTeam = async (userId: number, teamId: number) => {
+  // Add user to team
```
jimniels, line 5: Remove comment or change to `Remove user from team`

quadratic-api/src/routes/v0/teams.$uuid.GET.ts
```diff
@@ -68,9 +71,14 @@ async function handler(req: Request, res: Response<ApiTypes['/v0/teams/:uuid.GET
       },
     },
   });
-  const dbFiles = dbTeam?.File ? dbTeam.File : [];
-  const dbUsers = dbTeam?.UserTeamRole ? dbTeam.UserTeamRole : [];
-  const dbInvites = dbTeam?.TeamInvite ? dbTeam.TeamInvite : [];
+
+  if (!dbTeam) {
+    return res.status(404).send();
+  }
```
jimniels, line 77: I can't remember if you did this already, but this can be removed (as `getTeam` above on line 33 will return a 404 if the team doesn't exist)

quadratic-api/src/internal/addUserToTeam.ts
```diff
@@ -0,0 +1,21 @@
+import { TeamRole } from '@prisma/client';
+import dbClient from '../dbClient';
+import { updateSeatQuantity } from '../stripe/stripe';
+
+export const addUserToTeam = async (userId: number, teamId: number, role: TeamRole) => {
```
jimniels, line 5: Minor: I would pass these as a single argument (object of keys) rather than multiple/positional arguments, e.g.

```js
// function
export const addUserToTeam = async (
  { userId, teamId, role } : 
  { userId: number, teamId: number, role: TeamRole) => { ... }

// usage
addUserToTeam({ userId: 5, teamId: 10, role: 'OWNER' });
```

quadratic-api/src/routes/v0/teams.POST.ts
```diff
@@ -31,9 +33,20 @@ async function handler(req: Request, res: Response<ApiTypes['/v0/teams.POST.resp
     picture: picture ? true : false,
   };
 
+  // Get user email from Auth0
+  const auth0Record = await getUsersFromAuth0([{ id: userId, auth0Id }]);
+  const auth0User = auth0Record[userId];
+  if (!auth0User) {
+    throw new Error(`User ${userId} not found in Auth0`);
+  }
```
jimniels, line 41: `auth0User` here should never be undefined, because `getUsersFromAuth0` will throw if it doesn't find a user with an email for the id you provided.

jimniels: If you look for other uses of `getUsersFromAuth0` in the codebase, you'll see we don't check for the existence of the ids we passed to it. They will be there.

quadratic-api/src/routes/v0/webhooks.stripe.POST.ts
```diff
@@ -0,0 +1,36 @@
+import { Request, Response } from 'express';
+import { STRIPE_WEBHOOK_SECRET, handleSubscriptionWebhookEvent, stripe } from '../../stripe/stripe';
+
+export default [handler];
+
+async function handler(req: Request, res: Response) {
+  console.log('webhook hit');
```
jimniels, line 7: Probably want to remove this?

quadratic-api/src/routes/v0/teams.$uuid.GET.ts
davidkircos: We are actually getting the team again, so this check still makes sense. The dbteam also contains Files, UserTeamRoles, and Invites. So if it fails to retrieve anything 404 is appropriate. This should not happen, but it makes the type system happy.

quadratic-client/src/api/apiClient.ts
```diff
@@ -47,6 +47,22 @@ export const apiClient = {
         ApiSchemas['/v0/teams.POST.response']
       );
     },
+    billing: {
+      async getPortalSessionUrl(uuid: string) {
+        return fetchFromApi(
+          `/v0/teams/${uuid}/billing/portal/session`,
+          { method: 'POST' },
```
jimniels, line 54: Minor: seems like we should make both of these GET endpoints since they don't actually post any data? they're just retrieving data.

PR #1101, opened 2024-02-16
Remove python package files from git

Asaaj:
.whl and .tar.gz are generated binary files. In general binary files should not be tracked by git (unless using LFS), but especially those which are generated.

Pending discussion and tie-in to package during build process

# Note to Developers:

This does mean `npm start` will have to install `poetry` (the virtual environment tool we're using for Python), then install the Python dependencies, then package the Python code and put the `.whl` file in your `quadratic-client/public` folder. All of this _should_ happen automatically, but obviously things can go wrong. If you have trouble loading Python, it could be because this step went wrong in some way. Feel free to contact me if that happens.

# TODO
- [x] ~~Catch missing Python binary artifacts (at least `quadratic-client/public/quadratic_py-0.1.0-py3-none-any.whl`) in CI somehow~~ (While this isn't directly resolved, I did ensure if the packaging fails, the deployment fails. This is a partial fix that should be sufficient for now)
- [x] Package Python code in CI (possibly done in most recent commit)
- [x] Package Python code in developer scripts


davidfig (approved):

ddimaria (approved):

ddimaria:
I validated that Python is working in the UI.

PR #1104, opened 2024-02-18
Add Python support to dev scripts

Asaaj:


Asaaj:
This would actually be bad to merge separately from #1101, since it would mean everyone would always end up with a modified tar.gz file that causes issues checking out different branches. Closing to add this commit to that PR instead.

PR #1106, opened 2024-02-20
Large Imports Fail with Rust error

davidfig:
This fixes a bug where it attempts to convert a string like `980E92207901934` into a number. (This is from a CSV.) The entry was probably intended text, but `BigDecimal` parsed it to a number, and then when it tried to serialize it, it ran out of memory when allocating the string since that's a humungous number.

For now, I put an upper limit as to the number of digits allowed for a number. If converting a text to a number exceeds that threshold, then it leaves it as `CellValue::Text`.

There are some interesting workarounds for this if we ever want to support huge numbers like this. See https://github.com/declanvk/bigdecimal-rs/commit/b0a2ea3a403ddeeeaeef1ddfc41ff2ae4a4252d6, which is a not-accepted PR that outputs the overly large number in E form.

- [x] locate cause of memory failure
- [x] increase `max_message_size` for ws in quadratic-multiplayer


ddimaria (approved):

quadratic-client/src/debugFlags.ts
```diff
@@ -62,7 +62,7 @@ export const debugShowFileIO = debug && false;
 
 export const debugGridSettings = debug && false;
 
-export const debugShowMultiplayer = debug && false;
+export const debugShowMultiplayer = debug && true;
```
davidkircos, line 65: @davidfig do you intend to change this?

davidfig: Doesn't matter. Not used in production.

PR #1111, opened 2024-02-20
teams: centralize env vars

jimniels:
Centralize all env vars into a single file and indicate which ones are required, required for prod only, and optional.

### DIFF
--- a/quadratic-api/src/env-vars.ts
+++ b/quadratic-api/src/env-vars.ts
@@ -1,0 +1,50 @@
+// Optional
+export const RATE_LIMIT_AI_WINDOW_MS = process.env.RATE_LIMIT_AI_WINDOW_MS;
+export const RATE_LIMIT_AI_REQUESTS_MAX = process.env.RATE_LIMIT_AI_REQUESTS_MAX;
+export const SENDGRID_API_KEY = process.env.SENDGRID_API_KEY;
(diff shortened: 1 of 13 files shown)


quadratic-api/src/env-vars.ts
```diff
@@ -0,0 +1,50 @@
...
+  'AUTH0_CLIENT_SECRET',
+  'AUTH0_JWKS_URI',
+  'AUTH0_ISSUER',
+  'AUTH0_AUDIENCE',
+  'AWS_S3_REGION',
+  'AWS_S3_ACCESS_KEY_ID',
+  'AWS_S3_SECRET_ACCESS_KEY',
+  'AWS_S3_BUCKET_NAME',
+  'STRIPE_SECRET_KEY',
+].forEach(ensureEnvVarExists);
```
ddimaria, line 35: You could make this less brittle by converting these variables into a config class and looping through the property descriptors (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors).  If you then export a singleton instance of the class, you would have a similar API to what you're already using in the consuming files.

ddimaria (approved):
Code looks good!  Approved, but you may be able to enhance the design a bit (added a comment).

PR #1121, opened 2024-02-23
updating language

davidkircos:


quadratic-client/src/components/ShareDialog.tsx
```diff
@@ -110,13 +110,13 @@ export function ShareTeamDialog({
 
   return (
     <ShareDialog
-      title={`Share ${name}`}
+      title={`${name} Team`}
```
jimniels, line 113: If we change this here, we should do the same for file, e.g. rather than stating the name

CleanShot 2024-02-23 at 18 32 32@2x

State the noun

CleanShot 2024-02-23 at 18 32 44@2x

davidkircos: I think they are different. 

Share Q4 Report vs Quadratic Team

"Share Quadratic" feels weird to me. It's a team you invite someone to a team, you don't share it.

jimniels (approved):

PR #1134, opened 2024-02-29
fix AI tab when output has a chart

davidkircos:
AI doesn't work in prod on any cell with a chart output. This is because the ~ 3mb that each chart takes up (wow!) is sent as context to the AI API.

I changed it to simply send the std error, this fixes the issue.


jimniels (approved):
Seems to work. I don't see a huge blob sent to AI when I have a chart cell open and use the AI bot.

And the code is 👍  — great that we have explicit types now instead of just `any`

PR #1135, opened 2024-02-29
feat: "Press '=' to code" placeholder

jimniels:
CleanShot 2024-02-29 at 13 43 15@2x

Builds off https://github.com/quadratichq/quadratic/pull/996/files

This implementation uses a flag in localstorage which tracks how many times the cell type menu has been triggered (when the user presses `=`). When the cell type menu has been triggered more than 3 times (configurable), the placeholder/hint will disappear.

Also comes with the recommendation that we put Python first in the list of code options

Note: the overflow works by just doing a white background on the placeholder. It's not the best, but I think it works:

CleanShot 2024-02-29 at 10 37 00@2x

**Later improvements:**

- Consider showing/hiding the placeholder based on a more nuanced state of the code editor. For example: when the code editor is open, don't show the "Press '=' to code", but _only for the currently active code editor cell_ (e.g. if the code editor is open for (0,0) but my cursor is active on (1,0) still show the message because that's technically still a valid state where you can type `=`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ _Quadratic is in Beta._
 - [x] Undo / Redo (issue [#42](https://github.com/quadratichq/quadratic/issues/42))
 - [x] Multiplayer Support
 - [x] Charts and Graphs
-- [ ] Teams support 
+- [x] Teams support 
 - [ ] JS support 
 - [ ] SQL Database Support
 - [ ] AI Auto Complete
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "quadratic-multiplayer"
       ],
       "dependencies": {
+        "usehooks-ts": "^2.15.1",
         "zod": "^3.22.4"
       },
       "devDependencies": {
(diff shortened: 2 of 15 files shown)


jimniels:
How do we solve the overflow problem (especially if someone has shrunk a column width)

CleanShot 2024-02-29 at 09 57 43@2x

davidkircos (changes requested):
Looks great!

But, does not disappear for me after 3 code runs

jimniels:
Was seeing an issue in the deployment branch that I wasn't seeing locally. Seemed, oddly, like an environment problem with reading values from localstorage.

https://github.com/quadratichq/quadratic/assets/1316441/8fc41bda-6b6d-44bd-afe1-8946d53bb1a3

Turns out this was a result of `<StrictMode>` running on localhost. If you turned off `<StrictMode>` locally, you would see what was happening in production.

The bug was tracked down to this file: `quadratic-client/src/hooks/useLocalStorage.ts` which appears to be some custom copy/paste of a hook that is very different than what's on the [usehooks-ts](https://usehooks-ts.com/) page now

Because it wasn't immediately apparent, and I didn't want to troubleshoot that hook, I just installed their hook and use that one instead.

TBH: we should probably move all implementations of our custom `useLocalStorage.ts` hook to using the one from the usehooks-ts package entirely.

jimniels:
There's some kind of reactivity bug with the useLocalStorage that isn't consistent depending on whether you're in strict mode or not (in react).

I opted out of using the hook and am instead using recoil for this piece of state and then syncing it to browser storage.

davidkircos (approved):

PR #1161, opened 2024-03-07
Ship: Fix bug with duplicate sheet not working

davidfig:
This fixes the bug we discovered when trying to duplicate sheets in main.


quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
```diff
@@ -164,6 +166,57 @@ impl GridController {
...
+        {
+            let Some(sheet) = self.try_sheet(sheet_id) else {
+                // sheet may have been deleted
+                return;
+            };
+            let mut new_sheet = sheet.clone();
+            new_sheet.id = new_sheet_id;
+            let right = self.grid.next_sheet(sheet_id);
+            let right_order = right.map(|right| right.order.clone());
+            new_sheet.order = key_between(&Some(sheet.order.clone()), &right_order).unwrap();
```
ddimaria, line 188: unwrap here

quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
```diff
@@ -164,6 +166,57 @@ impl GridController {
...
+            let name = format!("{} Copy", sheet.name);
+            let sheet_names = self.sheet_names();
+            if !sheet_names.contains(&name.as_str()) {
+                new_sheet.name = name;
+            } else {
+                new_sheet.name = crate::util::unused_name(&name, &self.sheet_names());
```
ddimaria, line 194: You may be able to avoid this conditional since you're already passing in the sheet names into the unused_name() fn.

quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
```diff
@@ -164,6 +166,57 @@ impl GridController {
...
+            }
+            self.grid.add_sheet(Some(new_sheet.clone()));
+
+            transaction.summary.sheet_list_modified = true;
+            transaction.summary.html.insert(new_sheet_id);
+            transaction
```
ddimaria, line 200: Is this supposed to happen twice?

davidfig: No, it's not. Not sure how I didn't catch this.

quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
davidfig: I think it's necessary unless I move the conditional into the unused_name function

quadratic-core/src/util.rs
```diff
@@ -214,8 +255,12 @@ pub fn unused_name(prefix: &str, already_used: &[&str]) -> String {
         .collect();
 
     // Find the first number that's not already used.
-    let i = (1..).find(|i| !already_used_numbers.contains(i)).unwrap();
-    format!("{prefix} {i}")
+    if let Some(i) = (1..).find(|i| !already_used_numbers.contains(i)) {
+        format!("{prefix} {i}")
```
ddimaria, line 259: Some duplication here, you could just return `i` and have a single `format!("{prefix} {i}")` after the conditional.

ddimaria (approved):
2 comments, non-blocking

PR #1165, opened 2024-03-08
Force modal for refresh

davidfig:
Dev notes: to change the recommended or required version, change the values in `/minVersion.json`. Increment either value and the dialog will pop for users not running the correct version.

- [x] add minVersion.json `{ requiredVersion: number, recommendedVersion: number }` referenced by both quadratic-multiplayer and quadratic-client
- [x] serve minVersion.json from quadratic-multiplayer when sending `usersInRoom` (first message after logging in)
- [x] fix node dev so you can turn off HMR for React (still defaults to on)
- [x] pop dialog for either required or recommended version
- [x] fixed clippy

Screenshot 2024-03-08 at 7 19 02 AM

Screenshot 2024-03-08 at 7 19 53 AM

__Local testing__

This is tricky to test. In this branch I updated node dev and added a -r command that let’s you turn off React’s hmr. That way you can update the json and then press m to refresh the multiplayer server. The client and multiplayer will be out of sync and the refresh dialog should appear. (It won’t clear on refresh unless you press r again.) There's not a great way to do this in the preview branch.


minVersion.json
davidkircos, line 1: name `clientCoreSupportedVersion.json` ?

minVersion.json
```diff
@@ -0,0 +1,4 @@
+{
+  "recommendedVersion": 1,
+  "requiredVersion": 2
```
davidkircos, line 3: Add comments explaining how this work.

```
// Core requires at least requiredVerison to work.
// If multiplayer is ahead of client requiredVerison client will force the user to reload
// If multiplayer is ahead of client recomendedVersion client will ask the user to reload
// These version numbers are simply incremented every time we have breaking changes
```

jimniels:
Updated to use similar styles/placement as is used elsewhere in the app.

CleanShot 2024-03-08 at 16 47 37@2x

CleanShot 2024-03-08 at 16 47 48@2x

quadratic-client/src/ui/QuadraticUI.tsx
```diff
@@ -102,6 +103,7 @@ export default function QuadraticUI() {
       )}
       {presentationMode && <PresentationModeHint />}
       {!isEmbed && <PermissionOverlay />}
+      <VersionControl />
```
jimniels, line 106: Per our convo, we should probably name this something like `MultiplayerVersionControlAlert` or something to that effect.

minVersion.json
davidfig: Not a great name since multiplayer also uses this json. Maybe something like promptVersion.json (although none of these are self-explanatory, which is why I originally chose minVersion.json)

quadratic-client/src/ui/QuadraticUI.tsx
davidfig: Settled on `VersionControlAlert`, since it's not really limited to MP.

minVersion.json
davidkircos: `updateAlertVersion`

davidkircos (approved):

PR #1172, opened 2024-03-12
Python relative references

ddimaria:
- [x] Add pos() function in python
- [x] Add rel_cell()/rc() function in python
- [x] Document the functions for pyright
- [x] Test pos() and rel_cell() in python

Adds 2 new functions (and one alias):

**pos()**
```python
# x, y are coordinates of the calling code cell
(x, y) = pos()
```

**rel_cell(x, y)** or **rc(x, y)**
```python
# c is the cell to the left of this one
c = rel_cell(-1, 0)
c = rc(-1, 0)

# this is equivalent to
(x, y) = pos()
c = cell(x - 1, 0)
```

Screenshot 2024-03-14 at 5 39 46 PM

Screenshot 2024-03-14 at 5 38 20 PM


quadratic-kernels/python-wasm/quadratic_py/run_python.py
```diff
@@ -46,6 +46,19 @@ async def getCellsInner(p0: Tuple[int, int], p1: Tuple[int, int], sheet: str=Non
 
     return await getCells(p0, p1, sheet, first_row_header)
     
+# Wrapper to pos() to capture cells_accessed
+async def getPosInner() -> Tuple[int, int] | None:
+    (x, y) = await pos()
+    cells_accessed.append([x, y, None])
```
davidkircos, line 52: I don't believe that referencing the point should add it to cells accessed (because this is self)

quadratic-kernels/python-wasm/quadratic_py/quadratic_api/quadratic.py
```diff
@@ -142,3 +143,38 @@ async def cells(p0: Tuple[int, int], p1: Tuple[int, int], sheet: str=None, first
...
+    Returns:
+        The tuple (x, y) coordinates of the current cell.
+
+    Typical usage example:    
+        (x, y) = pos()
+    """
+
+    return await getPos()
+
+async def rel_cell(x: int, y: int) -> int | float | str | bool | None:
```
davidkircos, line 159: why can't we call getPos and reuse getCell here?

quadratic-kernels/python-wasm/quadratic_py/run_python.py
```diff
@@ -46,6 +46,19 @@ async def getCellsInner(p0: Tuple[int, int], p1: Tuple[int, int], sheet: str=Non
...
+
+    return (x, y)
+
+# Wrapper to rel_cell() to capture cells_accessed
+async def getRelCellInner(x: int, y: int) -> int | float | str | bool | None:
+    [result, cells_accessed] = await rel_cell(x, y)
+    cells_accessed.append(cells_accessed)
```
davidkircos, line 59: Would be automatically handled if we rely on getCell instead of a new function

quadratic-client/src/web-workers/pythonWebWorker/pythonTypes.ts
```diff
@@ -26,6 +26,8 @@ export interface PythonMessage {
     | 'execute'
     | 'not-loaded'
     | 'get-cells'
+    | 'get-pos'
+    | 'get-rel-cell'
```
davidkircos, line 30: Instead of messages can we pass the cell pos on `execute` and save it in python state for all calls?

davidkircos: The variable `first` on this line of code has a property `sheetPos`, which is the position of the cell being executed. You can pass that along to the worker, then to run_python as an arg. Then getPos in python is just a pure function. https://github.com/quadratichq/quadratic/blob/9dbc0bc31cd6954fbf4d205ce1abaaeb6ef7febe/quadratic-client/src/web-workers/pythonWebWorker/python.ts#L192

quadratic-kernels/python-wasm/quadratic_py/run_python.py
ddimaria: Can you elaborate on the usage of cells_accessed and it's full meaning?

quadratic-kernels/python-wasm/quadratic_py/run_python.py
ddimaria: I don't think it would as we get the cells accessed in a different way.  Can you please show code to illustrate this point?

quadratic-client/src/web-workers/pythonWebWorker/pythonTypes.ts
ddimaria: This is done

quadratic-kernels/python-wasm/quadratic_py/run_python.py
ddimaria: This is resolved.

quadratic-kernels/python-wasm/quadratic_py/run_python.py
ddimaria: This is removed.

quadratic-kernels/python-wasm/quadratic_py/quadratic_api/quadratic.py
ddimaria: This is no longer an issue.  This function is just a stub for pyright to consume.

davidkircos (approved):

PR #1187, opened 2024-03-15
🚢 Refactor run_python to make it more testable

davidkircos:
Moves processing output into it's own function to make it possible to test

- [x] move logic to a function
- [x] write tests

### DIFF
--- a/quadratic-kernels/python-wasm/quadratic_py/process_output.py
+++ b/quadratic-kernels/python-wasm/quadratic_py/process_output.py
@@ -1,0 +1,95 @@
+import pandas as pd
+
+from .utils import to_quadratic_type
+
+
+def process_output_value(output_value):
+    # return array_output if output is an array
+    array_output = None
+    output_type = type(output_value).__name__
+    output_size = None
+
+    # TODO(ddimaria): figure out if we need to covert back to a list for array_output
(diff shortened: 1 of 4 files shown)


quadratic-kernels/python-wasm/quadratic_py/process_output.py
```diff
@@ -0,0 +1,95 @@
...
+    output_type = type(output_value).__name__
+    output_size = None
+
+    # TODO(ddimaria): figure out if we need to covert back to a list for array_output
+    # We should have a single output
+    if isinstance(output_value, list):
+        if len(output_value) > 0:
+            array_output = output_value
+            if (isinstance(output_value[0], list)):
+                output_size = (len(output_value[0]), len(output_value))
```
davidkircos, line 18: Fixed a bug here with 2d array sizing

ddimaria: We lost the diff.  Is this the only change in this code?

davidkircos: yes

ddimaria (approved):
Great test additions!  Python testing is improving greatly!

PR #1189, opened 2024-03-15
Remove cell highlighting for python

ddimaria:
Removes cell highlighting for Python.  Code is commented out with notes for future revival.

The current regex is naive and misses variables in the cell/cells call:

```python
a = 0
b = 0
c = cell(a, b)
```


ddimaria:
After a discussion with @davidkircos, closing this one as the current implementation is good enough to be useful.

PR #1190, opened 2024-03-15
Properly cast blank in python, test to_python_type()

ddimaria:
- [x] Override "blank" value, cast based on usage
- [x] Test to_python_type()

 
Screenshot 2024-03-15 at 12 57 18 PM

### DIFF
--- a/quadratic-kernels/python-wasm/tests/test.py
+++ b/quadratic-kernels/python-wasm/tests/test.py
@@ -229,10 +229,8 @@ class TestUtils(TestCase):
         # logical
         assert to_python_type("True", "logical") == True
         assert to_python_type("False", "logical") == False
-        assert to_python_type("True", "logical") == True
-        assert to_python_type("False", "logical") == False
-        assert to_python_type("True", "logical") == True
-        assert to_python_type("False", "logical") == False
+        assert to_python_type("true", "logical") == True
+        assert to_python_type("false", "logical") == False
 
         # string
         assert to_python_type("abc", "text") == "abc"
@@ -240,10 +238,6 @@ class TestUtils(TestCase):
         assert to_python_type("abc123", "text") == "abc123"
 
         # instant
-        assert to_python_type("1352505600", "instant") == pd.Timestamp("2012-11-10 00:00:00")
-        assert to_python_type("1352518200", "instant") == pd.Timestamp("2012-11-10 03:30:00")
-        assert to_python_type("1352505600", "instant") == pd.Timestamp("2012-11-10 00:00:00")
-        assert to_python_type("1352518200", "instant") == pd.Timestamp("2012-11-10 03:30:00")
         assert to_python_type("1352505600", "instant") == pd.Timestamp("2012-11-10 00:00:00")
         assert to_python_type("1352518200", "instant") == pd.Timestamp("2012-11-10 03:30:00")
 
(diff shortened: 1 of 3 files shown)


quadratic-kernels/python-wasm/tests/test.py
```diff
@@ -213,6 +213,39 @@ def test_to_quadratic_type(self):
 
         # TODO(ddimaria): implement when we implement duration in Rust
         # duration
+class TestUtils(TestCase):
+    def test_to_python_type(self):
+        # blank
+        assert to_python_type("", "blank") == 0
```
davidkircos, line 219: I think we use `self.assert()` instead of the built-in assertions 

@Asaaj do you know if it matters?

Asaaj: The only real difference is that the `assert` keyword is optimized out in "optimized" mode, but I don't run tests with that flag because that's kind of useless. I used `assert` because that's what I'm used to, I forgot you were using `unittest` instead of `pytest`

quadratic-kernels/python-wasm/tests/test.py
```diff
@@ -213,6 +213,39 @@ def test_to_quadratic_type(self):
...
+        
+        # number
+        assert to_python_type("1", "number") == 1
+        assert to_python_type("1.1", "number") == 1.1
+        assert to_python_type("-1", "number") == -1
+        assert to_python_type("1", "number") == 1
+        assert to_python_type("1.1", "number") == 1.1
+        assert to_python_type("-1", "number") == -1
```
davidkircos, line 227: duplicates

quadratic-kernels/python-wasm/tests/test.py
```diff
@@ -213,6 +213,39 @@ def test_to_quadratic_type(self):
...
+
+        # logical
+        assert to_python_type("True", "logical") == True
+        assert to_python_type("False", "logical") == False
+        assert to_python_type("True", "logical") == True
+        assert to_python_type("False", "logical") == False
+        assert to_python_type("True", "logical") == True
+        assert to_python_type("False", "logical") == False
```
davidkircos, line 235: duplicates. try diffent casing

quadratic-kernels/python-wasm/tests/test.py
```diff
@@ -213,6 +213,39 @@ def test_to_quadratic_type(self):
...
+        assert to_python_type("123abc", "text") == "123abc"
+        assert to_python_type("abc123", "text") == "abc123"
+
+        # instant
+        assert to_python_type("1352505600", "instant") == pd.Timestamp("2012-11-10 00:00:00")
+        assert to_python_type("1352518200", "instant") == pd.Timestamp("2012-11-10 03:30:00")
+        assert to_python_type("1352505600", "instant") == pd.Timestamp("2012-11-10 00:00:00")
+        assert to_python_type("1352518200", "instant") == pd.Timestamp("2012-11-10 03:30:00")
+        assert to_python_type("1352505600", "instant") == pd.Timestamp("2012-11-10 00:00:00")
+        assert to_python_type("1352518200", "instant") == pd.Timestamp("2012-11-10 03:30:00")
```
davidkircos, line 248: duplicates

quadratic-kernels/python-wasm/tests/test.py
ddimaria: Done

quadratic-kernels/python-wasm/tests/test.py
ddimaria: Removed

quadratic-kernels/python-wasm/tests/test.py
ddimaria: Done

davidkircos (changes requested):
remove `quadratic-client/.env.sentry-build-plugin`

davidkircos (approved):

davidkircos (approved):

Issue #1132, opened 2024-02-28
Parquet drag and drop 


luke-quadratic:
Same as .CSV, add Parquet drag and drop to sheet 

Export parquet scoped in sep issue: https://github.com/quadratichq/quadratic/issues/1205

PR #1203, opened 2024-03-20
Support drag and drop for parquet files

ddimaria:
Closes: https://github.com/quadratichq/quadratic/issues/1132

- [x] Drag and drop parquet file on the grid
- [x] Support basic arrow types

In future work: add additional type support as needed

pr-parquet-drag-and-drop


quadratic-core/src/controller/operations/import.rs
```diff
@@ -54,4 +55,54 @@ impl GridController {
...
+
+        let reader = builder.build()?;
+
+        for (row_index, batch) in reader.enumerate() {
+            let batch = batch?;
+            let num_rows = batch.num_rows();
+            let num_cols = batch.num_columns();
+
+            for col_index in 0..num_cols {
+                let col = batch.column(col_index);
```
ddimaria, line 90: Parquet decides on batch size, typically about 1024 items per column.

quadratic-client/src/ui/components/FileUploadWrapper.tsx
davidfig, line 1: Can you add .parquet to the use a supported format error message

davidfig: Line 70

quadratic-core/src/controller/operations/import.rs
davidfig: This makes sense. But we lose the ability to display early since we won't have all the columns in the viewport.

davidfig (approved):

davidfig:
A few comments and one suggested change.

quadratic-core/src/controller/operations/import.rs
ddimaria: It's not clear, but ` for (row_index, batch) in reader.enumerate()` loops through rows in batches, and gets the columns in batches.

quadratic-client/src/ui/components/FileUploadWrapper.tsx
ddimaria: Nice catch!  This is done.

PR #1213, opened 2024-03-22
Import excel v1

ddimaria:
Closes https://github.com/quadratichq/quadratic/issues/1154

- [x] File selector on dashboard for xlsx files
- [x] Map common excel types to quadratic types

upload-excel-file


quadratic-client/src/dashboard/components/CreateFileButton.tsx
```diff
@@ -18,36 +31,100 @@ export default function CreateFileButton() {
...
+      <DropdownMenuItem
+        asChild
+        onSelect={(e) => {
+          // We have to prevent this (and handle the `open` state manually)
+          // or the file input's onChange handler won't work properly
+          e.preventDefault();
+        }}
+      >
+        <label className="cursor-pointer">
+          Import <span className="mx-1 font-mono">.{extension}</span> file
```
davidkircos, line 103: remove

quadratic-client/src/helpers/files.ts
```diff
@@ -18,3 +18,27 @@ export function readFileAsArrayBuffer(file: File): Promise<Uint8Array> {
     reader.readAsArrayBuffer(file);
   });
 }
+
+export function stripExtension(name: string): string {
+  return name.replace(/\.[^/.]+$/, '');
+}
+
+export function isCsv(file: File): boolean {
+  return file.type === 'text/csv' || file.type === 'text/tab-separated-values';
```
davidkircos, line 27: Some of these we are using file extension, and some are using mime type. Should we be consistent?

jimniels (dismissed):
Somehow a new button got added? And the border radius for the split button got changed

CleanShot 2024-03-26 at 13 16 55@2x

Should still look like what's in prod:
CleanShot 2024-03-26 at 13 17 55@2x

quadratic-client/src/dashboard/components/CreateFileButton.tsx
ddimaria: Done, replaced with a separate drop-down: 
Screenshot 2024-03-26 at 1 20 16 PM

jimniels:
Should we also have excel in the import menu in the app?

CleanShot 2024-03-26 at 13 20 05@2x

(and the command palette)

quadratic-client/src/helpers/files.ts
ddimaria: Mime type is preferred.  For parquet, we have to use the file extension as the mime type was just registered this month and isn't supported in current browsers (empty mime type).  I don't have a strong opinion about consistency here (I kept  the original csv detection as it existed in the codebase).  I'll add a `.csv` extension check as well.

ddimaria:
> Somehow a new button got added? And the border radius for the split button got changed
> 
> CleanShot 2024-03-26 at 13 16 55@2x
> 
> Should still look like what's in prod: CleanShot 2024-03-26 at 13 17 55@2x

@jimniels that was by accident.  @davidkircos requested a separate drop-down, so it was replaced with this:
Screenshot 2024-03-26 at 1 29 23 PM

ddimaria:
> Should we also have excel in the import menu in the app?

@jimniels Excel imports only happen on the dashboard (i.e. not drag and drop).  Does that change the need to have Excel in the import menu and the command pallete?

PR #1225, opened 2024-03-26
fix: resize col with "press / to code"

### LINKED ISSUE #1208: Press / to code and fix resizing error
Press '/' to code instead of '=' 

Bug alongside this: resizing column leaves the press = to code behind 

Image

jimniels:
Fixes #1208


jimniels:
Note: there is a known issue on this PR where resizing a column in the negative region for the x-axis has lag when re-rendering the placement of the "Press / to code" placeholder, whereas resizing a column in the positive region of the x-axis has no lag.

https://github.com/quadratichq/quadratic/assets/1316441/1b181e9f-732e-458e-85b0-2d41ca4d308e

davidkircos (approved):

PR #1248, opened 2024-03-29
fix: layout bugs on mobile

jimniels:
Currently:

- Command palette shows on mobile
- Permission overlay buttons wrap

We don't want either of these.

Screenshot 2024-03-28 at 3 55 14 PM

To test:

Open [a shared file on the preview branch on mobile](https://quadratic-git-fix-mobile-layout-bugs.vercel.quadratic-preview.com/file/f32db818-2fe5-4861-923d-d9871466a3ca) and see that the command palette icon is gone and the buttons don't wrap.


quadratic-client/src/ui/components/PermissionOverlay.tsx
```diff
@@ -30,9 +30,9 @@ export function PermissionOverlay() {
     return (
       <FixedBottomAlert>
         <Type>
-          <strong>Welcome to Quadratic.</strong> You must log in to edit this file.
+          <strong>Welcome to Quadratic.</strong>
```
ddimaria, line 33: Just making sure this copy deletion was intentional?

ddimaria:
Confirmed that the command pallet is hidden and that there is no wrapping on `log in` and `sign up` in iPhone 15 (smaller format).

ddimaria (approved):
One non-blocking question.

jimniels:
Yup, it was. Implied the presence of the buttons. Merging...

PR #1257, opened 2024-04-03
replace Blank with None

davidkircos:


luke-quadratic:
this works, but nan may be better, you'd be able to do math on empty cell that sometimes may be populated and not get error

luke-quadratic (approved):
lgtm

Issue #1158, opened 2024-03-06
Relative References + Autocomplete Formulas


luke-quadratic:
Autocomplete fills in formulas with relative references.
Copy/paste should also paste relative cell references in formulas, not absolute references.

Tasks
- [x] Auto-complete code cells (formulas and python): https://github.com/quadratichq/quadratic/pull/1230
- [ ] Relative references in auto-complete
- [ ] Relative references in copy/paste

PR #1266, opened 2024-04-08
Relative formula references

ddimaria:
Closes https://github.com/quadratichq/quadratic/issues/1158

- [x] Support relative and absolute references in formulas
- [x] Support relative and absolute references for formulas in cut/copy/paste
- [x] Support relative and absolute references for formula in autocomplete
- [x] Support relative and absolute references for formula in cell highlighting
- [x] Add tests

Under the hood, we're storing cell references as the internal cell reference notation (`R[0]C{-1}`).  Users only see the A1 notation (`nA0`) in the code editor.  Using the approach, cut/copy/paste and autocomplete work without modification.  Additionally, existing syntax is supported without any required changes by users.


quadratic-client/src/ui/menus/CodeEditor/useEditorCellHighlights.ts
```diff
@@ -127,27 +139,19 @@ export const useEditorCellHighlights = (
...
       }
     };
 
     onChangeModel();
     editor.onDidChangeModelContent(onChangeModel);
-  }, [
-    isValidRef,
-    editorRef,
-    monacoRef,
-    editorInteractionState.selectedCell,
```
davidfig, line 149: Why do you remove the editorInteractionState.var from the dependencies? Doesn't that trigger them less frequently when teh change is unrelated to the selectedCell, etc.?

davidfig (approved):

PR #1267, opened 2024-04-08
feat: configurable panel for console + ai assistant

jimniels:
- Allows the positioning of the console + ai assistant to be configurable, either underneath the code editor, or to the left of it.
- Updates some mui components to use shadcn 

CleanShot 2024-04-10 at 12 59 28@2x


quadratic-client/src/gridGL/HTMLGrid/htmlCells/htmlCellsHandler.ts
```diff
@@ -10,7 +10,7 @@ class HTMLCellsHandler {
   private div?: HTMLDivElement;
 
   attach(parent: HTMLDivElement) {
-    if (this.div) {
+    if (this.div && parent) {
```
jimniels, line 13: This is breaking HMR, so I added this and now it works.

ddimaria:
This branch needs to merge in main.  I went through the code and the UI and everything acts as expected.  

I don't have great vision, so the copy was hard to read, but it's the same as before  (CMD + only zooms in the grid).  

I wasn't familiar with the intent of the asterisk next to `Console *` in the bottom mode, seems like it was a reference to something more descriptive below rather than an indicator of content in the console.  I don't have any great suggestions to make this more clear though.

ddimaria (approved):
Approved, but @davidkircos will review/approve before merging.

jimniels:
FWIW: I filed a separate issue for the font size increase/decrease discrepancies. These are bugs that exist today and aren't necessarily introduced by this PR.

https://github.com/quadratichq/quadratic/issues/1277

Issue #1269, opened 2024-04-08
Improved alerts/logs for Redis and other services


ddimaria:
- [x] Add Redis monitoring to Datadog and set utilization thresholds for alarms
- [ ] Convert Redis connection warnings in Multiplayer to errors
- [ ] Send multiplayer errors in the frontend to Sentry

PR #1271, opened 2024-04-08
Log errors in multiplayer and the client for pubsub errors

ddimaria:
- [x] Convert Redis connection warnings in Multiplayer to errors
- [x] Send multiplayer errors in the frontend to Sentry

Closes: https://github.com/quadratichq/quadratic/issues/1269

Console errors now for multiplayer error messages sent back to the client:
Screenshot 2024-04-08 at 3 22 58 PM

### DIFF
--- a/quadratic-client/src/multiplayer/multiplayer.ts
+++ b/quadratic-client/src/multiplayer/multiplayer.ts
@@ -604,7 +604,7 @@ export class Multiplayer {
     } else if (type === 'CurrentTransaction') {
       this.receiveCurrentTransaction(data);
     } else if (type === 'Error') {
-      if (data.error_level == 'Error') {
+      if (data.error_level === 'Error') {
         Sentry.captureException({
           message: `Error response from the multiplayer server: ${data.error}`,
           level: 'error',
--- a/quadratic-client/src/multiplayer/multiplayer.ts
+++ b/quadratic-client/src/multiplayer/multiplayer.ts
@@ -10,6 +10,7 @@ import { SheetPos } from '@/gridGL/types/size';
 import { displayName } from '@/utils/userUtil';
 import { pythonWebWorker } from '@/web-workers/pythonWebWorker/python';
 import { User } from '@auth0/auth0-spa-js';
+import * as Sentry from '@sentry/react';
 import { v4 as uuid } from 'uuid';
 import sharedConstants from '../../../updateAlertVersion.json';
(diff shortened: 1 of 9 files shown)


quadratic-client/src/multiplayer/multiplayer.ts
```diff
@@ -603,7 +604,15 @@ export class Multiplayer {
     } else if (type === 'CurrentTransaction') {
       this.receiveCurrentTransaction(data);
     } else if (type === 'Error') {
-      console.warn(`[Multiplayer] Error`, data.error);
+      if (data.error_level === 'Error') {
+        Sentry.captureException({
+          message: `Error response from the multiplayer server: ${data.error}`,
```
davidkircos, line 609: In the logs we get this error all the time from Multiplayer. Will this also trigger Sentry now?
CleanShot 2024-04-08 at 23 15 11@2x

quadratic-multiplayer/src/server.rs
```diff
@@ -197,10 +200,19 @@ async fn handle_socket(
             Ok(ControlFlow::Continue(_)) => {}
             Ok(ControlFlow::Break(_)) => break,
             Err(error) => {
-                tracing::warn!("Error processing message: {:?}", &error);
+                let mut error_level = ErrorLevel::Warning;
+
+                match error {
+                    MpError::PubSub(_) => {
+                        tracing::error!("Error processing message: {:?}", &error);
```
davidkircos, line 207: Should we add a test? Code Cov is complaining

quadratic-client/src/multiplayer/multiplayer.ts
ddimaria: If you look at the Rust code, we only send an error_level of `Error` for PubSub errors, not the other errors in the logs.  Are there PubSub errors that you think should be warnings?

quadratic-multiplayer/src/server.rs
ddimaria: I may be able to add a test for this unique case (assuming I can simulate a PubSub error in an integration test (tests on server.rs need to be integration tests).  I'll report back and timebox the effort to 2 hours.

ddimaria: I refactored this logic to occur outside of this function for better isolation.

quadratic-client/src/multiplayer/multiplayer.ts
davidkircos: Makes sense

davidkircos (approved):

PR #1273, opened 2024-04-09
Javascript

davidfig:
- [x] basic javascript computation
- [x] typescript support using esbuild-wasm
- [x] allow async operations in javascript
- [x] getCell/getCells in javascript (requires await)  
- [x] map errors to proper line number (for execution errors)
- [x] show an error when trying to import modules 
- [x] ensure good errors during esbuild transform
- [x] add code highlighting and language server support (comes out of the box with monaco for typescript) 
- [x] add output type information to results
- [x] getPos and get relative cells
- [x] JS logo in code editor 
- [x] add icons/prettify UI
- [x] better console.log support for formatting Objects, Arrays, DateTime, and boolean
- [x] return line numbers
- [x] handle Infinity and NaN
- [x] import ESM modules
- [x] reimplement return line numbers
- [x] output png from OffscreenCanvas
- [x] merge conflicts with main
- [x] corner resize
- [x] return Canvas instead of Blob to make it easier on user
- [x] tests for getCells without y1
- [x] Address Jim & Luke's comments 
- [x] Better handling of import errors (it fails silently w/o reporting the error)
- [x] CodeRunning cell indicator
- [x] new control for code running 
- [x] bug: insert cell reference in JS code editor

## Out of scope for this PR
* Typescript support (requires a bit of work around packaging)
* better console colors/highlighting


luke-quadratic:
# First independent review 

## Blocking items 

### No errors 
As described, bug of having no errors in my console 

### Add d3.js 
Visualizations, data wrangling, etc. 

### Make sure getCells() is sufficiently similar to Python 

### Arrays in console 
As image shows 
https://github.com/quadratichq/quadratic/assets/146771258/9945aaf9-cc27-49b3-aca6-d7b191f6ad6a

### Objects should be output in console and sheet
https://github.com/quadratichq/quadratic/assets/146771258/5ca53213-cf56-4a53-ab77-ab859b015ea6

### Code snippets 
Let's make sure we create code snippets that sufficiently cover the base use-cases so we know the minimal stuff actually works 

### Press '/' to code should disappear
Not recognizing js currently

### Freezing by trying to pick other code editor 
To recreate select other code cells from currently open code cell 

### JS image 
For cell type selection and top left of editor use following image 
https://github.com/quadratichq/quadratic/assets/146771258/bd8fc2cb-5a84-4b38-b486-a8127b1204a7

### AI assistant for JavaScript 
@davidkircos presumably a near clone for Python, @luke-quadratic will provide text file of docs for fine-tuning 

### Basic docs
Add docs

jimniels:
This code is bad (its incorrectly using `Object.entries()`)

```js
let res = await fetch("https://jsonplaceholder.typicode.com/todos/1");
let json = await res.json();

let data = Object.entries((key, value) => {
    console.log([key, value]);
    return [key, value];
});

return data
```

If you try to run that, it will result in an error that makes the code editor become unresponsive to running code. You can still type and it looks like you're running it, but you're not. Only once you open the console do you see you got stuck

This is the proper code that should run

```js
let res = await fetch("https://jsonplaceholder.typicode.com/todos/1");
let json = await res.json();

let data = Object.entries(json).map((key, value) => {
    console.log([key, value]);
    return [key, value];
});

return data
```

CleanShot 2024-04-16 at 13 22 54@2x

jimniels:
JavaScript needs first class support as a cell type in the app (e.g. in the cell type menu not looking disabled, as well as a matching icon in the code editor for the "JS" code type)

CleanShot 2024-04-16 at 13 27 28@2x

CleanShot 2024-04-16 at 13 27 39@2x

I can assign this to @jimniels to do

jimniels:
Nice-to-have: prettier that formats the code when you save/run any code cell changes

jimniels:
CleanShot 2024-04-16 at 13 31 42@2x

jimniels:
Using `console.log` produces unexpected results for primitive data types in javascript, e.g. logging and object just shows `[object Object]` and logging an array (or nested array) doesn't show the values as being in an array (`[1,2,3]` logs as `1,2,3`)

See screenshot for basic example discrepancies

CleanShot 2024-04-16 at 13 36 14@2x

jimniels:
The UI that indicates what is being returned is different in javascript than it is in Python.

For example, here's the UI for a javascript cell that returns a string

CleanShot 2024-04-16 at 13 53 16@2x

And here's a cell that returns a string in python

CleanShot 2024-04-16 at 13 52 49@2x

Note that 1) the line number is missing, and 2) the data type isn't highlighted

jimniels:
People are probably going to forget `await` a lot and encounter an issue like this:

CleanShot 2024-04-25 at 13 50 20@2x

And there's no help that all they need to do is add `await`. Even the console says it's an object:

CleanShot 2024-04-25 at 13 50 49@2x

Would be great if we could throw a warning and tell people that what that is is a promise

jimniels:
Minor: some of the UI has the return statement as a sentence with full punctuation (like this JS cell with a period)

CleanShot 2024-04-25 at 13 55 33@2x

While others do not (like this python output with no period)


CleanShot 2024-04-25 at 13 55 38@2x

Whatever we do, we should be consistent. I’d probably lean towards not making them full sentences (so leave off the trailing period)

jimniels:
Updating the range of `getCells` when using JavaSrcript doesn't seem to update the dashed outline on the grid (whereas it does if you're using python)

https://github.com/quadratichq/quadratic/assets/1316441/d0236fea-949b-4153-9cb1-6fd4b1321336

jimniels:
Logging a new line results in a quote mark?

CleanShot 2024-04-25 at 14 30 40@2x

jimniels:
Not passing a third arg to the `getCells` function still works. The function signature says that third param is required, but if you don't provide it, everything still works.

```
// This is wrong, 3 params are required, it should throw an error or something
const oneCell = await getCells(0,0)
```

CleanShot 2024-04-25 at 14 49 35@2x

luke-quadratic:
Why is the output of this not 1? 

CleanShot 2024-04-25 at 16 03 56@2x

luke-quadratic:
Need parity with Python across the board - in the case of references this means not just getCell and getCells should be supported 
Should also support cell, c, and cells.

CleanShot 2024-04-25 at 16 05 41@2x

jimniels:
For `getCellsWithHeadings` you get back an array of objects with keys for the headings, but you can't just return them right back to the sheet untouched. This feels a bit...strange.

For example, if you use  `getCellsWithHeadings ` to get some data like this:

`[{foo: 1, bar: 2 }, {foo: 10, bar: 20}]`

You can't return that back to the sheet. 

To return data like that, you'd have to do an array of arrays, e.g.

`[['foo', 'bar'], [1, 2], [10, 20]]`

CleanShot 2024-04-25 at 16 54 13@2x

Not sure if we want to standardize something where if you return an array of objects, it'll try to put that on the sheet with the keys as col headings, e.g.

`return [{foo: 1, bar: 2 }, {foo: 10, bar: 20}]`

would result in

CleanShot 2024-04-25 at 16 58 03@2x

luke-quadratic:
for chart let's just call it returned a chart, user is indifferent/doesn't know what a blob is, image, etc. 

CleanShot 2024-04-30 at 09 46 56@2x

luke-quadratic:
Seemed like @jimniels was experiencing this as well but the app regularly just stops working, as in you can move about the sheet but you can't open the editor, can't run things, etc. 

This happens after a minute or two of playing around usually.

luke-quadratic:
Should this match Python? rel_cell vs relCell
CleanShot 2024-04-30 at 10 46 38@2x

luke-quadratic:
rogue arrows in editor 

CleanShot 2024-04-30 at 11 34 59@2x

luke-quadratic:
This is how cell ranges are treated, shouldn't it just be like this when it's 1-D: 

array: [ 
0: value1
1: value2
2: value3
] 

when it's 2D
array: [ 
0: value1
1: value 2
etc
],
[
0: value1
etc.
]

CleanShot 2024-05-01 at 09 22 49@2x

davidfig:
> Why is the output of this not 1?
> 
> CleanShot 2024-04-25 at 16 03 56@2x

Your await is in the wrong place. Needs to be in `const x = await getCell(...)` line.

jimniels:
I'm assuming this is a known thing, but the default chart that gets generated is low resolution on my MacBook display.

CleanShot 2024-05-03 at 16 05 01@2x


Can we control the image that gets generated? can we make it 2x? e.g. draw it at 800x600 but size it to 400x300

jimniels:
JS is lacking the same mechanisms as Python for inspecting your code cell run, e.g. cancelling a run, seeing that it's in progress, etc. The "play" button doesn't even go disabled while code is executing.

This is fine for a lot of synchronous code, but when you start doing async stuff (like fetching modules over the network) it feels really weird that there's no feedback that _something_ is happening. 

In this example, I slowed down the network, but you can see how I hit "Play", the button remains enabled, there's no indication that anything is happening, and then after a few seconds the module gets loaded, the code executes, and you see the new value show up in the sheet (and the new output in the console)

https://github.com/quadratichq/quadratic/assets/1316441/bad439ac-fd28-4c3d-bdf4-4623c2996b64

Are we ok just shipping this like this? without the same mechanisms we have in python for cancelling execution, showing when something is executing, etc.?

quadratic-client/package.json
```diff
@@ -68,6 +69,7 @@
     "semver": "^7.5.4",
     "tailwind-merge": "^2.0.0",
     "tailwindcss-animate": "^1.0.7",
+    "typescript": "^5.4.4",
```
jimniels, line 72: is there a reason we're adding typescript here to `quadratic-client` package? it's already a dep in the top-level `package.json`

quadratic-client/src/app/gridGL/UI/Cursor.ts
```diff
@@ -149,12 +149,7 @@ export class Cursor extends Graphics {
         editor_selected_cell.x === cell.x &&
         editor_selected_cell.y === cell.y
       )
-        color =
-          editorInteractionState.mode === 'Python'
-            ? colors.cellColorUserPython
-            : editorInteractionState.mode === 'Formula'
-            ? colors.cellColorUserFormula
-            : colors.cursorCell;
+        color = colors.cursorCell;
```
jimniels, line 152: It sounds like @davidkircos [may not be sold on this](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1713478308835169), so we may want to pull it out — or find another solution — before merging

jimniels:
Minor: any occurrence of "Javascript" in the UI should be represented as "JavaScript" 

CleanShot 2024-05-03 at 16 37 09@2x

quadratic-client/src/app/gridGL/loadAssets.ts
```diff
@@ -44,6 +44,7 @@ export function loadAssets(): Promise<void> {
     // CellsMarker
     Loader.shared.add('/images/formula-fx-icon.png');
     Loader.shared.add('/images/python-icon.png');
+    Loader.shared.add('/images/JavaScript-logo.png');
```
jimniels, line 47: minor: seems like we should keep our casing consistent between these images, e.g `javascript-logo.png`

jimniels:
We should update the cell type menu for JavaScript so it reads like the others and includes a link to the JS docs

Example:

CleanShot 2024-05-03 at 16 45 46@2x

quadratic-client/src/app/ui/menus/CodeEditor/CodeEditorBody.tsx
```diff
@@ -153,9 +166,12 @@ export const CodeEditorBody = (props: Props) => {
             horizontal: 'hidden',
           },
           wordWrap: 'on',
+
+          // need to ignore unused b/c of the async wrapper around the code and import code
+          showUnused: language === 'Javascript' ? false : true,
         }}
       />
-      {language === 'Python' && (
+      {['Python', 'Javascript'].includes(language as string) && (
```
jimniels, line 174: Just a thought: would be nice to have a standard definition for these keys. Just looking at this PR, I see `javascript`, `Javascript` (and the UI display of `JavaScript`) and have a hard time keeping track of when to use which ha.

package.json
```diff
@@ -50,7 +50,8 @@
     "prisma:migrate": "cd quadratic-api && npm run prisma:migrate",
     "docker:up": "docker compose up -d --wait && npm run prisma:migrate --workspace=quadratic-api",
     "docker:down": "docker compose down",
-    "docker:down:kill-volumes": "docker compose down -v"
+    "docker:down:kill-volumes": "docker compose down -v",
+    "copy:esbuild": "cp -f node_modules/esbuild-wasm/esbuild.wasm quadratic-client/public/"
```
ddimaria, line 54: Move this to `quadratic-client`?

luke-quadratic:
CleanShot 2024-06-26 at 19 13 37@2x

luke-quadratic:
Latest build is missing the corner little box for dragging to copy the cell 
CleanShot 2024-06-27 at 12 46 07@2x

davidkircos (approved):

PR #1276, opened 2024-04-10
fix: HMR

jimniels:
Hot module reloading on the sheet is broken.

Load the app, change some code in a component, then save the file. The app will break and you'll see something like this:

CleanShot 2024-04-10 at 14 06 33@2x

Turns out it's this piece of code. It's passing `node` to `htmlCellsHandler.attach` which expects it to be a div, but in this cased't isn't always a div. Sometimes it undefined (like on initial render).

Changing to this fixes it.


davidfig:
And this fixes the issue? I thought the HMR issue was that the python web worker received a window.reload() command, even though workers don't have window defined.

Issue #732, opened 2023-10-18
store sheet state in URL


jimniels:
Today if you have multiple sheets, you lose the state of the sheet you're looking at if you refresh.

For example, if I'm on a sheet with tabs: [Foo][Bar] and "Foo" is the active sheet, if I switch to "Bar" that navigation is not tracked in the URL. So if I then refresh the page, I will no longer be looking at "Bar" but at "Foo" because by default it loads the first sheet. 

Not only does refreshing lose your state in the sheet, but there's also no way to share a link with someone to a particular sheet.

Again with the example of the sheets [Foo][Bar], if I made that sheet public and linked someone to it, I wouldn't be able to link them directly to sheet "Bar". It would always load "Foo"

We should have a query param that updates as the user navigates so, at minimum, you can link someone to the sheet you're looking at (and if you reload, you don't lose the sheet you were looking at). 

Example: `https://app.quadratichq.com/file/:uuid?sheet=:sheetId`

PR #1281, opened 2024-04-15
Url params

davidfig:
Closes https://github.com/quadratichq/quadratic/issues/732

- [x] cursor and sheet: `?x=0&y=1&sheet=name` 
- [x] code editor (if defined, editor is open): `?codeX=0&codeY=1&codeSheet=name`
- [x] viewport rect `?vx=1&vy=2&vw=1000&vh=1001`
- [x] redesign for the share dialog so it includes an option to include these variables (@jimniels)
- [x] discussion on whether to update the URL. Two options: (1) clear the URL after using these values; or (2) keep the URL updated with the latest information so you can always refresh or copy the URL to get back to your current state

Note: `name` for 1 and 2 is uri encoded and the comparison uses toLowerCase (so there's a chance it will conflict if you have two sheets with the same name except for cases)


jimniels:
When I move the cursor around, the URL updates. When I change sheets, the cursor updates. However, if I have the code editor open, those things aren't true. When the code editor is open, updates to positioning in the URL freeze.

So if I have `[Sheet 1, Sheet 2, Sheet 3]` and I go to sheet 2, open a code cell, then navigate to sheet 3 and hit refresh, I am not in sheet 3. I am back in sheet 2 with the code editor open.

That seems odd because it's not what I was just looking at.

The point of preserving the state is to bring me back (as close as possible) to what I was just looking at, the state I was in. We can't always preserve _all_ state, so maybe we need better priority as to which state is saved.

I feel like sheet state takes priority over code editor state. So if we have to drop something, I would say drop that the code editor was open.

https://github.com/quadratichq/quadratic/assets/1316441/e91ca400-33a1-403a-a6ed-565faef17593

That said, the description in the PR says that it's going to preserve both `x` and `codeX` but that's not what I'm seeing in the preview branch.

quadratic-client/src/app/gridGL/interaction/viewportHelper.ts
```diff
@@ -72,3 +72,18 @@ export function moveViewport(options: { center?: Coordinate; topLeft?: Coordinat
 
   pixiApp.viewportChanged();
 }
+
+export function getShareUrlParams(): string {
+  let url = `x=${sheets.sheet.cursor.cursorPosition.x}&y=${sheets.sheet.cursor.cursorPosition.y}`;
```
ddimaria, line 77: If we start adding to the size of the url, we should consider a [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) object.

quadratic-client/src/app/grid/controller/Sheets.ts
```diff
@@ -138,13 +138,13 @@ class Sheets {
...
-      pixiApp.loadViewport();
+      pixiApp.viewport.loadViewport();
     }
   }
 
-  getSheetByName(name: string): Sheet | undefined {
+  getSheetByName(name: string, urlCompare?: boolean): Sheet | undefined {
     for (const sheet of this.sheets) {
-      if (sheet.name === name) {
+      if (sheet.name === name || (urlCompare && decodeURI(name).toLowerCase() === sheet.name.toLowerCase())) {
```
ddimaria, line 147: I don't think we're comparing case elsewhere for sheet name equality.  Is the b/c the user can manipulate a URL?

quadratic-client/src/app/gridGL/pixiApp/PixiApp.ts
```diff
@@ -385,6 +314,21 @@ export class PixiApp {
...
+    const cellsSheet = sheets.getById(sheetId);
+    if (cellsSheet) {
+      cellsSheet.cursor.viewport = {
+        x: this.viewport.x,
+        y: this.viewport.y,
+        scaleX: this.viewport.scale.x,
+        scaleY: this.viewport.scale.y,
+      };
+    } else {
+      console.error('Sheet not found in urlViewportLoad');
```
ddimaria, line 329: Maybe putting the error in the getById function would remove the need to have this type of conditional.

quadratic-client/src/app/gridGL/pixiApp/Viewport.ts
```diff
@@ -0,0 +1,81 @@
...
+    super();
+    this.drag({
+      pressDrag: true,
+      wheel: false, // handled by Wheel plugin below
+      ...(isMobile ? {} : { keyToPress: ['Space'] }),
+    })
+      .decelerate()
+      .pinch()
+      .clampZoom({
+        minScale: 0.01,
```
ddimaria, line 20: Magic numbers, might be good to make them constants.

quadratic-client/src/app/ui/menus/CodeEditor/CodeEditor.tsx
```diff
@@ -71,6 +71,16 @@ export const CodeEditor = () => {
   // TODO(ddimaria): leave this as we're looking to add this back in once improved
```
ddimaria, line 71: Is this comment still relevant...seems like the code it was above was moved or deleted.

ddimaria (approved):
Code looks good.  I'll let @davidkircos or @jimniels  validate the UI.

ddimaria:
Screenshot 2024-05-06 at 7 56 23 AM

@davidfig should this move to an issue if not addressed in this PR?

quadratic-client/src/app/gridGL/pixiApp/PixiApp.ts
davidfig: Actually, I don't think we want this error as it's possible the sheet was deleted since the last load.

PR #1293, opened 2024-04-19
refactor: quadratic-client file reorganization

jimniels:
Re-organize `quadratic-client` using the natural split inherent in the nature of the app: the dashboard and not the dashboard (the "app")

("app" sounds inclusive of the dashboard, so it's not the best word, but felt right here so I'm using it)

Everything that was app-specific under `src/` got moved into `src/app/` and (most things) that were shared between the app the dashboard got moved into the root `src/shared` so you have a directory structure that looks like this:

```
quadratic-client/src/
  dashboard/
  app/
  shared/
```

A slightly longer example given what's in there today:

```
quadratic-client/src/
  dashboard/
    components/
    ...
  app/
    atoms/
    grid/
    gridGL/
    helpers/
    hooks/
    multiplayer/
    schemas/
    theme/
    ui/
    utils/
    web-workers/
  shared/
    api/
    constants/
    components/
    shadcn/
  routes/
  router.tsx
```

### Notes

- `quadratic-core` and the rust-related stuff required I look for all occurrences of `quadratic-client/src` and replace them with `quadratic/src/app` so those get generated in the right place
- tsconfig files in `web-workers` had to be updated because they got moved _down_ a directory (into `src/app`), which means their paths to the tsconfig they were extending had to be updated
- A lot of paths that had to be updated I updated using the non-relative syntax, e.g. `@/app/ui/MyComponent` instead of `../../../ui/MyComponent`. We can do this more in the app as we work with each file.


.vscode/settings.json
```diff
@@ -63,5 +63,6 @@
   "files.associations": {
     "*.grid": "json"
   },
-  "jest.rootPath": "quadratic-client"
+  "jest.rootPath": "quadratic-client",
+  "typescript.preferences.importModuleSpecifier": "non-relative"
```
jimniels, line 67: This gives us non-relative import paths with typescript

CleanShot 2024-04-19 at 17 10 09@2x

jimniels:
If I run `node dev` everything seems to compile fine, but when I go to the app in the browser I get this:

CleanShot 2024-04-19 at 17 10 52@2x

davidkircos (approved):

Issue #1117, opened 2024-02-22
Find in sheet (ctrl + F) hidden behind code editor 


luke-quadratic:
If code editor is open one can't attempt to Find in Sheet as it's presumably hidden behind the editor

jimniels:
I looked into this a bit and got a rough sense for what it's going to take to get it done. 

Gonna leave some notes here for when I'm able to come back to it

---

Primary conecern is going to be with top-level application layout and being able to make it clear in the code how that's working.

```jsx
// QuadraticUI.tsx

<TopBar />
<AppLayout />
<BottomBar />
{Dialogs}


// AppLayout.tsx

<div>
  <div>
    <FileUploadWrapper>
      <QuadraticGrid />
      <Following />
    </FileUploadWrapper>
    {!presentationMode && <SheetBar />} 
  </div>

  {showCodeEditor && 
    <div>
      <div>
        <CodeEditorBody />
      </div>
      <div>
        <CodeEditorPanel />
      </div>
    </div>
  }

  OR


  {showCodeEditor && <div><CodeEditor /></div>}
</div>
```

It's also worth considering, though not necessarily implementing, how we will want to handle these panels across mobile (since that is on the horizon).

I sketched out ideas for this on paper.

![Image](https://github.com/quadratichq/quadratic/assets/1316441/e7c77ea9-1774-47f4-a672-66430afc3273)

PR #1307, opened 2024-04-23
fix: don't hide code editor

jimniels:
Fixes the issue where, if you resize your viewport <1000px, the code editor would become hidden.

Now if the code editor is open and you resize your window to something smaller than your viewport, that's on you.

We can probably look at making this better in #1117


ddimaria:
Works as expected.  I'm not sure this is new, but the console/AI parts seems a little narrow:

Screenshot 2024-04-24 at 9 49 35 AM

ddimaria (approved):

jimniels:
Yeah that's definitely not a new thing. It's the min width for that panel. Could probably be increased but that's for a separate PR.

PR #1322, opened 2024-04-26
fix: show sheet name in code editor header

jimniels:
Because the code editor stays open, even when you switch between code cells, we need to show the sheet name when the code editor is open for a cell that's not in the currently active sheet.

This PR does that. It also refines aspects of the code editor header, including:

- Fixes problem where code editor header height was different across different cells
- Moves the unsaved indicator over the file to account for sheet names being variable in length
- Cell type head is hoverable and displays the cell type as well as a note that changes are unsaved (where relevant)
- Fixes problem where double clicking in canvas highlights tab text

https://github.com/quadratichq/quadratic/assets/1316441/797a0396-4234-40b6-b8a6-e1facfea25a5

CleanShot 2024-04-26 at 12 15 22@2x

### DIFF
--- a/quadratic-client/src/app/gridGL/pixiApp/pixiApp.css
+++ b/quadratic-client/src/app/gridGL/pixiApp/pixiApp.css
@@ -3,4 +3,6 @@
   height: 100%;
   outline: none;
   -webkit-tap-highlight-color: rgba(255, 255, 255, 0);
+  -webkit-user-select: none;
+  user-select: none;
 }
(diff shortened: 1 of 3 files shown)


quadratic-client/src/app/gridGL/pixiApp/pixiApp.css
```diff
@@ -3,4 +3,6 @@
   height: 100%;
   outline: none;
   -webkit-tap-highlight-color: rgba(255, 255, 255, 0);
+  -webkit-user-select: none;
+  user-select: none;
```
davidkircos, line 7: Random but double clicking in the top bar selects the "Share" button text. Maybe we should do this same CSS there?

PR #1334, opened 2024-05-06
Code editor paneling

davidfig:
Changes layout so `<CodeEditor />` is a flex sibling of the pixi canvas, instead of positioned over it.


jimniels:
When you try to resize beyond the maximum bounds, there's still like a ~100px it's still resizing on the right (even though it's not on the left)

Happens when the console/AI assistant is expanded out

https://github.com/quadratichq/quadratic/assets/1316441/4b5889c5-5384-4a25-a381-570220e0414f

As well as when it's collapsed.

https://github.com/quadratichq/quadratic/assets/1316441/ea20c8c4-1b32-4980-b441-1f5b17e5401d

jimniels:
A similar thing happens when you resize below the minimum size and you can end up in a position where it starts resizing before you hit the resize threshold.

https://github.com/quadratichq/quadratic/assets/1316441/9fb1d8aa-4c66-4acb-94a3-8b96d1bfa849

jimniels (dismissed):
FWIW: [I started thinking about this](https://github.com/orgs/quadratichq/projects/1/views/3?pane=issue&itemId=54175499) in the context of how we'd do it so this works on mobile but probably not worth biting off yet.

That said, there are a few problems when you're resizing larger or smaller than the min/max sizes.

jimniels:
The dividing/resizing line has disappeared when the console/ai tabs are under the code

CleanShot 2024-05-07 at 09 31 00@2x

It's actually still there, if you hover the area you'll find it, but it's visibly hidden.

davidkircos (approved):

jimniels:
I'm still seeing issues where resizing is happening even when your cursor isn't over the resize control


https://github.com/quadratichq/quadratic/assets/1316441/0174603a-ce93-4e05-9c5e-4a77d643f583

PR #1345, opened 2024-05-11
Improvements to floating menu

### LINKED ISSUE #1343: bug: FloatingContextMenu overflows the viewport on the right side
FloatingContextMenu overflows the viewport on the right side, even when sufficient space is available for the whole component.

[Screencast from 11-05-24 08:30:07 AM IST.webm](https://github.com/quadratichq/quadratic/assets/54364088/a359b5f2-7451-442f-9368-6a90985ccf68)

davidfig:
Closes #1343 

- [x] floating menu properly positions itself relative to the right side of the canvas
- [x] floating menu fades in and out when changing states


quadratic-client/src/app/ui/menus/ContextMenu/FloatingContextMenu.tsx
```diff
@@ -106,63 +110,92 @@ export const FloatingContextMenu = (props: Props) => {
...
-    } // left
+    }
+
+    // right
+    else if (
+      menuDiv.current.offsetWidth < window.innerWidth &&
+      x > container.offsetLeft + menuDiv.current.offsetWidth - 35
+    ) {
+      x = container.offsetLeft + menuDiv.current.offsetWidth - 35;
+    }
```
AyushAgrawal-A2, line 191: @davidfig 
This calculation is incorrect as shown in the screencast:
[incorrect.webm](https://github.com/quadratichq/quadratic/assets/54364088/d731dd2c-8855-45d2-8fc0-98a2b7bb4269)

AyushAgrawal-A2:
Hey I fixed this error in my PR #1344, seeing as there are many additional fixes in this PR, shall I close it ?

quadratic-client/src/app/ui/menus/ContextMenu/FloatingContextMenu.tsx
AyushAgrawal-A2, line 191: @davidfig 
This calculation is incorrect as shown in the screencast:
[incorrect.webm](https://github.com/quadratichq/quadratic/assets/54364088/2210c44c-2a57-48a1-a61f-b936e5132f01)

AyushAgrawal-A2:
Hey I fixed this error in my PR #1344, seeing as there are many additional fixes in this PR, shall I close it ?

Also, the issue created was 1343, you have mentioned `Closes 1344` in PR description which is my PR fix.

davidfig:
Thanks. I fixed the right offset. I didn't notice your PR until you mentioned it. That would be easier :)

AyushAgrawal-A2:
@davidfig 
1. Now when there is insufficient space for the floating menu, it will overflow on the left. The common function are on the left of the menu, these are now hidden. These are probably the functions which the user wants.
image

   To address this, I used Math.max to never allow left overflow.
image

2. The left and top offset are needed to be larger i.e. 35px as they have the x, y axis labels. These labels are not present on the right side. Using 35px offset for right felt like wasting space, thats why I used 15px in my PR.

Also, I have closed my PR in favour of this one. Mentioning these point if you would like to consider.

davidfig:
I did the math for the column/row header to add its actual size based on its visibility. I think the original implementation used 35px to control for those headers, but this is a better solution. 

I also used your `Math.max` to ensure that the left side was always visible. Longer term, we will have to wrap the floating menu when we add mobile support. But for now, it should be okay. 

Thanks for the help!

davidkircos (approved):

PR #1346, opened 2024-05-11
feat: multiline ai assist prompts

### LINKED ISSUE #1210: feature: multiline ai assistant prompts
When you prompt the AI with something, today it's a single line input:

CleanShot 2024-03-21 at 13 49 30@2x

Whereas in clients like chatGPT you have an input that starts as a single line, 

CleanShot 2024-03-21 at 13 51 02@2x

but as you add more data, it goes to multiple lines (with a max row height)

CleanShot 2024-03-21 at 13 51 23@2x

Super useful when you're pasting code or have prompts that are longer than a single sentence

AyushAgrawal-A2:
closes #1210 

Changes:
- Added Textarea component which can auto grow unto a max height, used this for AI assist promts. Max height is currently set to 120px, configurable via props.
- Fixed auto focus of prompt input on selecting AI assistant tab.

[ai assist prompt input.webm](https://github.com/quadratichq/quadratic/assets/54364088/98940367-27ce-40a9-a5c4-e304bebcbca5)


quadratic-client/src/shared/shadcn/ui/textarea.tsx
```diff
@@ -0,0 +1,55 @@
...
+        }}
+        onKeyDown={(event: React.KeyboardEvent<HTMLTextAreaElement>) => {
+          if (event.key === 'Enter' && !event.shiftKey) {
+            onKeyDown?.(event);
+            if (autoHeight) {
+              adjustHeight();
+            }
+          }
+        }}
+        {...props}
```
jimniels, line 48: There's a lot of duplication going on in this base component. `authHeight`, `onChange`, and `onKeyDown` all seem to be re-implementing what you're already doing with `<Textarea>` in the `<AiAssistant>` component?

This component should be a very generic textarea input, so for example hardcoding shift + ENTER to fire the `onKeyDown` seems problematic. Somebody could easily need to pass a different `onKeyDown` function that fires regardless of whether that keyboard combination happens.

Seems like `onChange` and `onKeyDown` should be props that spread on the base `textarea`.

I would think usage of this component should be something like:

`<Textarea autoResize={true} {...props} />`

And this component will handle the ref, resizing the input (if the prop is set), etc.

jimniels: Or maybe we could just remove the `onKeyDown` handler entirely and let the spread syntax take care of assigning that event handler? 

In that case, the existing `onChange` would handle updates to the component that re-render it's size.

jimniels (changes requested):

quadratic-client/src/shared/shadcn/ui/textarea.tsx
AyushAgrawal-A2: I don't believe it is duplicate code. I'll explain the purpose of event handlers for both the components:

In the `TextArea` component,
 - `onChange` - this triggers auto height adjusting when text-area value changes
 - `onKeyDown` - the submit action i.e. pressing Enter _without_ Shift key, does not trigger the `onChange` event. The value of text-area is emptied and the height needs to be re-adjusted. This requires additional `onKeyDown` event here. Otherwise only having `onChange` was sufficient.

In the `AiAssistant` component,
 - `onChange` - updates the local prompt state. This is controlled input.
 - 'onKeyDown` - triggers submit prompt and clears prompt state, when Enter _without_ Shift key is pressed.

The logic is different in both the components and nothing is being re-implemented. 

I am sorry but I am not able to correlate your feedback to the implementation logic, please correct me if I am wrong here.

quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
```diff
@@ -6,7 +6,7 @@ import { AI } from '@/app/ui/icons';
 import { authClient } from '@/auth';
 import { useRootRouteLoaderData } from '@/router';
 import { apiClient } from '@/shared/api/apiClient';
-import { Input } from '@/shared/shadcn/ui/input';
+import { Textarea } from '@/shared/shadcn/ui/textarea.js';
```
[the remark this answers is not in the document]
AyushAgrawal-A2: Ah.. this project does not require file extensions in imports... this was default on my IDE
p.s.: import will always be `.js`, even for `.tsx` file as it gets transpiled into `.js` eventually

quadratic-client/src/shared/shadcn/ui/textarea.tsx
```diff
@@ -0,0 +1,55 @@
...
+        }
+      });
+    }, [textareaRef]);
+
+    return (
+      <textarea
+        ref={textareaRef}
+        className={cn(
+          'flex h-8 max-h-[120px] min-h-8 w-full resize-none rounded-md border border-input bg-transparent px-3 py-1 text-sm shadow-sm transition-colors file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50',
+          `max-h-[${maxHeight || 'none'}]`,
```
jimniels, line 31: Does this actually work? My understanding is tailwind won't work with dynamic class names. You'd have to turn this into a style attribute instead or something

https://tailwindcss.com/docs/content-configuration#dynamic-class-names

jimniels: huh, it looks like it does. ok let's keep it.

AyushAgrawal-A2: yup, my bad. I have fixed it.

jimniels (approved):
I added a few additional fixes, tweaks to this, including:

- CMD+Enter submits the textarea, with Enter and Shift+Enter now inserting a line. This aligns the input more broadly with web app textarea inputs (like twitter) as well as our own code editor which uses CMD+Enter to "submit" and run the code while Enter just adds a new line. Added a shortcut to the tooltip to indicate this to the user

CleanShot 2024-05-21 at 10 10 44@2x

- Multi-line user inputs were not displaying as multi-line messages from the user who typed them. Now they do.

CleanShot 2024-05-21 at 10 00 23@2x

- "Send" button aligns to the bottom of the textarea as it expands in height (so does loading indicator)

CleanShot 2024-05-21 at 09 43 52@2x

jimniels:
Will merge this once all tests pass...

PR #1347, opened 2024-05-12
fix: inequality comparison issue using <>

### LINKED ISSUE #1328: Inequality comparison issue using <> logic
When attempting to use the not equal to logic "<>" in the formula portion of quadratic we get the following error:

Example Code: IF(C23 <> 1, 11, 22)
Result:
ERROR: colum 14 to 15: Expected expression

We think this is due to quadratic reading the less than sign (<) seperately from the greater than sign (>) rather than holistically as (<>)

AyushAgrawal-A2:
closes #1328

Changes:
- Updated regex to match `<>` as well

image

image


ddimaria:
I checked and there are tests already associated with this feature, so not sure why they were passing 🤔 

I validate the UI and confirmed the issue is fixed.

ddimaria (approved):
Thanks for the contribution @AyushAgrawal-A2!

PR #1357, opened 2024-05-16
Add button to insert code references into code editor

davidfig:
- [x] button to insert either relative or absolute (for different sheets) for the code editor
- [x] fixed a visual bug with Python cell references when referencing a different sheet


jimniels:
My thoughts on the current state of this branch:

## Problem

The initial implementation works functionally, but just sticks icons in the code editor header 

image

This feels problematic for a few reasons

1. We start to run out of room in that header real quick
2. It's hard to communicate the idea of "Insert an absolute/relative cell reference to the current selection" with icons only
3. The action can be so far away from the control itself (e.g. you're way over on the right making a selection and you have to move way over to the right to actually insert it)

image

## Possible Solutions

We don't have to do _just_ one of the below. We implement both solutions in different places.

### Formatting bar

Put the controls to insert a selection in the formatting bar. This would mean that once you make a selection, you get the option to insert a reference to the current selection that is in close proximity to the action you just completed (making a selection).

This could be in the dropdown:

image

Or it could be new icons that appear _only_ when the functionality is possible, e.g. once you've made a selection (or right-click for a single cell) _and_ the code editor is open.

We could use a single icon with a modifier key:

CleanShot 2024-06-05 at 11 01 38@2x

Or two icons, so it's only one click for either option:

image

Or some kind of textual dropdown:

CleanShot 2024-06-05 at 10 57 37@2x

Each of these have their benefits and drawbacks:

- One click or two clicks to do it
- Textual label (more clear) icon (less clear, but saves space and can become more clear over time)
- Clarity

### Editor

Put the ability to insert a cell reference in the context menu. Then the user can right click to insert a reference to the current selection wherever they want in the code editor (and we could create keyboard shortcuts so you just make a selection, then like `CMD+[` or `SHIFT+CMD+[`.

CleanShot 2024-06-05 at 11 05 23@2x

Or we could add it back to the header via some kind of single icon toggle:

image

Again, all of these have their benefits and drawbacks. It's helpful to see contrasting options.

### Command palette

Make a selection, then trigger the command palette and (if the code editor is open) we have two options for inserting a reference to the current selection

CleanShot 2024-06-05 at 11 31 04@2x

davidfig:
Add: first_row_header = False as default to cell_reference

davidkircos:
Two items:
- Update icon `HighlightAlt`
- Edit text

quadratic-client/src/app/ui/menus/CodeEditor/CodeEditorRefButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+      // we do not yet support multiple multiCursors for inserting cell references
+      if ((sheets.sheet.cursor.multiCursor && sheets.sheet.cursor.multiCursor.length > 1) || sheets.sheet.cursor.columnRow !== undefined) {
+        setDisabled(true);
+      } else {
+        setDisabled(
+          !sheets.sheet.cursor.multiCursor &&
+          editorInteractionState.selectedCell.x === sheets.sheet.cursor.cursorPosition.x &&
+          editorInteractionState.selectedCell.y === sheets.sheet.cursor.cursorPosition.y &&
+          editorInteractionState.selectedCellSheet === sheets.sheet.id &&
+          !sheets.sheet.cursor.multiCursor
```
ddimaria, line 30: This also appears on line 26

quadratic-kernels/python-wasm/quadratic_py/utils.py
```diff
@@ -13,60 +13,109 @@
...
     # Convert c((x,y), ...) cell((x,y), ...) to cells((x,y), ...)
-    code = re.sub(r"(^|\W)c(?:ell)?(\([^\(\)]*\([^\)]*\)(?:[^\(\)]*|\([^\)]*\))*\))", r"\1cells\2", code) # captures c((x,y), ...) cell((x,y), ...)
-    
-    # Wrap known async functions with "(await <function-call>)" to improve the UX
-    code = re.sub(r"(^|\W)(?:await +)?(c(?:ells?)?\((?:[^\(\)]*|\([^\)]*\))*\))", r"\1(await \2)", code) # captures c( cell( cells(
-    code = re.sub(r"(^|\W)(?:await +)?(cells\[[^\]]*\])", r"\1(await \2)", code)                         # captures cells[
-    code = re.sub(r"(^|\W)(?:await +)?(getCells?\((?:[^\(\)]*|\([^\)]*\))*\))", r"\1(await \2)", code)   # captures get_cell( get_cells(
-    code = re.sub(r"(^|\W)(?:await +)?(rel_cell\([^\)]*\))", r"\1(await \2)", code)                      # captures rel_cell(
-    code = re.sub(r"(^|\W)(?:await +)?(rc\((?:[^\(\)]*|\([^\)]*\))*\))", r"\1(await \2)", code)          # captures rc(
+    code = re.sub(
```
ddimaria, line 16: I'm assuming this is just a code formatter change.  @davidfig can you confirm?

quadratic-kernels/python-wasm/quadratic_py/utils.py
```diff
@@ -13,60 +13,109 @@
...
         elif type(value) == bool:
             return (str(bool(value)), "logical")
-        elif isinstance(value, (pd.Timestamp, np.datetime64, date, time, datetime)) or pd.api.types.is_datetime64_dtype(value):
+        elif isinstance(
+            value, (pd.Timestamp, np.datetime64, date, time, datetime)
+        ) or pd.api.types.is_datetime64_dtype(value):
             return (str(to_unix_timestamp(value)), "instant")
-        
+
         # TODO(ddimaria): implement when we implement duration in Rust
```
ddimaria, line 110: Remove these comments?  I think so, unless you have more work to do in Rust?

ddimaria (approved):

quadratic-kernels/python-wasm/quadratic_py/utils.py
davidfig: confirmed

quadratic-kernels/python-wasm/quadratic_py/utils.py
davidfig: might as well leave them. we'll one day need it

PR #1381, opened 2024-05-27
feat: toggle borders

### LINKED ISSUE #1231: feature: toggle borders
When you're applying borders, such as "Border left", if the border is already there we turn it off. If it's not present we turn it on.

This is especially useful when you have some borders present but need fine-grained control over styling a particular border. For example, imagine you have a cell with borders on all sides

```
 -----
|     |
 -----
```

But you want to remove a border on just the left side

```
 -----
      |
 -----
```

To do this today, it requires multiple steps:

1. Clear all borders
2. Apply top border
3. Apply bottom border
4. Apply right border

However, if border controls allowed "toggling" then it would only be one step:

1. Toggle border left

AyushAgrawal-A2:
closes #1231 

Changes:-
- Check if the new SheetBorder render for the rect is same as current SheetBorder render, then toggle the selected border.
- If new render is different, apply new border selection as before.

[toggle_borders.webm](https://github.com/quadratichq/quadratic/assets/54364088/7294d7f5-bf37-4d5e-957c-05fc9aef0a37)

### DIFF
--- a/quadratic-core/src/controller/operations/borders.rs
+++ b/quadratic-core/src/controller/operations/borders.rs
@@ -1,6 +1,6 @@
 use crate::{
     controller::GridController,
-    grid::{generate_borders, BorderSelection, BorderStyle},
+    grid::{generate_borders, get_rect_borders, BorderSelection, BorderStyle},
     SheetRect,
 };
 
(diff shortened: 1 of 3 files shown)


davidkircos:
@AyushAgrawal-A2, thanks for your contribution! This seems like how borders should work.

Could you also add tests to this behavior?

AyushAgrawal-A2:
> @AyushAgrawal-A2, thanks for your contribution! This seems like how borders should work.
> 
> Could you also add tests to this behavior?

@davidkircos 
Done

davidfig (approved):

PR #1400, opened 2024-05-30
fix(doc): update contributing.md

AyushAgrawal-A2:
Configuration for local dev
- Redis, Postgres and AWS(localstack) will run in docker.
- Use Auth0 developers account to login.
- All quadratic packages run locally in terminal.

Changes:
- Added `.nvmrc` file.
- Updated `control.ts` & `s3.ts` to point to docker.
- Updated all `.env.example` files to point to services in docker.
- Added representative for Auth0 variables in `.env.example` files.
- Updated setup and running instructions in `CONTRIBUTING.md`.

This is tested for Ubuntu 22.04. I am not sure if any changes are required for Mac / Windows.


quadratic-api/.env.example
```diff
@@ -0,0 +1,27 @@
+ENVIRONMENT=local
```
ddimaria, line 1: Nice, this old `_example` style naming has been a nit of mine for a while.

quadratic-api/src/aws/s3.ts
```diff
@@ -8,7 +8,7 @@ import {
   ENVIRONMENT,
 } from '../env-vars';
 const endpoint =
-  ENVIRONMENT === 'docker' ? 'http://localstack:4566' : ENVIRONMENT === 'local' ? 'http://0.0.0.0:4566' : undefined;
+  ENVIRONMENT === 'docker' ? 'http://localstack:4566' : ENVIRONMENT === 'local' ? 'http://localhost:4566' : undefined;
```
ddimaria, line 11: Note: I need to verify this before approving

quadratic-client/.env.example
```diff
@@ -0,0 +1,16 @@
+VITE_GOOGLE_ANALYTICS_GTAG=G-0000000000
```
ddimaria, line 1: @AyushAgrawal-A2 I think if you rename the file as the only activity in a single commit, git should track the rename and retain the history.

AyushAgrawal-A2: I tried:
1. Renaming (as the only activity in the commit) and then modifying contents of this file
2. Modifying contents first and then renaming in a separate commit.

Final diff of the PR still show file `.env_example` deleted and new file `.env.example` created.

AyushAgrawal-A2:
`quadratic-community` account works now, I have added its credentials in `.env.example` and updated the doc.

davidfig:
This looks great. Can you add a CLI option to stop docker from running in the isRedisRunning() check? I'm okay with it running by default, but I don't like having docker running on my machine.

AyushAgrawal-A2:
> This looks great. Can you add a CLI option to stop docker from running in the isRedisRunning() check? I'm okay with it running by default, but I don't like having docker running on my machine.

@davidfig 
Done

ddimaria (approved):
Great contribution, thanks for the submission @AyushAgrawal-A2 💯

quadratic-client/.env.example
ddimaria: I typically rename the file only as the only action in a commit to make sure git picks it up. Since you already made the change, you might have to do other operations to preserve history.  I think we're OK skipping this for now.

quadratic-api/src/aws/s3.ts
ddimaria: I'm reverting this change as it breaks my local setup.

PR #1402, opened 2024-05-31
Build client in docker

ddimaria:


quadratic-client/package.json
```diff
@@ -145,6 +147,7 @@
     "sass": "^1.69.5",
     "serve": "^14.2.0",
     "tailwindcss": "^3.3.5",
+    "typescript": "^5.4.5",
```
jimniels, line 150: Why was this added? `typescript` is in the root package.json, so does it need to be here?

ddimaria: I was having build issues without it, it might be the same issue we're getting in the next-connections PR

davidkircos (approved):

client.Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+COPY ./quadratic-rust-client/. ./quadratic-rust-client/
+COPY ./quadratic-shared/. ./quadratic-shared/
+
+# Run the packaging script for quadratic_py
+RUN ./quadratic-kernels/python-wasm/package.sh --no-poetry
+
+# Build wasm
+WORKDIR /app/quadratic-core
+RUN rustup target add wasm32-unknown-unknown
+RUN echo 'Building wasm...' &&  wasm-pack build --target web --out-dir ../quadratic-client/src/app/quadratic-core --weak-refs
```
davidkircos, line 35: This can also be done form root with `npm run build:wasm:javascript`

client.Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+# Export TS/Rust types
+RUN echo 'Exporting TS/Rust types...' && cargo run --bin export_types
```
davidkircos, line 38: `num run build:wasm:types`

client.Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+# Build the quadratic-rust-client
+WORKDIR /app
+ARG GIT_COMMIT
+ENV GIT_COMMIT=$GIT_COMMIT
+RUN echo 'Building quadratic-rust-client...' && npm run build --workspace=quadratic-rust-client
```
davidkircos, line 44: `npm run build:rust-client`

davidkircos (changes requested):

client.Dockerfile
ddimaria: That script changes directories, so can't be used.

client.Dockerfile
ddimaria: That script changes directories, so can't be used.

client.Dockerfile
ddimaria: That script changes directories, so can't be used.

davidkircos (approved):

PR #1407, opened 2024-06-04
Fixes rendering for overly long text entries 

davidfig:


quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellsTextHash.ts
```diff
@@ -253,7 +253,25 @@ export class CellsTextHash {
     this.labelMeshes.prepare();
 
     // populate labelMeshes webGL buffers
-    this.labels.forEach((cellLabel) => cellLabel.updateLabelMesh(this.labelMeshes));
+    let minX = Infinity,
```
ddimaria, line 256: Some duplication here, it might be good to have a default function for this if more duplication occurs.

ddimaria (approved):
Good looks good 💯

PR #1422, opened 2024-06-13
use gpt-4o

davidkircos:
gpt-4-32k is expensive and being deprecated


ddimaria (commented):
Code looks good, I'll wait for @luke-quadratic to approve.

luke-quadratic (approved):
so much better than the prior model

PR #1423, opened 2024-06-14
feat: teams for all

jimniels:
Everybody gets a team! Team for you! Team for you! Team for you!

giphy

## Known issues

Dashboard:

- [x] Billing - make `stripeCustomerId` optional in the schema and have the server operate on it conditionally
- [x] Move "Members" out of dialog
- [x] Move "Connections" out of dialog
  - Needs billing decisions to make firm direction on how to build this dialog (whether it's shared between app/dashboard)
- [x] Basic connections CRUD working
- [ ] Migration script needs definition & document roll-out rules @davidkircos 
- [ ] Update tests, ensure API changes work as expected (document breaking changes)
- [ ] When you come into the app for the first time, you either 1) get a team created for you, or 2) get assigned to the team via an invite
- [x] Make sure you can’t access team routes you don’t have access to
- [x] Don't hard-code the email in the team switcher

In-app:

- [x] In-app label next to file name needs to be updated
- [x] Create & duplicate always in context of current file (and private). Not available if it's a shared file

## Outstanding questions

- What do we want to do with billing?
  - Pull all billing-related stuff from the UI. Make stripcustomerid optional. Then make strip logic conditional in the API.
- How can we know "the current team" from inside the app? (e.g. when you go "create file" where is it putting it? what if you've never created yet...lots to think through)
  - Answer: For teams you have access to, we'll show "Team / File". For private files, we'll do "Private / File". And actions related to teams, like create/duplicate/connections, will not be available in-app when looking at a file that's not in a team you have edit access to.

## Things to test

Dashboard: really, _everything_ that's a button in the UI , but a few things as an example:

- Login with 0 teams, see that a default team is created for you
- Move a file back and forth between your public and private files on a team
- Creating/importing a file (creates a file _in the right spot_ based on where you are in the UI)
- Private team files are private
- Sharing works as expected, e.g.
  - Can share a public team file with someone not on a team
  - Can share a private team file with someone on and not on a team
- Switching between teams works as expected
- Leave a team, see that you're redirected to another team (or a new one is created for you)
- Permissions:
  - Make sure you can’t access team routes you don’t have access to
  - Proper things are hidden for people who have access to a team but are read-only

In-app:

- Duplicating a file duplicates to your private files in the team
- Creating a file creates a _private_ file in the team you're in
- In-app label shows correctly whether you're looking at a team file or a private draft (by you or someone else)

## Breaking changes

Conceptual breaking changes:

- You can no longer duplicate files across teams. 
  - “Duplicate" is now an action that will _only_ show up for a file if: 1) it belongs to a team you also belong to,  and 2) you have a role of EDITOR on the team. Otherwise you will never see "duplicate" as an option.
- Cloud files migration route is being removed (see [Slack convo](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1718295077571489))
- Manually go to `/files/create` to create a file no longer works. We have to know what team you're creating it on (FWIW: `/teams/:uuid/files/create` does this now)

Things that would break for people who don't reload the client after this ships:

- If the API automatically creates a team for people, then they’re gonna see a team which they then cannot access because there’s no stripe ID 
- `files.POST` requires passing a team UUID now (you can't create a file apart from a team)
- `files.$uuid.PATCH` you can't move files across teams anymore. also can't move files to your own private space apart from a team. The only way to move a file is from public -> private or private -> public within a team.

## Notes around specific

- If a user ever loads the app with access to 0 teams, we'll create one for them
- `TODO: (team-schema)` comments are things to fix once we merge & migrate
- When you remove someone from a team, their "private" files will remain assigned to them which means nobody will know they exist (and if that user is re-added to the team, they'll see them again)


quadratic-api/package.json
```diff
@@ -17,13 +17,13 @@
...
     "prisma:migrate:create": "npx prisma migrate dev --create-only",
-    "prisma:dev:reset": "npx prisma migrate reset && node scripts/seed.js",
+    "prisma:dev:reset": "npx prisma migrate reset",
     "docker:up": "docker-compose up -d && sleep 3",
     "docker:down": "docker-compose down",
     "docker:dev": "npm run docker:up && dotenv -e .env.test -- prisma migrate deploy && dotenv -e .env.test -- ts-node-dev --watch src --project ./tsconfig.json -- ./src/server.ts && npm run docker:down",
     "docker:test:dev": "npm run docker:up && dotenv -e .env.test -- prisma migrate deploy && dotenv -e .env.test -- jest --watch --detectOpenHandles && npm run docker:down",
     "docker:test:ci": "npm run docker:up && dotenv -e .env.test -- prisma migrate reset --force && dotenv -e .env.test -- jest --runInBand && npm run docker:down",
-    "script:migrateFiles": "ts-node-dev --project ./tsconfig.json -- ./scripts/migrateFiles.ts"
+    "script:migrateUsersToTeams": "ts-node-dev --project ./tsconfig.json -- ./scripts/migrateUsersToTeams.ts"
```
jimniels, line 26: TODO: the migration script still needs thought

PR #1432, opened 2024-06-18
Encrypt/decrypt typeDetails in API and store encrypted in the DB

ddimaria:
- [x] Add encrypt/decrypt functionality in quadratic-api
- [x] Test crypto functions
- [x] Incorporate encrypt/decrypt functionality when saving/reading typeDetails from the DB
- [x] Add an environment variable to hold the encryption key
- [x] Add a package.json script to generate a 32 bit key
- [x] Restricts all jobs in CI to run for 10 minutes


quadratic-api/src/data/universityDomains.ts
```diff
@@ -1,3 +1,3 @@
 // NOTE: this file is automatically generated by quadratic-api/scripts/updateUniversityDomains.js
```
ddimaria, line 1: I re-generated this.

quadratic-api/.env.docker
```diff
@@ -1,2 +1,5 @@
 ENVIRONMENT=docker
-DATABASE_URL='postgresql://postgres:postgres@postgres:5432/postgres'
\ No newline at end of file
+DATABASE_URL='postgresql://postgres:postgres@postgres:5432/postgres'
+
+# Hex string to be used as the key for enctyption, use npm run key:generate
+ENCRYPTION_KEY=eb4758047f74bdb2603cce75c4370327ca2c3662c4786867659126da8e64dfcc
```
davidkircos, line 5: Can we make the example key, which is obviously not for use in production?

davidkircos: potentially even add a check to ensure that if `ENVIRONMENT=production` that it isn't using our example key, this will be good for selfhosting

quadratic-api/prisma/schema.prisma
```diff
@@ -152,7 +152,7 @@ model Connection {
 
     // Type-specific details
     type        ConnectionType
-    typeDetails Json
+    typeDetails Bytes // JSON
```
davidkircos, line 155: Update comment 

"InitializationVector:Cyphertext"

quadratic-api/src/utils/crypto.ts
```diff
@@ -0,0 +1,55 @@
...
+  const cipher = crypto.createCipheriv(algorithm, key, iv);
+  const encrypted = Buffer.concat([cipher.update(text), cipher.final()]);
+
+  return iv.toString('hex') + ':' + encrypted.toString('hex');
+};
+
+// Encrypts the given text using the env variable ENCRYPTION_KEY
+// Store the IV with the encrypted text (prepended).
+export const encryptFromEnv = (text: string): string => {
+  const key = getKeyBufferFromEnv();
```
davidkircos, line 29: could probably set once on file load

quadratic-api/src/routes/v0/connections.$uuid.GET.ts
```diff
@@ -21,15 +22,14 @@ async function handler(req: RequestWithUser, res: Response<ApiTypes['/v0/connect
     params: { uuid },
   } = parseRequest(req, schema);
   const connection = await getConnection({ uuid, userId });
+  const typeDetails = decryptFromEnv(connection.typeDetails?.toString('utf-8') || '');
```
davidkircos, line 25: is a blank string valid here?

quadratic-api/src/utils/crypto.ts
```diff
@@ -0,0 +1,55 @@
...
+  return encrypt(key, text);
+};
+
+// Decrypts the given encrypted text using the given key.
+// The IV is prepended to the encrypted text.
+export const decrypt = (key: Buffer, encryptedText: string): string => {
```
davidkircos, line 35: should this be public or just `decryptFromEnv`?

quadratic-api/src/utils/crypto.ts
```diff
@@ -0,0 +1,55 @@
...
+export const hexStringToBuffer = (key: string): Buffer => {
+  return Buffer.from(key, 'hex');
+};
+
+// Get the encryption key from the env and convert it to a buffer.
+export const getKeyBufferFromEnv = (): Buffer => hexStringToBuffer(ENCRYPTION_KEY);
+
+// Encrypts the given text using the given key.
+// Store the IV with the encrypted text (prepended).
+export const encrypt = (key: Buffer, text: string): string => {
```
davidkircos, line 18: don't export?

quadratic-api/src/utils/crypto.ts
davidkircos, line 35: don't export?

quadratic-api/src/utils/crypto.ts
davidkircos: make it clear which functions are for developer use around the app and which ones are internal

quadratic-api/src/routes/v0/connections.$uuid.GET.ts
davidkircos: Pass to Zod for typing to ensure this is the structure we expect it to be

quadratic-api/.env.docker
ddimaria: I'm adding this to the `env-var.ts` file: 
```typescript
function ensureSampleTokenNotUsedInProduction() {
  if (NODE_ENV !== 'production') return;

  if (process.env.ENCRYPTION_KEY === SAMPLE_ENCRYPTION_KEY) {
    throw new Error(
      `Please set a unique ENCRYPTION_KEY in production. Do not use the sample value.  A new key can be generated with: npm run key:generate`
    );
  }
}
```

quadratic-api/src/env-vars.ts
```diff
@@ -1,3 +1,5 @@
+const SAMPLE_ENCRYPTION_KEY = 'eb4758047f74bdb2603cce75c4370327ca2c3662c4786867659126da8e64dfcc';
```
davidkircos, line 1: Should we define this as an ENV var?

quadratic-api/src/utils/crypto.ts
```diff
@@ -0,0 +1,47 @@
+import crypto from 'crypto';
+import { ENCRYPTION_KEY } from '../env-vars';
```
davidkircos, line 2: This function looks good. We may consider building in a key rotation system, this may be difficult later. ChatGPT suggests

--- 

Implementing key rotation for an encryption system that is already in production requires a careful approach to ensure data integrity and security. Here’s how you can implement key rotation for the given code:

### Steps for Key Rotation

1. **Introduce a Key Versioning System**:
   - Maintain multiple keys with different versions.
   - The current key version should be stored and accessible.

2. **Encrypt New Data with the Latest Key**:
   - Use the latest key for encrypting any new data.

3. **Decrypt Existing Data with the Appropriate Key**:
   - Identify the key used to encrypt the data (e.g., by storing the key version with the encrypted data) and use the appropriate key for decryption.

4. **Re-encrypt Old Data with the New Key**:
   - Gradually re-encrypt existing data with the new key to ensure a smooth transition.

### Implementation

#### 1. Key Management

Create a key management file (e.g., `keyManager.js`) to handle multiple keys and their versions.

```javascript
import crypto from 'crypto';
import { ENCRYPTION_KEYS } from '../env-vars'; // Assumes ENCRYPTION_KEYS is a JSON string with versioned keys

const keys = JSON.parse(ENCRYPTION_KEYS);

// Get the latest key version
const getCurrentKey = () => {
  const latestVersion = Math.max(...Object.keys(keys).map(Number));
  return { key: Buffer.from(keys[latestVersion], 'hex'), version: latestVersion };
};

// Get a specific key version
const getKey = (version) => {
  if (!keys[version]) {
    throw new Error(`Key version ${version} not found`);
  }
  return Buffer.from(keys[version], 'hex');
};

export { getCurrentKey, getKey };
```

#### 2. Encrypt with Key Versioning

Update the encryption functions to include the key version in the output.

```javascript
import crypto from 'crypto';
import { getCurrentKey, getKey } from './keyManager';

const algorithm = 'aes-256-cbc';

const encrypt = (key, text) => {
  const iv = crypto.randomBytes(16);
  const cipher = crypto.createCipheriv(algorithm, key, iv);
  const encrypted = Buffer.concat([cipher.update(text), cipher.final()]);

  return iv.toString('hex') + ':' + encrypted.toString('hex');
};

export const encryptWithKeyRotation = (text) => {
  const { key, version } = getCurrentKey();
  const encrypted = encrypt(key, text);
  return `${version}:${encrypted}`;
};
```

#### 3. Decrypt with Key Versioning

Update the decryption functions to handle the key version.

```javascript
import crypto from 'crypto';
import { getKey } from './keyManager';

const algorithm = 'aes-256-cbc';

const decrypt = (key, encryptedText) => {
  const [ivHex, encryptedHex] = encryptedText.split(':');

  if (!ivHex || !encryptedHex) {
    throw new Error('Invalid encrypted text format');
  }

  const iv = Buffer.from(ivHex, 'hex');
  const encrypted = Buffer.from(encryptedHex, 'hex');
  const decipher = crypto.createDecipheriv(algorithm, key, iv);
  const decrypted = Buffer.concat([decipher.update(encrypted), decipher.final()]);

  return decrypted.toString();
};

export const decryptWithKeyRotation = (encryptedText) => {
  const [version, data] = encryptedText.split(':', 2);
  const key = getKey(version);
  return decrypt(key, data);
};
```

#### 4. Re-encryption Process

Create a script to re-encrypt existing data with the new key. This script should be run as a one-time migration or periodically until all data is re-encrypted.

```javascript
import { decryptWithKeyRotation, encryptWithKeyRotation } from './encryption';
import { getCurrentKey } from './keyManager';
import { getAllEncryptedData, updateEncryptedData } from './database'; // Pseudocode for database operations

const reencryptData = async () => {
  const { version: currentVersion } = getCurrentKey();
  const dataEntries = await getAllEncryptedData();

  for (const entry of dataEntries) {
    const { id, encryptedText } = entry;
    const [version] = encryptedText.split(':', 1);

    if (version < currentVersion) {
      const decryptedText = decryptWithKeyRotation(encryptedText);
      const newEncryptedText = encryptWithKeyRotation(decryptedText);
      await updateEncryptedData(id, newEncryptedText);
    }
  }
};

reencryptData().catch(console.error);
```

### Summary

This implementation covers the following:

- Key management with versioning.
- Encryption that includes the key version.
- Decryption that retrieves and uses the appropriate key version.
- A re-encryption script to update old encrypted data with the new key.

Ensure to test this implementation thoroughly in a staging environment before deploying to production.

davidkircos (changes requested):
Left a couple of comments, no specific changes just looking for your response. Also waiting on resolving merge conflicts.

quadratic-api/src/utils/crypto.ts
ddimaria: I need to change these back to exporting for tests to work.

quadratic-api/src/env-vars.ts
ddimaria: I don't think so.  This is what we used to match against the env var `ENCRYPTION_KEY`.  The sole purpose of this var is to catch accidental use of publicly used keys in production, and has no use for developers.

quadratic-api/src/utils/crypto.ts
ddimaria: I think this is a good idea, but probably overkill for this stage.  This has been in PR for 3 weeks already and keeping this PR up to date is cumbersome.  Maybe we can merge and just update with a more sophisticated system before launch?

quadratic-api/src/routes/v0/connections.$uuid.GET.ts
ddimaria: Blank string option removed.

davidkircos (approved):

quadratic-api/src/utils/crypto.ts
davidkircos: These are only used in this file right? 

rename `decryptFromEnv` `decryptUsingEnv` and `encryptUsingEnv`?

Issue #1435, opened 2024-06-19
Excel import improvements


luke-quadratic:
Import Excel button gets used quite often - this is a tracking list for improvements to importing Excel sheets from the dashboard. 

- [ ] Import formatting 
- [ ] Import formulas 

Please feel free to edit and add to the list (and please comment any additional features you want as part of this!)

ddimaria:
We haven't done any formula imports yet.  It might be a good PoC to convert a single formula to get a sense of the LOE on doing this en mass.

davidfig:
We should be able to just import the formulas. Some of them won't work (since we don't support all the excel functions), but it should mostly work.

ddimaria:
Sadly, the excel import crate we're currently using doesn't parse formatting data.  When we start this, I'll look into other tools and/or the effort involved in parsing formatting information in a fork of our current excel import crate.

luke-quadratic:
duplicate (formulas solved, issue created for formatting #2981

PR #1441, opened 2024-06-20
feat: import formulas when importing xlsx file

AyushAgrawal-A2:
part of #1435 
ref: https://github.com/quadratichq/quadratic/issues/1435#issuecomment-2179020914

Importing formulas works and results are correct for supported functions.


davidkircos:
We should add all Excel formulas to the Rust Formulas parser so that ones that people attempt to use, but are not implemented yet can trigger a mixpanel event and we can prioritize implementation.

AyushAgrawal-A2:
> We should add all Excel formulas to the Rust Formulas parser so that ones that people attempt to use, but are not implemented yet can trigger a mixpanel event and we can prioritize implementation.

@davidkircos 
Changes:
   - Added excel's function list and check for unimplemented functions
   - Added mixpanel tracking for unimplemented error

image

Currently, tracking will trigger only when user runs a formula with a valid excel function that is not implemented yet. Importing xlsx file is a server transaction and does not trigger mixpanel tracking. Is this ok ?

quadratic-core/src/controller/execution/execute_operation/execute_code.rs
```diff
@@ -97,8 +97,8 @@ impl GridController {
         op: Operation,
     ) {
         if let Operation::ComputeCode { sheet_pos } = op {
-            if !transaction.is_user() {
-                unreachable!("Only a user transaction should have a ComputeCode");
+            if !transaction.is_user() && !transaction.is_server() {
+                unreachable!("Only a user/server transaction should have a ComputeCode");
```
ddimaria, line 101: I never liked all of this `unreachable` code.  Ideally, all of these failable functions would return results, but that's a lot of restructuring to do.  Could you try using `dbgjs!` instead and avoid the computation in the error case?

quadratic-client/src/app/gridGL/cells/CellsArray.ts
```diff
@@ -64,14 +65,37 @@ export class CellsArray extends Container {
...
+          const runError = JSON.parse(codeCell.evaluation_result) as RunError;
+          // track unimplemented errors
+          if (typeof runError.msg === 'object' && 'Unimplemented' in runError.msg) {
+            mixpanel.track('[CellsArray].updateCodeCell', {
+              type: codeCell.language,
+              error: runError.msg,
+            });
+          }
+        } catch (error) {
+          console.warn('[CellsArray] Error parsing codeCell.evaluation_result', error);
```
ddimaria, line 96: If this is an error case, convert to `console.error` so that Sentry can pick up this error.

quadratic-core/src/controller/operations/import.rs
```diff
@@ -191,12 +199,38 @@ impl GridController {
                     );
                 }
             }
+
+            // formulas
+            let formula = workbook
+                .worksheet_formula(&sheet_name)
+                .map_err(|e: XlsxError| error(e.to_string()))?;
+            let insert_at = formula.start().map_or_else(Pos::default, xlsx_range_to_pos);
+            let mut formula_compute_ops = vec![] as Vec<Operation>;
```
ddimaria, line 208: You should be able to drop the `as Vec<Operation>` as this should be inferenced by the push on line 222.

quadratic-core/src/controller/execution/execute_operation/execute_code.rs
AyushAgrawal-A2: @ddimaria 
Done

quadratic-client/src/app/gridGL/cells/CellsArray.ts
AyushAgrawal-A2: @ddimaria 
Done

quadratic-core/src/controller/operations/import.rs
AyushAgrawal-A2: @ddimaria 
Done

ddimaria:
Good PR @AyushAgrawal-A2 !  Can you either add all supported functions to the `imports_a_simple_excel_file()` test located at `src/controller/user_actions/import.rs` or create a new test there for functions?  I think I just generated the initial Excel file from the free online version.

AyushAgrawal-A2:
> Good PR @AyushAgrawal-A2 ! Can you either add all supported functions to the `imports_a_simple_excel_file()` test located at `src/controller/user_actions/import.rs` or create a new test there for functions? I think I just generated the initial Excel file from the free online version.

@ddimaria 
Done

quadratic-core/src/controller/operations/import.rs
```diff
@@ -139,19 +139,31 @@ impl GridController {
...
+        // first cell in excel is A1, but first cell in quadratic is A0
+        // so we need to offset rows by 1, so that values are inserted in the original A1 notations cell
+        // this is required so that cell references (A1 notations) in formulas are correct
+        let xlsx_range_to_pos = |(row, col)| Pos {
+            x: col as i64,
+            y: row as i64 + 1,
+        };
+
+        // remove the _xlfn. _xludf. prefix from the function name
+        let re = Regex::new(r"(?:^|[^A-Za-z\d])(?:_xlfn|_xludf)\.").unwrap();
```
HactarCE, line 151: I think this consumes the last character before the `_xlfn.`/`_xludf.`. For example, running the replacement on `1+_xlfn.SUM(A1:10)` results in `1_xlfn.SUM(A1:10)`. Would it be ok to just use `r"(_xlfn|_xludf)\."`?

This also eats any string literal containing these characters, which is admittedly a pathological input but nonetheless may be a reason to properly parse & reserialize formulas when doing the replacement.

quadratic-core/src/controller/operations/import.rs
```diff
@@ -130,7 +131,6 @@ impl GridController {
         file_name: &str,
     ) -> Result<Vec<Operation>> {
         let mut ops = vec![] as Vec<Operation>;
-        let insert_at = Pos::default();
         let error =
```
HactarCE, line 134: This error is always used with an `XlsxError`, so we can factor the `.to_string()` into here and then simplify the call sites:

```rs
let error = |e: XlsxError| anyhow!("Error parsing Excel file {file_name}: {e}");

// ... then at call sites:
.map_err(error)
```
String formatting implicitly calls `.to_string()`.

quadratic-core/src/controller/operations/import.rs
```diff
@@ -191,12 +203,39 @@ impl GridController {
...
+            // formulas
+            let formula = workbook
+                .worksheet_formula(&sheet_name)
+                .map_err(|e: XlsxError| error(e.to_string()))?;
+            let insert_at = formula.start().map_or_else(Pos::default, xlsx_range_to_pos);
+            let mut formula_compute_ops = vec![];
+            for (y, row) in formula.rows().enumerate() {
+                for (x, cell) in row.iter().enumerate() {
+                    if !cell.is_empty() {
+                        let code = re.replace_all(cell, "").into_owned();
```
HactarCE, line 216: Add a comment here explaining what this `.replace_all()` is for, and maybe move the regex closer to the call site. It took me a minute to figure it out because the regex is defined ~65 lines earlier.

You could even put it in a `lazy_static!` _right_ above this line, so that it's only executed once for the whole program but it's still close to where it's used. If that makes the loop body too big, pull out the whole string replacement into a separate function.

quadratic-core/src/controller/operations/import.rs
```diff
@@ -191,12 +203,39 @@ impl GridController {
...
+                        };
+                        let cell_value = CellValue::Code(CodeCellValue {
+                            language: CodeCellLanguage::Formula,
+                            code,
+                        });
+                        sheet.set_cell_value(pos, cell_value);
+                        // add code compute operation, to generate code runs
+                        formula_compute_ops.push(Operation::ComputeCode {
+                            sheet_pos: pos.to_sheet_pos(sheet.id),
+                        });
```
HactarCE, line 229: @ddimaria Do you know if this will work well with formula dependencies? Will it run some code cells multiple times or use outdated results or do any other silly things?

Actually, is this even necessary if we `rerun_all_code_cells_operations()` up in `import_excel()`?

quadratic-core/src/controller/user_actions/import.rs
```diff
@@ -63,6 +69,8 @@ mod tests {
 
     // const EXCEL_FILE: &str = "../quadratic-rust-shared/data/excel/temperature.xlsx";
     const EXCEL_FILE: &str = "../quadratic-rust-shared/data/excel/basic.xlsx";
+    const EXCEL_FUNCTIONS_FILE: &str =
```
HactarCE, line 72: Why not [`include_str!()`](https://doc.rust-lang.org/std/macro.include_str.html) and [`include_bytes!()`](https://doc.rust-lang.org/std/macro.include_bytes.html)?

quadratic-core/src/controller/user_actions/import.rs
```diff
@@ -212,6 +220,50 @@ mod tests {
...
     }
 
+    #[test]
+    fn import_all_excel_functions() {
+        let mut grid_controller = GridController::test_blank();
+        let pos = Pos { x: 0, y: 0 };
+        let mut file = File::open(EXCEL_FUNCTIONS_FILE).unwrap();
+        let metadata = std::fs::metadata(EXCEL_FUNCTIONS_FILE).expect("unable to read metadata");
+        let mut buffer = vec![0; metadata.len() as usize];
+        file.read_exact(&mut buffer).expect("buffer overflow");
```
HactarCE, line 230: If not [`include_str!()`](https://doc.rust-lang.org/std/macro.include_str.html), why not [`std::fs::read()`](https://doc.rust-lang.org/std/fs/fn.read.html)?

HactarCE (changes requested):

quadratic-core/src/controller/operations/import.rs
AyushAgrawal-A2: @HactarCE 
I should have used positive look behind `r"(?<=^|[^A-Za-z\d])(?:_xlfn|_xludf)"` to assert a non word character before this prefix. This will not consume other characters.

In first iteration I did not clean the input, instead after lexar tokenization... I used just remove the prefix `_xlfn. / _xludf.` inside `lookup_function`. In this case prefix will always be at the start of string. This approach did not have the issue of removing characters from strings or other places.

I thought it is bad to save a dirty input so switched to cleaning it at import.

quadratic-core/src/controller/user_actions/import.rs
AyushAgrawal-A2: I copy-pasted file reading code from other functions in the same file.
switched all to `std::fs::read`

quadratic-core/src/controller/operations/import.rs
AyushAgrawal-A2: look-ahead is not supported, switched back to earlier implementation. now storing code string as it is.

HactarCE (approved):
Thanks for the changes!

davidkircos:
Looks great to me! Going to merge this branch. I repurposed this Issue to track our overall Excel File import support https://github.com/quadratichq/quadratic/issues/1154

Issue #1447, opened 2024-06-20
Python improvements - functions, classes, etc.


luke-quadratic:
Spend time making this experience better - at the moment Python in Quadratic is really meant to ignore functions, classes, etc. 

This work is somewhat related to #1115 but not really, just making this stuff work inside a cell's code is the scope of the issue. 

It's also acceptable if the conclusion for this issue is that for non-cloud runners this is just how Python works and that for cloud runners this is solved. 

**Rationale:** it's not uncommon to get help requests related to people who were trying to use functions and ran into all sorts of weird issues. Minimally the app needs some indication when certain language features are not recommended/don't work well. 

Related issue(s): #1114

luke-quadratic:
Largely resolved by #1450 

Please reopen if there are elements of this experience that are poor - #1115 exists for reusable code, otherwise functions etc work much nicer now

PR #1450, opened 2024-06-23
feat: make getCells synchronous

### LINKED ISSUE #1114: error `'await' outside async function` for functions with `cell` in them
I'm sure you're aware of this but I couldn't find an issue referencing it (though #1057 would fix this?)

```
def a():
    return cell(0, 0)

a()
```
fails with error `ERROR: SyntaxError on line 2: 'await' outside async function` assuming because `cell` is async

**Current Fix**
```
async def a():
    return cell(0, 0)

await a()
```
This might be unintuitive for folks not well versed in python and is annoying

**Reproducible Example**
https://app.quadratichq.com/file/19b4314e-d828-4bfb-9b8f-4784c1320acd

### LINKED ISSUE #1409: Dev: improve how c, cells work
Currently we use a bunch of string replaces to make these work without users needing to use `await`. In an ideal world these functions would be inherently blocking and we would neither need to use a string replace as a solution nor have the end user use `await`. 

This is a long-term dev improvement, the current solution in prod works just fine but isn't optimal.

AyushAgrawal-A2:
related to #1447
closes #1409
closes #1114 

This pr is meant to explore usage of [Atomics](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics) and [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer), for communication between workers.
Using this we can block(sleep) a worker thread and execute synchronously, instead of using async / await

Pros:
   - c,  cell, cells, getCells, rel_cell and rc works now synchronously without using async / await. This clear ambiguity of when to add async / await, for enclosing functions (#1114).

Cons:
   - Overhead due to encoding and decoding from shared memory.
   - To use `SharedArrayBuffer`, we need to meet the [Security requirements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements). 
      - For local dev - Chrome treats localhost / 127.0.0.1 as secure context. `COOP`, `COEP` headers are added in `vite.config.js`. So everything works without switching to https.
      - For deployment - It will require all deployment to be secure and for `quadratic-client` we need to add headers mentioned in [Security requirements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements).


davidfig:
This works great. Outside of the necessary server changes, is there anything else that makes this a POC and not ready for review? We'll also need this ported to JS so we can remove the await requirement there as well.

AyushAgrawal-A2:
@davidfig 
ready for review

davidfig:
Awesome. Works wonderfully.

davidfig (approved):

davidfig:
Please resolve conflicts (hopefully last one)

AyushAgrawal-A2:
> Please resolve conflicts (hopefully last one)

@davidfig 
done

PR #1456, opened 2024-06-25
fix: preventDefault and stopPropagation should be called before await

### LINKED ISSUE #1453: bug: keyboard shortcuts don't work on safari
CMD + G doesn't fire Go To menu.

CMD + P doesn't trigger the command palette (it brings up the browser print menu)

AyushAgrawal-A2:
closes #1453 

Changes
   - Moved `event.stopPropagation` and `event.preventDefault` before await fn call.
   - Also, I believe it is good practice to call these first before running any internal code. I have moved all occurrences of these to the top of executing code without changing functional logic.


davidfig:
Please resolve conflict @AyushAgrawal-A2

AyushAgrawal-A2:
> Please resolve conflict @AyushAgrawal-A2

@davidfig 
Done

PR #1479, opened 2024-06-26
fix bug with cmd+clicking the first cell in a boxed selection

### LINKED ISSUE #1478: Cmd select single cell + insert into Python breaks sheet
**To recreate:** 
Select large range of cells 
Cmd + click with thought that you can remove some of the cells from your selection (not possible, is now an issue as well #1476)
Insert cells into Python
Sheet is bricked 

Image

davidfig:
Fixes #1478

To replicate:

1. select multiple cells
2. click cmd+click on the top-left cell

old behavior: crashes b/c of an exception
new behavior: properly selects that cell

Ideal behavior: unselects that cell (but this will require work)


luke-quadratic (approved):
fix works

PR #1480, opened 2024-06-26
fix bug with numeric decimals

### LINKED ISSUE #1477: Can't reduce precision when entire column or row is selected
See gif 

Image

davidfig:
Fixes bug with numeric decimal size when using columns and rows

Fixes #1477


quadratic-core/src/grid/sheet.rs
```diff
@@ -329,7 +319,7 @@ impl Sheet {
     /// get or calculate decimal places for a cell
     pub fn decimal_places(&self, pos: Pos, is_percentage: bool) -> Option<i16> {
         // first check if numeric_decimals already exists for this cell
-        if let Some(decimals) = self.get_column(pos.x)?.numeric_decimals.get(pos.y) {
+        if let Some(decimals) = self.format_cell(pos.x, pos.y, true).numeric_decimals {
```
ddimaria, line 322: This seems less efficient than before.  Do you need to get all of the formats for the column before accessing the numeric decimals?

luke-quadratic:
Works on columns but not rows

quadratic-core/src/grid/sheet.rs
davidfig: You're right. I fixed it. It'll be fine once we move to an all-in format style for cells.

davidkircos (approved):

PR #1493, opened 2024-06-28
Static ip endpoint

ddimaria:
- [x] Create `/static-ips` endpoint
- [x] Add STATIC_IPS env var
- [x] Add test  

## Example

**Request**: `curl --location '127.0.0.1:3003/static-ips'`

**Response**:
```json
{
  "static_ips":["0.0.0.0","127.0.0.1"]
}
```


HactarCE:
I'm not too familiar with web server code, but this looks good to me.

PR #1501, opened 2024-07-02
Get static ips at startup

ddimaria:
Add the following function the the `connectionsClient.ts` file: 

```typescript
const staticIps = await connectionClient.staticIps.list();
```

I added a TODO for @jimniels in `Connections.tsx`: 

```typescript
// TODO(jimniels): implement in the UI
console.log('staticIps', staticIps);
```


quadratic-client/src/shared/api/connectionClient.ts
```diff
@@ -74,4 +82,19 @@ export const connectionClient = {
...
   },
+  staticIps: {
+    list: async (): Promise<StaticIpsResponse | null> => {
+      try {
+        const res = await fetch(`${API_URL}/static-ips`, {
+          method: 'GET',
+          headers: new Headers(await jwtHeader()),
+        });
+        const data = await res.json();
+        return StaticIpsSchema.parse(data);
```
jimniels, line 93: Minor: if we're only ever going to need the `static_ips`, maybe we should just put that part of the code here rather than in each place where we call this? 

For example, rather than having this in multiple places:

```
const staticIps = (await connectionClient.staticIps.list())?.static_ips;
```

Can we just return that here:

```
const data = await res.json();
const { static_ips } = StaticIpsSchema.parse(data);
return static_ips;
```

And then in each place where we use the connectorClient, it's a clean call for the data

```
const staticIps = await connectionClient.staticIps.list();
```

jimniels: That would change our type signature too so it would now be `string[] | null`, since null is what we explicitly define if we catch an error.

jimniels (approved):
One small comment, but otherwise this LGTM.

I'll add a TODO on the main PR to wire this up in the UI.

quadratic-client/src/shared/api/connectionClient.ts
ddimaria: This is done

PR #1509, opened 2024-07-02
Improve mouse interaction with move cells

### LINKED ISSUE #1470: Drag to move cells slightly off
Image

davidfig:
Fixes #1470

- [x] Move cells is now always relative to the cell, not the mouse position
- [x] Fixed bug where a reverse selection (selecting cells from bottom-right to top-left) would not allow move to work


davidkircos (changes requested):
Selecting bottom right to top left, doesn't work properly. You can only grab from an invisible top left position. Not the rest of the borders of the multiselection.

If it's not too difficult adding moving a codecell while editing with the code editor would be awesome.

davidkircos (approved):

PR #1510, opened 2024-07-02
fix: formula clicking to reference a cell

### LINKED ISSUE #1505: bug: formula clicking to reference a cell
CleanShot 2024-07-02 at 10 26 14

AyushAgrawal-A2:
closes #1505 

~~removing this `Some(Token::RBrace | Token::RBracket | Token::RParen) => true,` solves the issue
EmptyExpression inside any bracket / parens should be invalid syntax, same as in excel
I am not able to figure out what was the purpose of this line and if removing it breaks some logic.~~


HactarCE:
In Excel, `AVERAGE(1,)` evaluates to `0.5` because the trailing comma implies an empty expression, and blank evaluates to zero. I'm not a fan of this behavior, and I don't think anyone should rely on it, but it is necessary for compatibility.

AyushAgrawal-A2:
> In Excel, `AVERAGE(1,)` evaluates to `0.5` because the trailing comma implies an empty expression, and blank evaluates to zero. I'm not a fan of this behavior, and I don't think anyone should rely on it, but it is necessary for compatibility.
> 
> Unfortunately, this PR breaks that. :(
> 
> Is there another way to make clicking on a cell to add a reference work the way we want?

@HactarCE 
This is super helpful, any other situation for empty expression in brackets...
We need to handle it on per case basis... rather than allowing empty expressions always inside brackets

AyushAgrawal-A2:
@HactarCE 
I think in excel empty expression are converted to default value only for formula(function) arguments
here it is applied for all cases... i.e. for =(1+) as well
image

HactarCE:
Oh geez ... yeah that's frightening

AyushAgrawal-A2:
@HactarCE 
Sorry, I overlooked that cell referencing check is handled separately in `InlineEditorFormula`
Good to go now

HactarCE:
I think the boolean condition here is reversed. Also, is it possible for us to check only the portion of the formula that's to the left of the cursor? The screenshot in #1505 ends with `)`, but immediately before the cursor is `-` which is why a cell reference is likely to follow.

Excel does this partially by syntactic analysis, but only based on _how you opened the formula editor_. If you click in the inline formula editor after opening it, it completely disables using the arrow keys to select a cell. I'm constantly tripped up by this when using Excel, but also I do like using the left/right arrow keys to jump around with a formula string. I don't know what the best solution here is.

AyushAgrawal-A2:
@HactarCE 
Sorry again :(

> I think the boolean condition here is reversed. Also, is it possible for us to check only the portion of the formula that's to the left of the cursor? The screenshot in #1505 ends with `)`, but immediately before the cursor is `-` which is why a cell reference is likely to follow.

Done



> Excel does this partially by syntactic analysis, but only based on _how you opened the formula editor_. If you click in the inline formula editor after opening it, it completely disables using the arrow keys to select a cell. I'm constantly tripped up by this when using Excel, but also I do like using the left/right arrow keys to jump around with a formula string. I don't know what the best solution here is.

Can I do the arrow key part in another pr...?

HactarCE (approved):

HactarCE:
> Can I do the arrow key part in another pr...?
Certainly! This PR is good to merge now, assuming tests pass etc.

HactarCE:
Ah, I've just realized that we should also allow insert cell reference after `(` or `{`

AyushAgrawal-A2:
> Ah, I've just realized that we should also allow insert cell reference after `(` or `{`. Does that seem right to you @AyushAgrawal-A2?

Yup, I missed that... Thanks

davidkircos (approved):
Functionality looks great!

PR #1514, opened 2024-07-03
Snippets fixes and insert cell references uniformly 

luke-quadratic:
Snippets changes: 
- makes more of these run and changes some formatting, adds semicolons across the board, etc. 

Insert cell ref changes: 
- at present these don't run because it doesn't create a variable, this is fine in Python because Python doesn't need a variable to run; fixed by just creating variable for the reference 
- switched getCells() to cells() for uniformity across languages


davidkircos (commented):
Left comments in Slack.

davidkircos (approved):

davidkircos (approved):

PR #1517, opened 2024-07-03
Delete works properly with column/row selections

### LINKED ISSUE #1444: papercut: selection bugs
- [x] borders applies +1/+1 when selecting more than 1 cell
- [ ] Delete multiple columns also deletes between columns 
CleanShot 2024-06-20 at 15 17 02
- [x] Selection can become invalid (autocomplete top right instead of bottom right) (Command Shift Down, Down, Left, Left, Up)
https://github.com/quadratichq/quadratic/assets/3479421/3a6fd223-5372-4ccb-a2be-481a4907803e

- [x] When moving cells, mouse position and cell position is off. (screenshots don't show cursor, just try dragging cells around)

davidfig:
Fixes #1444 

https://github.com/quadratichq/quadratic/assets/442879/8148648a-9593-432f-9db5-e1badebfb9c8


davidkircos (approved):

davidkircos (approved):

quadratic-core/src/position.rs
```diff
@@ -105,6 +105,7 @@ pub struct Rect {
     pub max: Pos,
 }
 impl Rect {
+    /// Creates a rect from two x, y positions
```
HactarCE, line 108: Either mention that `x0 <= x1 && y0 <= y1` is required, or swap them as needed.

Even better: just call `Rect::new_span()` in this function or vice versa so that there's only one implementation.

HactarCE: Rename to `from_xy_xy()` or similar to distinguish from `from_numbers()` because this is not _the canonical way_ to construct a `Rect`

quadratic-core/src/position.rs
```diff
@@ -136,6 +137,7 @@ impl Rect {
         }
     }
 
+    /// Creates a rectangle from one x, y position and a width and height
     pub fn from_numbers(x: i64, y: i64, w: i64, h: i64) -> Rect {
```
HactarCE, line 141: Rename to `from_xy_wh()` because there's now another way to construct a `Rect` from numbers. The doc comment you added is good.

PR #1494, opened 2024-06-29
Open code editor added to data menu

davidfig:
Screenshot 2024-06-29 at 6 24 50 AM

PR #1520, opened 2024-07-04
feat: add chart interactions on click / dblclick

### LINKED ISSUE #1504: Double click chart to interact
Charts are clunky at the moment and leave much to be desired on interactivity. Double click to interact is a good first start.

AyushAgrawal-A2:
closes #1504

- [x] Hover - Cursor pointer
- [x] Click - Selects corresponding code cell and move chart to top in case of overlapping charts
- [x] Meta / Ctrl + Click - Select code cell + Move to top + Add to Multicursor selection
- [x] DoubleClick - Opens code editor for corresponding code cell

Resizing functionality is same as before.

Built on top of #1494 to reuse `openCodeEditor` function extracted in that pr.


quadratic-client/src/app/gridGL/HTMLGrid/hoverCell/HoverCell.tsx
```diff
@@ -189,6 +189,7 @@ export const HoverCell = () => {
         position: 'absolute',
         visibility: cell ? 'visible' : 'hidden',
         pointerEvents: 'none',
+        zIndex: 999999,
```
davidfig, line 192: Let's try to keep zIndex as small as possible to get the behavior we want. There may be something that needs to go on top of this.

AyushAgrawal-A2: @davidfig 
I needed an arbitrarily large number, when moving a chart on top I am increasing its zIndex to 1 + current max. So zIndex of charts can keep increasing. All charts have 0 zIndex on sheet open.
I think hover popups should always be on top of everything, and zIndex max can go upto int32 max value so there is room for placing something above this.
Still practically even 100 should be enough. It this ok ?

davidfig: Ah. So charts are always z+1. It makes sense why you picked a large number. Maybe 1000 to ensure we don't run into a problem here.

AyushAgrawal-A2: @davidfig 
Done

davidfig (approved):

davidkircos:
@AyushAgrawal-A2 can you move this to a new branch on `quadratic` so we get a preview branch for it?

AyushAgrawal-A2:
move to quadratic repo from fork #1625

PR #1525, opened 2024-07-05
feat: import excel file into open file on xlsx file drop

### LINKED ISSUE #1466: Can only import Excel in dashboard, extend to in-sheet
Add Excel import same way as CSV, Parquet, etc

AyushAgrawal-A2:
closes #1466


davidkircos:
Hey @AyushAgrawal-A2 a couple of questions:

What is the behavior if the excel sheet has multiple tabs? 
What is the behavior if the excel sheet has formulas? Will the formula references break if you don't drop it at 0,0?

AyushAgrawal-A2:
> Hey @AyushAgrawal-A2 a couple of questions:
> 
> What is the behavior if the excel sheet has multiple tabs? What is the behavior if the excel sheet has formulas? Will the formula references break if you don't drop it at 0,0?
> 
> Also, could you move to a PR against `qa` on the `quadratic` fork so it'll launch a preview branch?
> 
> Thanks!

@davidkircos 
Ah, I kinda cheated. It creates new sheets in the existing file. Completely ignores the drop position.
This will be useful only if excel data is required in existing sheet.
As new sheets are created formulas will work same as when importing the file.

Is this behaviour is OK ?

davidkircos:
> > Hey @AyushAgrawal-A2 a couple of questions:
> > What is the behavior if the excel sheet has multiple tabs? What is the behavior if the excel sheet has formulas? Will the formula references break if you don't drop it at 0,0?
> > Also, could you move to a PR against `qa` on the `quadratic` fork so it'll launch a preview branch?
> > Thanks!
> 
> @davidkircos Ah, I kinda cheated. It creates new sheets in the existing file. Completely ignores the drop position. This will be useful only if excel data is required in existing/open file. As new sheets are created formulas will work same as when importing the file for all tabs.
> 
> Is this behaviour is OK ?

What happens if the imported file has a sheet with the same name as an existing sheet in the file? Does this break formulas?

AyushAgrawal-A2:
> What happens if the imported file has a sheet with the same name as an existing sheet in the file? Does this break formulas?
@davidkircos 
Yup, new sheet is added with a `(1)` and sheet references are incorrect, which breaks formula.
We need auto sheet references update operation to resolve this i.e. on renaming sheet, formula reference should auto update.
Currently if a sheet is renamed all its references also break. They should auto update.

AyushAgrawal-A2:
moved to quadratic repo from fork #1610

PR #1527, opened 2024-07-05
fix: allow inline formula editing while editing a code cell

### LINKED ISSUE #1489: Allow inline formula editing while editing a code cell
You can create a new formula cell when a code cell is open. If you try to edit an existing one it will close the code editor. This feels inconsistent.

- [ ] allow editing of inline formula when code cell editor is open
- [ ] Fix bug where codecellcursor disappears when editing an inline formula

AyushAgrawal-A2:
closes #1489 

 - [x] allow editing of inline formula when code cell editor is open
 - [ ] Fix bug where codecellcursor disappears when editing an inline formula


davidkircos:
This is already in `main`

PR #1529, opened 2024-07-06
fix: dispose undo-redo event listener

golok727:


davidfig (approved):

davidfig:
Thanks, this is a good change.

PR #1530, opened 2024-07-08
Remove gridlines overflow

davidfig:
This removes gridlines under cells that overflow to the left or right (mimicking Excel's behavior for overflowing cells).

Screenshot 2024-07-08 at 10 03 51 AM


davidkircos (changes requested):
I noticed two bugs

1. Only rerenders when the viewport moves, not when cell values are updated
2. If a cell would overflow, but doesn't (because there is content in the next cell) the gridline is removed, but shouldn't be

Both are shown in this video

https://github.com/quadratichq/quadratic/assets/3479421/67dd22f1-d4ee-42e7-836a-536054e271b0

davidfig:
Both bugs are fixed.

davidkircos (changes requested):
Cutoff point calculation cuts off text sometimes 


https://github.com/quadratichq/quadratic/assets/3479421/259d485f-750e-49b9-bd00-d71df431eff2

davidfig:
> Cutoff point calculation cuts off text sometimes

It's a super subtle bug. I'll revisit it down the road. It's not new.

davidkircos (approved):

davidkircos (changes requested):
Causes lots of lag when zoomed out. My suggestion is not to calculate these border past a zoom level where they would be hard to see.

Once zoomed far enough out it stops rendering the gridlines at all, but when they are barely visible this causes a lot of lag.

https://github.com/user-attachments/assets/0670d392-7050-4145-bf86-85a21d672828

davidfig:
I changed the percent that we stop drawing gridlines to 25% (it was 10%). See if that helps perf.

davidkircos:
> I changed the percent that we stop drawing gridlines to 25% (it was 10%). See if that helps perf.

At around 39% zoom on the python example doc, it's still pretty laggy for me.

davidfig:
The perf should be much better now. I also reverted the change to when we stop showing the grid lines.

davidkircos (changes requested):
There are rendering bugs after resizing columns

https://github.com/user-attachments/assets/54669678-4974-4d21-9d8b-533ba26ae012

davidkircos (approved):

PR #1531, opened 2024-07-08
fix(client): redo not working with ctrl + shift + z

golok727:
- Convert event.key to lowercase to handle case-insensitive comparisons (excel does the same)
- Ensure functionality works correctly with Caps Lock enabled


davidfig (approved):

davidfig:
Thanks. Good change

PR #1537, opened 2024-07-09
Add ceiling, floor, `INT`, and `MOD` formula functions

HactarCE:
Depends on `next-connections` branch

This includes:
- `CEILING` and `FLOOR`
- `CEILING.MATH` and `FLOOR.MATH`
- `INT` and `MOD`


quadratic-core/src/formulas/functions/mathematics.rs
```diff
@@ -245,12 +296,70 @@ mod tests {
...
+            ("-3.5", "-2", Ok("-4")),
+        ];
+        for (n, increment, expected) in test_cases {
+            let formula = format!("CEIL({n}, {increment})");
+            match expected {
+                Ok(ok) => assert_eq!(ok, eval_to_string(&g, &formula)),
+                Err(err) => assert_eq!(err, eval_to_err(&g, &formula).msg),
+            }
+        }
+    }
```
davidkircos, line 332: Do you test leaving the increment argument undefined?

HactarCE: Both arguments are required, as per Excel.

quadratic-core/src/formulas/functions/mathematics.rs
```diff
@@ -65,6 +65,153 @@ fn get_functions() -> Vec<FormulaFunction> {
                 number.sqrt()
             }
         ),
+        formula_fn!(
+            /// Rounds a number up to the next multiple of `increment`. If
```
ddimaria, line 69: @luke-quadratic many docs to review in this file

ddimaria (approved):
Lots of great additions 🙌

quadratic-core/src/formulas/functions/mathematics.rs
HactarCE: Actually you're right, I should still add a test for this to make sure it gives the right error message.

HactarCE:
Oops, in the commit title I said `INT` and `SUM` but it should be `INT` and `MOD`

PR #1540, opened 2024-07-09
Add `MATCH` function

HactarCE:
Depends on `next-connections` branch


quadratic-core/src/formulas/functions/lookup.rs
```diff
@@ -1026,4 +1081,84 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_match() {
```
ddimaria, line 1086: Great test 🚀

quadratic-core/src/formulas/functions/lookup.rs
```diff
@@ -274,10 +273,66 @@ fn get_functions() -> Vec<FormulaFunction> {
                 Array::new_row_major(result_size, final_output_array)?
             }
         ),
+        formula_fn!(
+            /// Searches for a value in a range and returns the index of the
```
ddimaria, line 277: @luke-quadratic tagging you to review this documentation.

ddimaria (approved):
Great PR!

quadratic-core/src/formulas/functions/lookup.rs
HactarCE: Use `cargo run --bin docgen` to generate markdown

Issue #1536, opened 2024-07-08
Keyboard shortcut tracking/implementation list 


luke-quadratic:
Prioritizing essential keyboard shortcuts from mainstream spreadsheets that aren't yet supported in Quadratic. 

**Please leave a comment** if you want a specific shortcut prioritized.  

- [ ] F2: edit cell 
- [ ] Ctrl+page-up/ctrl+page-down: go to next/prior sheet: 
- [ ] Ctrl + d: fill down from the cell above
- [ ] Ctrl + r: fill right from the cell to the left
- [ ] Shift+space: select row
- [ ] Shift+space: select column 
- [ ] Ctrl+5: strikethrough 
- [ ] Ctrl + `: show formulas in cells instead of results

AyushAgrawal-A2:
@luke-quadratic 
> [ ] Ctrl+page-up/ctrl+page-down: go to next/prior sheet:
Ctrl+page-up/ctrl+page-down  - this combination does not fire an event on chrome (not sure about other browsers). It is reserved for browser tab change.

davidkircos:
I think we should implement shortcuts as part of a JSON "Theme" so users can customize it later. The ability to customize it doesn't need to be in the first iteration here, but building a framework for referencing shortcuts from a JSON definition would ba great foundation before adding a bunch more shortcuts.

luke-quadratic:
That makes sense - expanded the scope of the issue. 

@AyushAgrawal-A2 edited original issue for a JSON theme that users can edit plus changed the button presses for switching sheets (note that there are 3 possible ways in update)

luke-quadratic:
Ctrl + ` covered in another issue #2231 

If we want to prioritize a shortcut JSON it should be a sep issue

PR #1545, opened 2024-07-10
feat: add keyboard shortcuts

AyushAgrawal-A2:
part of #1536 

* [x]  Default JSON for shortcut default theme
* [x]  F2: edit cell
* [x]  Support both Ctrl + Shift + PageUp / Ctrl + Shift + PageDown **AND** option + up-arrow, option + down-arrow **AND** option + right arrow/option + left arrow: go to next/prior sheet (this covers as close as possible to all bases for both mainstream spreadsheets)
* [x]  Ctrl + d: fill down from the cell above
* [x]  Ctrl + r: fill right from the cell to the left
* [x]  Shift + space: select row
* [x]  Ctrl + space: select column


davidfig (approved):

davidfig:
Code looks good

quadratic-client/src/app/theme/shortcuts.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * Default shortcuts for the application.
+ * Mac key names are used for the combinations.
+ */
+export const defaultShortcuts = {
+  // QuadraticGrid.tsx
+  grid_pan_mode: ['Space'],
+
+  // keyboardViewport.ts
+  show_command_palette: ['Cmd + P', 'Cmd + P', 'Cmd + /'],
```
davidkircos, line 10: duplicate

quadratic-client/src/app/theme/shortcuts.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * Default shortcuts for the application.
+ * Mac key names are used for the combinations.
+ */
+export const defaultShortcuts = {
```
davidkircos, line 5: I think that our shortcut Schema should be typed with Zod, and should have platform specific options:

We may also want to add modifiers of when they are active: Check out VSCode keybindings schema 
https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-rules

davidkircos: They also have a way of making shortcuts keyboard layout independent 
https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-layoutindependent-bindings

davidfig: I like this approach.

AyushAgrawal-A2:
moved to quadratic repo from fork #1612

quadratic-client/src/app/theme/shortcuts.ts
AyushAgrawal-A2: done

PR #1548, opened 2024-07-13
Fix offline and multiplayer

davidfig:
- [ ] Add correct documentation links to `urls.ts`

### Multiplayer Fixes
- [x] refresh jwt token before attempting to reconnect to mp
- [x] put back the`Reconnecting...` after a broken connection (this broke in the web worker PR a long time ago)
- [x] add UI when disconnecting and reconnecting (only after a disconnect w/message) 

image

Screenshot 2024-07-13 at 11 04 26 AM

### Offline Fixes
- [x] add toast when offline transactions are applied
- [x] make offline transactions undoable
- [x] fix bug related to same offline transaction saved locally multiple times

image


davidkircos:
I tested this and it appears to work. 

One nit, can we close the popup once the connection is reestablished?
CleanShot 2024-07-18 at 14 31 25@2x

quadratic-client/src/shared/components/GlobalSnackbarProvider.tsx
```diff
@@ -1,7 +1,8 @@
-import CloseIcon from '@mui/icons-material/Close';
-import { Alert, AlertColor, IconButton, Snackbar } from '@mui/material';
-import * as React from 'react';
+// import CloseIcon from '@mui/icons-material/Close';
```
ddimaria, line 1: Remove?

quadratic-client/src/shared/components/GlobalSnackbarProvider.tsx
```diff
@@ -21,44 +27,46 @@ const defaultContext: GlobalSnackbar = {
...
       setMessageQueue((prev) => prev.slice(1));
       setStayOpen(!!messageQueue[0].stayOpen);
       setOpen(true);
-    } else if (messageQueue.length && activeMessage && open) {
-      // Close an active snack when a new one is added
-      setOpen(false);
     }
+
+    // we don't want a new message to replace the current message until the timer expires
+    // else if (messageQueue.length && activeMessage && open) {
```
ddimaria, line 67: Remove?

quadratic-client/src/app/web-workers/multiplayerWebWorker/worker/multiplayerServer.ts
```diff
@@ -110,30 +108,34 @@ export class MultiplayerServer {
...
+  // Attempt to connect to the server.
+  // @param skipFetchJwt If true, don't fetch a new JWT from the client
+  // (only true on first connection, since client already provided the jwt)
+  private async connect(skipFetchJwt = false) {
     if (this.state === 'connecting' || this.state === 'waiting to reconnect') {
       return;
     }
 
     this.state = 'connecting';
+    if (!skipFetchJwt) {
```
ddimaria, line 120: I'm glad we're refreshing the JWT, it's probably the cause of several of our issues.  Parsing JWTs are relatively cheap.  Should we consider checking for expiration (either with some event or on a timer) and refreshing then rather than here?

ddimaria (approved):
Code looks good.  I have a couple of nits and one general question.  This will be a very beneficial feature 💯

quadratic-client/src/app/web-workers/multiplayerWebWorker/worker/multiplayerServer.ts
davidfig: Isn't the JWT used only for initial authentication? So we only check it when the user first connects to mp. So as long as they remain connected, we don't need to renew the JWT. Or are we continuing to use the JWT?

PR #1549, opened 2024-07-14
Move cells no longer moves code results

### LINKED ISSUE #1546: Confusing Behavior When Dragging Code Output Without Source Cell
See video

Maybe some sort of warning when you do this? Or prevent it by default? Not sure.

https://github.com/quadratichq/quadratic/assets/3479421/437b5ccc-29cc-4a8f-a0cd-902be761b175

davidfig:
Fixes #1546.

We added copying code results to the Quadratic clipboard when the code cell itself is not included in the selection. This results in weird behavior when moving cells (since the move cell operations reuse the copy/paste code). This PR removes that behavior by adding an `is_move` flag for the cut operation.

New behavior:
https://github.com/user-attachments/assets/f35f6719-b25e-42f4-9016-fcdeecdacf26

Old behavior: 
https://github.com/quadratichq/quadratic/issues/1546


ddimaria:
Code looks good!  I'll wait for @davidkircos to approve/merge once he reviews this in the UI.

davidkircos:
Creates a odd behavior moving formatting but not values. Maybe we want to disable move unless you are moving the whole array output?

https://github.com/user-attachments/assets/b25b020c-795c-44dc-8cba-c1616f9ec8d5

davidfig:
Maybe we should keep the old behavior for this. It'll still have the weird format moving, but not much we can do about that without a lot of work.

davidkircos:
We decided we like the current behavior better.

PR #1553, opened 2024-07-15
Improvements to ensureVisible when moving/selecting cells

### LINKED ISSUE #1532: bug: multi selection calls ensureVisible on users cursor
If you are making a selection larger than the viewport and you use the mouse to change the selection viewport jumps back to the users cursor. See video: 

https://github.com/quadratichq/quadratic/assets/3479421/96e63f8e-d411-49cb-a064-7fe6155c3edf

On a related note. When holding down the selection (mouseDown), moving the viewport should also update the selection not just when moving the mouse.

davidfig:
- [x] Shift+click will move viewport to the edge of the selection instead of the cursor
- [x] Shift+arrow will move viewport to the edge of the selection instead of cursor

Fixes #1532


quadratic-client/src/app/grid/sheet/SheetCursor.ts
```diff
@@ -128,7 +128,7 @@ export class SheetCursor {
       columnRow?: ColumnRowCursor | null;
       cursorPosition?: Coordinate;
       keyboardMovePosition?: Coordinate;
-      ensureVisible?: boolean;
+      ensureVisible?: boolean | Coordinate;
```
ddimaria, line 131: I don't know enough context here, but this seems like it should be a separate variable (or maybe a variable rename)?

ddimaria:
Code looks good (I have one comment).  I'll let @davidkircos approve once he validates this in the UI.

davidkircos (changes requested):
I am not sure what all the changes in their PR do. Can you give a video example?

This does seem to fix the initial issue.

quadratic-client/src/app/grid/sheet/SheetCursor.ts
davidfig: This provides another option for ensureVisible. The default is to either do nothing (false) or move to the current cursor position (true). The Coordinate is when you want to move to a different cursor position. For example, when selecting a range, you want to move the viewport to the corner of the range, not the cursor position (which is where the range starts).

davidfig:
This changes when the viewport moves mostly during keyboard selection. Originally, it used the starting cursorPosition. This changes it to use the ending cursor position for selections, like shift + arrow or shift + cmd + arrow.

Old behavior: 

https://github.com/user-attachments/assets/5b3fa9a1-dcc8-4767-aef6-e3b2475e4458

New Behavior:

https://github.com/user-attachments/assets/5a0f517d-4377-48e1-95fe-849da5548a2b

davidkircos (approved):

Issue #1221, opened 2024-03-26
Add tabs to code editor so you can have more than one code cell open


davidfig:

luke-quadratic:
Brainstorming on expected behavior.. 

Expected behavior: 

- when editor is open and you click into another cell editor stays open and just opens additional tab, original tab you were on stays open 
- tab names are the cell the code belongs to, e.g. "Cell (9,2) Python"
- when you close editor and reopen via a cell the tabs will have closed and only tab open is the cell you just clicked into (not sure on this one, are tabs maintained throughout the session unless you explicitly close a tab?)

PR #1558, opened 2024-07-17
feat: new hierarchy for code editor

jimniels:
Creates a new hierarchical configuration for the code editor. Makes it look/feel a lot more like VSCode (also opens the door to the option of having multiple tabs "open" in the code editor

image

**Alternative**: don't make the tab like a cell, instead just make it the whole top bar

image

**Todos**

- [ ] Resizing controls are off because of the new top header. Will need fixing.


davidkircos:
Closed and linked to https://github.com/quadratichq/quadratic/issues/1221 for later.

PR #1567, opened 2024-07-18
feat: tweaked prompt for ai returning sql queries

jimniels:
Today, the AI returns a sql query that doesn't have the right quoting. This means the query might not run.

For example, it gave me this:

```sql
SELECT 
    Team.name AS team_name, 
    COUNT(UserTeamRole.user_id) AS user_count 
FROM 
    public.Team
JOIN 
    public.UserTeamRole 
ON 
    Team.id = UserTeamRole.team_id
GROUP BY 
    Team.name
ORDER BY 
    user_count DESC;
```

Which I had to tweak to this, to get it to run:

```sql
SELECT 
    "Team".name AS team_name, 
    COUNT("UserTeamRole".user_id) AS user_count 
FROM 
    public."Team"
JOIN 
    public."UserTeamRole" 
ON 
    "Team".id = "UserTeamRole".team_id
GROUP BY 
    "Team".name
ORDER BY 
    user_count DESC;
```

The idea is to tell the AI to return queries like that by default following the same logic we used in our code base for inserting a query: Postgres queries do `"schema"."Table"` and MySQL queries do `` `schema`.`Table` ``


quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
```diff
@@ -58,7 +58,15 @@ Do not use any markdown syntax besides triple backticks for ${getConnectionKind(
...
 The cell type is ${getConnectionKind(mode)}.
 The cell is located at ${selectedCell.x}, ${selectedCell.y}.
-${schemaJsonForAi ? `The schema for the database is:\`\`\`json\n${schemaJsonForAi}\n\`\`\`` : ``}
+${
+  schemaJsonForAi
+    ? [
+        `The schema for the database is:\`\`\`json\n${schemaJsonForAi}\n\`\`\``,
+        'When generating postgres queries, put schema and table names in quotes, e.g. "schema"."TableName".',
+        'When generating mysql queries, put schema and table names in backticks, e.g. `schema`.`TableName`.',
+      ].join('\n')
```
jimniels, line 67: Idk if this is the best way to write this prompt. It'll do it to everything, and maybe you don't always want that? I'm no SQL pro.

Example:

CleanShot 2024-07-18 at 14 14 54@2x

davidkircos (changes requested):
yeah I don't think we want it to do this for all field names

davidkircos:
I am running into this issue all the time, resurrecting this change

AyushAgrawal-A2:
added in https://github.com/quadratichq/quadratic/pull/1864

PR #1572, opened 2024-07-19
fix connections deploy on code change

davidkircos:
- [ ] Update IAM permissions
- [ ] Clean up old instances?


infra/connection/connection.ts
```diff
@@ -32,12 +32,20 @@ const domain = config.require("domain");
...
+  vpcId: connectionVPC.id,
+});
+
+// Step 1: Create or update the Launch Template
+const launchTemplate = new aws.ec2.LaunchTemplate("connection-lt", {
   instanceType: instanceSize,
   iamInstanceProfile: instanceProfileIAMContainerRegistry,
   imageId: latestAmazonLinuxAmi.id,
-  securityGroups: [connectionEc2SecurityGroup.id],
+  securityGroupNames: [connectionEc2SecurityGroup.name],
```
ddimaria, line 48: Odd that this uses names vs ids

ddimaria (approved):
I'm not 100% sure what this is doing, but I don't see any coding issues.

infra/connection/connection.ts
davidkircos: Good point I think I should use vpcSecurityGroupIds
CleanShot 2024-07-19 at 16 22 58@2x

ddimaria (approved):

PR #1587, opened 2024-07-23
Fixes a bunch of problems with hmr

davidfig:
- [x] remove index.tsx for both TopBar.tsx and BottomBar.tsx (react hmr really didn't like those and they didn't do much)
- [x] remove user thrown exceptions in a few files where the state becomes wonky during hmr

Future PR
- [ ] snippets popover results in circular hmr (package.json#start -> ``vite dev --debug hmr` to see the circular dependencies


quadratic-client/src/routes/useFileRouteLoaderData.tsx
```diff
@@ -0,0 +1,7 @@
+import { ROUTE_LOADER_IDS } from '@/shared/constants/routes';
+import { ApiTypes } from 'quadratic-shared/typesAndSchemas';
+import { useRouteLoaderData } from 'react-router-dom';
+
+type FileData = ApiTypes['/v0/files/:uuid.GET.response'];
+
+export const useFileRouteLoaderData = () => useRouteLoaderData(ROUTE_LOADER_IDS.FILE) as FileData;
```
jimniels, line 7: Why move this?

`useFileRouteLoaderData()` is basically just a convenience so you don't have to do `useRouteLoaderData("route-id")` in every single component where you need it.

I figured why not just co-locate it with the route it supplies data for (the file route). That's why it was in `routes/file.$uuid` to begin with.

There are a couple other routes where we do this same thing, so this would be breaking with that convention.

If you feel strongly about moving it out to its own file, we should move it into like a hooks folder and not in `src/routes` because the routes folder is reserved for react-router route files and this is not a route file.

davidfig: It was causing a circular dependency in React. Moving it out of this file fixed it.

davidkircos (approved):

PR #1590, opened 2024-07-23
Fix clear formatting removes borders if selection contains rects

### LINKED ISSUE #1583: bug: clear formatting (for borders) doesn't work
Doesn't seem to work at all (via floating context menu, short cut `CMD + \`, or command palette)

https://github.com/user-attachments/assets/3197324e-81ec-45a3-ad37-6fc57f152028

davidfig:
Fixes #1583 

Note: the clear borders implementation is limited to selections containing rects (similar to the limitation for setting border). We do not have a good way of setting borders for selections containing all, columns, or rows. (We also don't have bounds for borders, so we can't even fake it for clearing.)

### DIFF
--- a/quadratic-core/src/controller/execution/execute_operation/execute_borders.rs
+++ b/quadratic-core/src/controller/execution/execute_operation/execute_borders.rs
@@ -30,7 +30,7 @@ impl GridController {
                     },
                 );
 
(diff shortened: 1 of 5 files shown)


jimniels (approved):
I can't vouch for the code, as it's all rust, but functionality-wise, it works. I can now clear borders as expected.

PR #1594, opened 2024-07-23
Change when formula reference insertion happens

### LINKED ISSUE #1593: Secondary Formula arguments can't accept a range
CleanShot 2024-07-23 at 09 34 50

HactarCE:
Fixes #1593

This PR changes the behavior that determines whether arrow keys or clicking while typing a formula inserts a reference.

The intended behavior of this PR is that clicking outside a cell or using the up/down arrow keys inserts a reference if a reference is expected at the current cursor position in the formula, based only on the nearest non-whitespace token to the left of the cursor. Additionally, clicking in the formula text box confirms the reference insertion, so that the user can then navigate within the textbox using arrow keys.

There is another way that a user can confirm a reference insertion using just the keyboard, which is spoilered in case you'd like to try to figure it out yourself (to gauge whether it's too non-obvious).

<details>
  <summary>Spoiler warning</summary>
  
  Type any character, then delete it. Now you can use left/right arrow keys to navigate within the formula text.
  
</details>

Selecting a cell reference with either the mouse or the keyboard allows the user to click or use the up/down arrow keys to replace it with a new cell reference.

This isn't exactly equivalent to Excel's behavior, but it's quite close and in my opinion this behavior is easier to understand and more useful. (In Excel, it's possible to get into a state where it becomes impossible to insert cell references using the mouse and/or keyboard.)


HactarCE:
The cell selection visuals seem to be broken -- is that new here?
image

Also, typing `=SUM` and then clicking causes inserting a cell reference, so something's not right with this change ...

HactarCE:
This does not handle arrow keys + shift or cmd/ctrl the way I expect. IMO they should behave the same as arrow keys with no modifiers.

davidkircos (approved):
Functionality looks good to me. 

Waiting for review from Luke to merge.

Issue #1556, opened 2024-07-16
Bug: dragging large selection of data not syncing 


luke-quadratic:
See gif, to reproduce put any data in a sheet and try to drag to copy a few hundred or few thousand rows; you'll find you now can't click anywhere, and if you refresh the data that seems you've copied is no longer there - no indication in bottom that you aren't synced, says connected and no files currently syncing 

![CleanShot 2024-07-16 at 12 02 23](https://github.com/user-attachments/assets/762110f5-7c05-4bfe-ab4b-8cbc9d9fa6b6)

PR #1526, opened 2024-07-05
Combination of Wrap Text and Fit numbers PRs

davidfig:
This is a combination of @AyushAgrawal-A2 's Wrap Text PR and the @davidfig 's Fit Numbers PR

# feat: wrap text, align text vertically and auto resize row

closes #1227 
closes #1460 
closes #1463

Changes:
- [x] Cell Label - Wrap, Clip & Overflow
- [x] Cell Label - Vertical Align
- [x] Inline Editor - Same text rendering as cell format (horizontal align, vertical align and wrapping)
- [x] Re-wrap text on cell width change 
- [x] Auto row resize for wrapped text - on cell width change or content change
- [x] Double click on row heading to auto adjust row height
- [x] Fixed bugs related to neighbouring hash position and clipping
- [x] Update clipboard for vertical align
- [x] Multiplayer sync after transient / auto resize
- [x] Store manual / auto resize state. Disable auto resize for manually resized rows.
- [x] Auto resize column (on column heading dbl click) should resize to unwrapped text width.
- [x] Tests 

# Fit numbers

Change how numbers render. In Excel and Sheets, the numbers never overflow the cell. They either reduce their precision, change to scientific, or show up as ####. Fixes #1468.

- [x] redo how CellValue::Number.to_display() works (removing formatting)
- [x] keep some formatting within Rust (but outside the to_display function) so we can still search against what will likely be the displayed number
- [x] move all number formatting from Rust to TS :( - this is necessary b/c we don't know how to stringify the number until we are actually rendering and can compare glyph sizes to the cell width
- [x] move tests from Rust to TS for number rendering
- [x] show #### when number overflows content
- [x] figure out when to change the display from number to scientific to show larger numbers in a smaller space
- [x] fix search to use actual display value (ie, put back the rust to_display w/options for search purposes)

PR #1595, opened 2024-07-23
perf: improve autocomplete performance

AyushAgrawal-A2:
for #1556 
This PR is meant for trial and discussion, may not merge

Problems causing #1556 
All workers are busy due to large number of operations. After waiting for some time app becomes responsive again. To see this visually, create a cell with some formatting and sheet fill. Drag it, via autocomplete, to a large selection. Water falling format updates can be seen progressing.

This is happening because each formatting update for every cell is applied in a separate operation. This leads to:
- Large number of inter worker message which take a lot of time.
- Sending individual cell update for each operation, cause unnecessary render updates.
- Use of `transaction.reverse_operations.insert(0, ...` which is O(n^2) for n operations.

In this PR, I am trying to batch client updates. Sending update for the entire affected `sheet_rect` is faster than filter and sending only changes for each individual cell.

Current performance in this PR is a lot faster than before but still slower than google sheets / excel web. More optimisations can be done.
Also this can improve performance of copy paste as well. Not implemented in this PR.


davidfig:
Awesome. This was definitely on my list as well. I thought of two approaches for this problem: (1) your approach to batch the client calls to avoid rerendering hashes multiple times; and (2) rewriting the Operation::SetCellValues to be more efficient and accept something other than CellValues. The problem with CellValues is that it has to be a rectangle, and autocomplete is always a rectangle with a corner removed.

What are your thoughts on continued improvement?

AyushAgrawal-A2:
@davidfig 
I am already trying to figure out -
- Refactoring to Operation::SetCellFormatsSelection 
- How to do batch client update for multiplayer, currently only operations are sent to multiplayer.

Current bottleneck is in `finalize_transaction` step, after sending batch client update. Size of reverse_operations vector is huge. Processing it and sending all those transactions is taking >90% of time currently.

hash boundaries bug is fixed in #1526

davidfig:
Cool. Great on the hash boundaries bug. 

I'm not sure if we need to batch the client updates for multiplayer. We've been handling multiplayer via transactions, and I'm reluctant to change that.

Ideally, we should reduce the number of the operations.

AyushAgrawal-A2:
@davidfig 
I am also reluctant to change multiplayer, a lot of code I don't understand yet.
But I don't know how else to solve water falling client update and large number of worker messages for multiplayer.

Reducing operations is priority, I am trying to do that too but even if there are a ton of operations multiplayer should be able to batch update same as host client is doing.

davidfig:
If the batching works at the level of the transaction, then mp should work fine. It runs the ops via shared transactions.

AyushAgrawal-A2:
@davidfig 
alot faster now :man_dancing:

davidkircos:
Yeah, wow, this is way faster! Is this PR ready for review?

AyushAgrawal-A2:
@davidkircos 
I would like to add some test if we are to merge this PR

Also, I am not able to figure out how to batch client update for multiplayer. Water falling is still present for MP. Can be done in another PR.

Otherwise, ready for user testing

AyushAgrawal-A2:
@davidfig 
switch to `SetCellFormatsSelection`
faster than excel web now, I am satisfied with the performance

davidfig:
Awesome. Can you also increase test coverage? Thanks!

AyushAgrawal-A2:
> Awesome. Can you also increase test coverage? Thanks!

Yup, on it

AyushAgrawal-A2:
I wanted to add `expect_js_call_count` in test for `batch_client_update` but this required adding `#[parallel]` macro to all tests.
I have already done that in #1526. Can be done after merging it.

davidfig:
Can you add the batch_client_update_rect to receive_multiplayer#received_transction? Wouldn't that add this functionality to mp, or is there more that needs to happen?

quadratic-core/src/controller/execution/execute_operation/mod.rs
```diff
@@ -48,7 +48,10 @@ impl GridController {
                 }
             }
 
-            if cfg!(target_family = "wasm") && !transaction.is_server() {
+            if (cfg!(target_family = "wasm") || cfg!(test))
+                && !transaction.is_server()
+                && transaction.batch_client_update_rect.is_none()
```
davidfig, line 53: Why are you checking the batch here?

davidfig (changes requested):

quadratic-core/src/controller/execution/execute_operation/mod.rs
AyushAgrawal-A2: to solve too many worker messages for transactions having large number of operations
worker message are very slow

davidfig:
Cool. If you can make it dynamic, then mp should just work. We can be naive and use a Bounds struct that allows you to increase its size based on what the ops are doing.

AyushAgrawal-A2:
I hate this bot :angry: 
image

image

AyushAgrawal-A2:
> Cool. If you can make it dynamic, then mp should just work. We can be naive and use a Bounds struct that allows you to increase its size based on what the ops are doing.

yup, I figured this much. we still need a way to tell mp than it needs to batch operations and not send individual updated.
maybe based on operations vector size, I am sure now
current bottleneck is `send_transactions` which should be solve by #1607

davidfig:
> we still need a way to tell mp than it needs to batch operations and not send individual updated

If it's dynamically updated in a more complicated Batch struct, then we'll know exactly what needs to be sent to the client based on the execute operation (which should update the Batch struct as it executes).

AyushAgrawal-A2:
> > we still need a way to tell mp than it needs to batch operations and not send individual updated
> 
> If it's dynamically updated in a more complicated Batch struct, then we'll know exactly what needs to be sent to the client based on the execute operation (which should update the Batch struct as it executes).

I disagree actually, having fine grained updates was the biggest issue that this PR solve. This will again lead to multiple re-renders and a ton of worker messages depending on the complexity of Batch struct.

My idea behind batch update is to not figure out finer updates, but instead to send a larger rect re-render to avoid 
- multiple rerenders of  same hash 
- number of worker messages

Worker messages is the slowest step here, which I am trying to optimise

davidfig:
Hmm... I'm not sure I follow. My suggestion is to create a Batch struct that tracks all changes made by operations. Then in finalize_transaction, we use the Batch struct to send the necessary messages.

To avoid too many messages, you can combine the rects before sending the messages. For example, you collect a set of cell hashes that have changed based on the rects, then you only send those cell hashes (so each cell hash is only sent once).

You end up with the correct number of messages this way without having to define the rect before the transaction is executed. Does that make sense?

AyushAgrawal-A2:
@davidfig 
Batch update part is not doing much now after moving to 'SetCellFormatsSelection'. Earlier each cell had 10 operations for format update. Now we have just one operation for all format updates for the entire rect.
I should have done this before playing with other code.

As we have just ~2 operations now re-renders and worker messages are almost same as what we would achieve with a Batch struct.

I have disable it and push it here. Please try it on Vercel preview. It is pretty fast. I tried autocomplete for a very large rect `1x4` to `10000 x 8`. Add lot of different formatting in source rect to stress the operation. It looks OK to me.

I think we can remove & postpone Batch update logic.
I get what you are saying but I am not able to think of a scenario which I can implement and compare practically.

davidfig:
Sounds good. We'll likely revisit the batching down the road.

AyushAgrawal-A2:
> Sounds good. We'll likely revisit the batching down the road.

Great... Thanks for working this out with me. This PR was meant to do exactly this :smiley: 

Also removing batch_update logic might satisfy the insatiable codecov bot :stuck_out_tongue_closed_eyes:

PR #1599, opened 2024-07-23
fix: execute keyboard events only when suggestion widget is hidden

### LINKED ISSUE #1596: Formula pressing enter to complete hint runs cell, doesn't use hint
CleanShot 2024-07-23 at 11 58 32

AyushAgrawal-A2:
closes #1596


quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
```diff
@@ -300,8 +301,22 @@ class InlineEditorMonaco {
       language: inlineEditorHandler.formula ? 'formula' : undefined,
     });
 
+    const suggestionWidget = (this.editor.getContribution('editor.contrib.suggestController') as any | undefined)
```
ddimaria, line 304: @AyushAgrawal-A2 can you remove this `any` and replace with a type?

AyushAgrawal-A2: @ddimaria 
This is a hacky solution.
This code using Monaco's internal api to access suggestionWidget state. Ref(https://github.com/Microsoft/monaco-editor/issues/991#issuecomment-411356366)
I am not able to reach the correct TS type for this.

AyushAgrawal-A2: @ddimaria 
I checked, no way to get correct type from `monaco-editor`, `any` is required here

ddimaria (approved):

davidkircos (approved):

PR #1600, opened 2024-07-23
fix: focus grid after closing GoTo menu and CellType menu

### LINKED ISSUE #1592: Opening and closing command palette doesn't return focus to the sheet
To recreate: 
- open command palette with / 
- close command palette
- try to do a formula with = or open command palette again with /, won't be possible

AyushAgrawal-A2:
closes #1592


davidkircos:
Do you know where the underlying issue is? Why does adding a delay work?

AyushAgrawal-A2:
> Do you know where the underlying issue is? Why does adding a delay work?

Actually only `setTimeout(() => focusGrid());` also works for me, delay value is not required
Reason for
- setTimeout -> call `focusGrid` after state update and react re-render
- 100ms delay -> I saw this being used in other places for `focusGrid` and I assumed it may be required for other browsers. I am just testing on chrome. Firefox and safari can be finicky.

quadratic-client/src/app/ui/menus/CellTypeMenu/CellTypeMenu.tsx
```diff
@@ -98,7 +98,7 @@ export default function CellTypeMenu() {
       ...editorInteractionState,
       showCellTypeMenu: false,
     });
-    focusGrid();
+    setTimeout(() => focusGrid(), 100);
```
ddimaria, line 101: Is there an event to listen to in order to avoid possible timing issues with the current approach @AyushAgrawal-A2?

AyushAgrawal-A2: The correct approach to do this in React should be using useEffect + useRef 
I think it can be achieved using recoil atom effects as well. I'll give it a try.

AyushAgrawal-A2:
@ddimaria 
I have switched to recoil atom effects

PR #1601, opened 2024-07-23
feat: add basic markdown rendering in AI chat

### LINKED ISSUE #736: feat (enhancement): format AI answers as markdown (or tell it not to reply that way)
Currently, you can get responses that come back structured as markdown so you get asterisks that aren't being bolded

CleanShot 2023-10-23 at 14 08 13@2x

update 2/2024: issue still exists in assistant

AyushAgrawal-A2:
closes #736


jimniels:
I'm not sure if we want to solve this by adding a dependency for a markdown parser.

[Part of the prompt we send to OpenAI specifically says "do not give us back markdown"](https://github.com/quadratichq/quadratic/blob/a2fdc090fd0e3a36986b5b9d57799f71a0e92d7b/quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx#L57) but maybe it's just sending us markdown (sometimes) anyway? 

I think maybe we should reinforce via the prompt to _not_ send markdown back. Then we don't have to bring in a dependency for a markdown parser *and* we don't have to worry about additional styling considerations for the markdown itself (e.g. h1-5 headings, ordered/unordered lists, bold, italic, etc.)

What do you think @davidkircos ?

davidkircos:
@jimniels, what's the concern about adding a dependency? 

It seem like showing markdown could be a better option, although I need to try it out a bit and see how it looks. I am working on fixing the preview branch.

ddimaria (commented):
Code looks good.  @davidfig Are you OK with these additional packages added to the client?

* react-markdown
* remark-gfm

davidkircos (approved):

davidkircos:
I think we should add this, markdown packages are small.

PR #1606, opened 2024-07-24
Fix Heroku Build

davidkircos:
After deploying manually run
`ALTER TABLE "Team" DROP COLUMN "activated";`


ddimaria (approved):

ddimaria (approved):

Procfile
```diff
@@ -1,3 +1,3 @@
-web: cd quadratic-api && node -r newrelic dist/src/server.js
+web: cd quadratic-api && node dist/src/server.js
```
davidkircos, line 1: unrelated but we don't use newrelic so might as well remove

ddimaria (approved):

PR #1607, opened 2024-07-24
Serialize and compress transactions

ddimaria:
- [x] JSON serialize and Zlib compress operations from core-to-client, client-to-multiplayer, multiplayer-to-client, and client-to-core
- [x] Handle binary operations in multiplayer and files
- [x] Update tests


quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -204,7 +204,7 @@ impl GridController {
                 .transactions
                 .out_of_order_transactions
                 .iter()
-                .position(|t| t.sequence_num.unwrap() < sequence_num)
+                .position(|t| t.sequence_num.unwrap_or(0) < sequence_num)
```
HactarCE, line 207: Can you add a doc comment to `sequence_num` explaining why it might be `None`? Glancing at this code, I don't know if `.unwrap_or(0)` is correct.

quadratic-core/src/controller/transaction.rs
```diff
@@ -1,11 +1,27 @@
...
+
 use super::{
     active_transactions::pending_transaction::PendingTransaction, execution::TransactionType,
     operations::operation::Operation, GridController,
 };
 
+pub static SERIALIZATION_FORMAT: SerializationFormat = SerializationFormat::Json;
+pub static COMPRESSION_FORMAT: CompressionFormat = CompressionFormat::Zlib;
+pub static HEADER_SERIALIZATION_FORMAT: SerializationFormat = SerializationFormat::Bincode;
+pub static CURRENT_VERSION: &str = "1.0";
```
HactarCE, line 18: Will this need to be incremented any time we add a new kind of operation or new formatting option?

quadratic-files/src/truncate.rs
```diff
@@ -13,11 +13,13 @@ pub(crate) fn processed_transaction_key(file_id: &str, sequence_num: &str) -> St
...
 
-pub(crate) fn parse_processed_transaction_key(key: &str) -> Result<(String, String)> {
+pub(crate) fn parse_processed_transaction_key(key: &[u8]) -> Result<(String, String)> {
+    let key = from_utf8(key)
+        .map_err(|_| FilesError::Unknown("Could not parse key as UTF-8 string".into()))?;
     let split = key.split('.').collect::<Vec<&str>>();
 
     if split.len() < 2 {
-        return Err(FilesError::Unknown(format!("Could not parse key {key}")));
+        return Err(FilesError::Unknown("Could not parse key {key}".into()));
```
HactarCE, line 22: Needs `format!(...).into()`

quadratic-rust-shared/src/pubsub/redis_streams.rs
```diff
@@ -33,7 +36,7 @@ impl Debug for RedisConnection {
     }
 }
 
-type Message = (String, String);
+type Message = (String, Vec<u8>);
```
HactarCE, line 39: Can you add a doc comment that says what each of these is? And/or make it a `struct` with named fields.

HactarCE (commented):
Sorry, I forgot to do the official "start a review" thing so my first few comments landed outside.

Besides a few small things, the Rust side looks good to me!

quadratic-files/src/truncate.rs
ddimaria: Nice catch ❤️ , this is done.

quadratic-rust-shared/src/pubsub/redis_streams.rs
ddimaria: I added the following: 

> // A message consists of a key (String) and a value (Bytes).

quadratic-core/src/controller/transaction.rs
ddimaria: This is untested, but I imagine the version will bump anytime there is a structural (minor) or breaking change (major, i.e. when it's not backwards compatible).  Changing the compression or serialization format would cause a increase in the major version and the necessity to add in some pattern matching for handling the different versions (TBD when the version bumps).  Additions to the structure will likely bump the minor version (tests should indicate this need) and not require any additional work.

quadratic-core/src/controller/execution/receive_multiplayer.rs
ddimaria: After reviewing more of the quadratic-core code, I believe you're correct Andrew.  I updated this value to be:

```rust
self.transactions.out_of_order_transactions.len()
```

ddimaria: @davidfig can you confirm the above?

quadratic-client/vite.config.js.timestamp-1722017521018-562d6b4f12d52.mjs
davidfig, line 1: What's this?

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -223,14 +226,24 @@ impl GridController {
...
+                    cursor: None,
+                    ..Default::default()
+                };
+
+                self.client_apply_transaction(&mut transaction, t.sequence_num);
+            } else {
+                dbgjs!(
+                    "Unable to decompress and deserialize operations in received_transactions()"
+                );
+            }
```
davidfig, line 246: Can we get test coverage on these? You can use the debug version of js.rs. If the ops are large, then use the hash_test() instead of the actual string.

davidfig (approved):
Approved with a couple of nonblocking comments.

davidkircos (changes requested):
This creates invalid transactions and breaks the sheet. May be because of other code merged from `qa` into this branch...
https://github.com/user-attachments/assets/4d27e9b0-b5fa-4b2c-bac6-ad417b32370b

quadratic-client/vite.config.js.timestamp-1722017521018-562d6b4f12d52.mjs
ddimaria: I have no clue.  It seems to be an auto-generated file from Vite: https://github.com/vitejs/vite/issues/13267

I'll delete it.

ddimaria: Deleted

quadratic-core/src/controller/execution/receive_multiplayer.rs
ddimaria: The tests for compression and serialization are in the compression.rs file.  The only reason why there is no coverage for lines 243 - 246 is because this function does not return a result (it should since there are multiple fail-able situations).  Adding result support requires a bit of a rewrite since much of core isn't returning results.  Did you want me to implement a returned result for this function?

ddimaria:
> This creates invalid transactions and breaks the sheet. May be because of other code merged from `qa` into this branch... https://github.com/user-attachments/assets/4d27e9b0-b5fa-4b2c-bac6-ad417b32370b

I can duplicate this.  You have to refresh before the files service processes the transactions.

quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
```diff
@@ -298,11 +307,20 @@ class Core {
...
         if (!this.gridController) throw new Error('Expected gridController to be defined');
-        this.gridController.receiveMultiplayerTransactions(transactions);
+
+        for (let data of receive_transactions.transactions) {
+          // convert the base64 encoded string of operations into buffers
+          if (typeof data.operations === 'string') {
+            data.operations = Buffer.from(data.operations, 'base64');
+          }
+
+          this.gridController.multiplayerTransaction(data.id, data.sequence_num, data.operations);
```
ddimaria, line 321: @davidfig it's more performant to send the transaction individually.  Do you see a problem with this?

ddimaria:
> This creates invalid transactions and breaks the sheet. May be because of other code merged from `qa` into this branch... https://github.com/user-attachments/assets/4d27e9b0-b5fa-4b2c-bac6-ad417b32370b

@davidkircos This is fixed, can you verify once the build is done?

quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
ddimaria: We had a slack conversation.  We're going to keep the individual transaction sending for 5 - 7 days after deploying this branch, then send the full payload as binary from multiplayer and just forward to core (more optimal than the current approach).

davidkircos (changes requested):
Can not download .grid file 
CleanShot 2024-08-09 at 16 17 36@2x

davidkircos:
tested a 3.3mb Parquet file
On this branch is 8mb, previously it was 48mb! This is sweet.

AyushAgrawal-A2 (commented):
Importing 0.5M file fails, this works in main and qa.
>=1M file fails on main and qa.
Files link: https://github.com/datablist/sample-csv-files

image

ddimaria:
> Can not download .grid file CleanShot 2024-08-09 at 16 17 36@2x

@davidkircos this is fixed.  Please test once the branch deploys.

davidkircos:
> > Can not download .grid file CleanShot 2024-08-09 at 16 17 36@2x
> 
> @davidkircos this is fixed. Please test once the branch deploys.

Downloading files seems to work for me!

ddimaria:
> Importing 0.5M file fails, this works in main and qa.
> 1M and above file fails on main and qa.
> Files link: https://github.com/datablist/sample-csv-files

Resolved

davidkircos (approved):

PR #1609, opened 2024-07-25
Fix string length calculation for unicode strings

davidfig:
This fixes a bug with the new sendGetCellsLength function for strings that contain unicode characters


davidfig:
@AyushAgrawal-A2 This was a fun bug. We need to get the byte length of the string in sendGetCellsLength to ensure the SharedArrayBuffer is the correct length for the characters.

AyushAgrawal-A2:
This is definitely rigged
image

PR #1612, opened 2024-07-26
feat: add keyboard shortcuts 

AyushAgrawal-A2:
part of #1536 

* [x]  Default JSON for shortcut default theme
* [x]  F2: edit cell
* [x]  Support both Ctrl + Shift + PageUp / Ctrl + Shift + PageDown **AND** option + up-arrow, option + down-arrow **AND** option + right arrow/option + left arrow: go to next/prior sheet (this covers as close as possible to all bases for both mainstream spreadsheets)
* [x]  Ctrl + d: fill down from the cell above
* [x]  Ctrl + r: fill right from the cell to the left
* [x]  Shift + space: select row
* [x]  Ctrl + space: select column

ref old PR #1545 from fork


quadratic-client/src/app/keyboard/defaults.ts
```diff
@@ -0,0 +1,464 @@
...
+      mac: ['Cmd + 0'],
+      windows: ['Ctrl + 0'],
+    },
+  },
+  {
+    action: 'save',
+    shortcuts: {
+      mac: ['Cmd + S'],
+      windows: ['Ctrl + S'],
+    },
```
davidkircos, line 82: It does not appear to work for save and run while the code editor is open. Command + Enter still does, though.

quadratic-client/src/app/keyboard/defaults.ts
```diff
@@ -0,0 +1,464 @@
...
+  },
+  {
+    action: 'delete_cell',
+    shortcuts: {
+      mac: ['Backspace', 'Delete'],
+      windows: ['Backspace', 'Delete'],
+    },
+  },
+  {
+    action: 'show_code_editor',
```
davidkircos, line 421: rename to `show_cell_type_menu`?

davidkircos (changes requested):
This is great! I like the design. 

Fix Cmd + S in the code editor and the functionality looks like it's all there

quadratic-client/src/app/keyboard/defaults.ts
AyushAgrawal-A2: done

quadratic-client/src/app/keyboard/defaults.ts
AyushAgrawal-A2: done

davidkircos (changes requested):
We need to disable command + s everywhere except when the cell editor is open. Otherwise the browser opens the save dialog when the code editor isn't open

CleanShot 2024-07-29 at 20 34 15@2x

AyushAgrawal-A2:
> We need to disable command + s everywhere except when the cell editor is open. Otherwise the browser opens the save dialog when the code editor isn't open

@davidkircos 
done.
I changed event capture for viewport when trying to get ctrl+pageup to work. forgot to remove it. now behaviour is same as before.

davidkircos (approved):
Functionality looks great to me! 

Can merge into `qa` after code review from @davidfig

davidfig (approved):

davidfig (approved):

PR #1615, opened 2024-07-26
test vercel build cache

AyushAgrawal-A2:
ref: https://github.com/orgs/vercel/discussions/222


AyushAgrawal-A2:
caching is on by default. wasm-pack and vite build are the slowest steps.

PR #1620, opened 2024-07-27
chore: fix circular dependencies

AyushAgrawal-A2:
- [ ] Fix circular dependencies https://github.com/quadratichq/quadratic/pull/1587#issue-2424703221 
- [x] Add eslint rule to import paths.
- [x] Update `.prettierrc` & `.vscode` to auto lint and format on save.


AyushAgrawal-A2:
too many changes, not worth it

PR #1624, opened 2024-07-30
Adds back cmd+r to reload the page

davidfig:


AyushAgrawal-A2:
@davidfig 
`Cmd + R` / `Ctrl + R` is now autocomplete right. This is preventing browser default and page reload does not happen now.
https://github.com/quadratichq/quadratic/blob/ed4d7c81e3558d9e1ac533d2854074007dfb3002/quadratic-client/src/app/keyboard/defaults.ts#L120-L124

davidfig:
Ah, that makes sense. I raised the issue in slack. If we want to have it, then I'll change this to a dev-only override.

quadratic-client/src/app/gridGL/interaction/keyboard/useKeyboard.ts
```diff
@@ -52,6 +53,10 @@ export const useKeyboard = (props: IProps): { onKeyDown: (event: React.KeyboardE
   }, [editorInteractionState, presentationMode, setEditorInteractionState, setPresentationMode]);
 
   const onKeyDown = (event: React.KeyboardEvent<HTMLElement>) => {
+    if (debug && (event.metaKey || event.ctrlKey) && event.key === 'r') {
+      window.location.reload();
+      return;
+    }
```
AyushAgrawal-A2: This will not work if app is stuck, we need to just disable that action in debug
I have pushed a commit for it.

AyushAgrawal-A2:
Also after this PR https://github.com/quadratichq/quadratic/pull/1612 Shortcuts should match exactly.
Earlier only part of keyboard event was checked, e.g. pressing `Enter` would open cell edit whether `Shift` or other modifier is pressed or not.
Now if the shortcut says just `Enter`, pressing modifiers along with it will not match that shortcut. All keys are checked.
Thanks for adding `Shift+Enter`

davidfig:
Yes

PR #1640, opened 2024-08-02
Add `IFERROR` and make `IF` short-circuit

HactarCE:
This also generally changes formula error handling so that's more likely for errors to propagate as values so that they can be filtered out by the user using `IF`, `IFERROR`, etc.

~~In production right now, there's a bug where a formula referencing an error cell show an error on the grid. In this PR that's fixed, but there's a new bug that all formula errors on the grid are lost when reloading the file, which is a more significant bug. @davidfig says this will be fixed in #1350. Is it worth it to make a temporary fix for this now?~~ _Fixed!_

This also fixes a bug where referencing a formula with an error would not show an error:
image


davidkircos (approved):
Functionality looks good to me!

AyushAgrawal-A2:
few error are not handled
image

HactarCE:
> few error are not handled

Good catches! All of those are intended behavior except for the NaN and `VLOOKUP`, which I've just pushed fixes & tests for.

HactarCE:
@AyushAgrawal-A2 Good catches! I've fixed a few of them and put the remaining ones in a new issue #1702 since they're not issues with `IFERROR` per se, but issues with the formula parser itself.

AyushAgrawal-A2 (approved):

PR #1656, opened 2024-08-07
feat: code editor onboarding

### LINKED ISSUE #1662: bug: Return type inspector
- [ ] Should not show the enter indicator when there is an error and no return
- [ ] JS when there is no return, should not show enter indicator on last line
- [ ] "Last line returns to the sheet" placeholder while unsaved changes

jimniels:
## Feature

Includes:

- Adds empty states to code editor when there's 0 lines of code
- Fixes #1662 
- Activates "Schema" as the default tab when you're looking at a connection cell
- New code snippets (see below) including better parity between PY/JS

TODOs

- [x] Fix "loading" issue when code editor is loading and widget immediately renders
- [x] Add mixpanel events for buttons

### New Code Snippets

I think it's important that these snippets be immediately runnable. So if you click "reference cell" it should insert a snippet that you can immediately run (even though it might not return anything).

In that spirit, I went with really short snippets that provide working code with links to the docs (you can cmd+click in the editor) e.g.

```py
my_cell = cell(0, 0)

# Learn more:
# https://docs.quadratichq.com/python/reference-cells
```

We could try to fit more in there, e.g.

```py
my_cell = cell(0, 0)

# Learn more about references , including:
#  - Relative cell: rel_cell(x, y)
#  - Other sheets: cell(x, y, 'Sheet name')
#  - Multiple cells: cells((fromX, fromY), (toX, toY))
#  - And more...
# https://docs.quadratichq.com/python/reference-cells
```

But I think it'd be good to keep it short and simple and allow people to learn more from the docs. We'll never fit all our explaining in some code comments. Best to not overburden the user. Give them something immediately useful without too much cognitive overhead with a resource to learn the kitchen sink.

### More Snippets?

We could break out each cell reference into it's own "snippet", e.g

- Reference a cell
- Reference a cell in another sheet
- Reference a cell relatively
- Reference cells
- Reference cells in another sheet
- Reference cells with slicing syntax
- Get a reference to the current cell

All of these showing up here, instead of just the single "Read data from sheet" 

But I left that out of scope here. Not sure if we want to do that.


davidkircos (approved):
This is a huge improvement to the onboarding flow!

PR #1659, opened 2024-08-07
make live viewport+neighbour rect available to core during transaction

AyushAgrawal-A2:


quadratic-client/src/app/web-workers/renderWebWorker/worker/renderCore.ts
```diff
@@ -128,7 +140,16 @@ class RenderCore {
         height,
       };
       this.renderCorePort.postMessage(message);
-      this.waitingForResponse.set(id, resolve);
+      this.waitingForResponse.set(id, (cells: JsRenderCell[]) => {
```
davidfig, line 143: How does this rejection logic work? After 1s we don't render that hash, or will it be picked up on another update loop?

quadratic-core/src/controller/operations/import.rs
```diff
@@ -58,7 +58,7 @@ impl GridController {
 
         // then create operations using MAXIMUM_IMPORT_LINES to break up the SetCellValues operations
         let mut ops = vec![] as Vec<Operation>;
-        let mut cell_values = CellValues::new(width, height);
+        let mut cell_values = CellValues::new(width, height.min(IMPORT_LINES_PER_OPERATION));
```
davidfig, line 61: Nice catch

quadratic-core/src/grid/ids.rs
```diff
@@ -43,7 +43,6 @@ impl Display for SheetId {
     }
 }
 
-#[cfg(test)]
 impl SheetId {
     pub fn test() -> Self {
```
davidfig, line 47: Are you using this outside of tests?

davidfig (approved):

quadratic-core/src/grid/ids.rs
AyushAgrawal-A2: no, but it gets indirectly called in quadratic rust shared. I need this specific sheet id now everywhere, so that sheet id matches with mock viewport buffer.

davidfig: ah. ok.

davidkircos (commented):
What should I be testing here?
What do we need to do to get comfortable merging this?

PR #1668, opened 2024-08-07
fix: number formatting bugs

AyushAgrawal-A2:
- decimal places for scientific notation is now 2 by default, can be changed
- number will should pound labels only if Overflow is set to clip or there is content in the next cell


davidkircos:
There are still bugs when setting wrap for numbers
https://github.com/user-attachments/assets/56257248-c510-435e-8d96-aea1de33ce0c

davidfig:
I added a few more tests to convertNumber (I started working on it before i saw your PR).

I also updated Rust's `to_number_display` to account for the default decimals for scientific. `to_number_display` is used by the search function so we can match a formatted number (since formatting happens in TS, I had to duplicate the TS function). This ensures that a user can search either `100` or `1.00e2` (assuming the number set to scientific) to find the same cell.

davidfig (approved):

davidkircos (approved):
great fixes!

PR #1674, opened 2024-08-08
fix: multiline formula bug

### LINKED ISSUE #1657: qa: Multiline Formula Bugs
- [x] command + Enter should not insert a second line?
- [ ] Shift + Arrow while editing a range in an inline formula doesn't work

AyushAgrawal-A2:
closes #1657


davidkircos (approved):
lgtm!

Issue #1675, opened 2024-08-08
bug: getCells broken in Safari


davidkircos:
Cell references are broken in Safari due to the SharedArrayBuffer issue

![CleanShot 2024-08-08 at 13 03 25@2x](https://github.com/user-attachments/assets/4f11c89f-5c98-4c3d-8dc8-723710335021)

anthonycorletti:
Sad to see this 😢 I hope safari is supported again soon!

<img width="1367" alt="Screenshot 2024-08-26 at 8 14 05 PM" src="https://github.com/user-attachments/assets/08fed48c-d36d-4aa2-9a5d-f7a36d2c9e20">

davidkircos:
Should be! The fix is under review

https://github.com/quadratichq/quadratic/pull/1677

PR #1677, opened 2024-08-08
fix: SharedArrayBuffer on safari

AyushAgrawal-A2:
closes: #1675


davidkircos:
This blocks:
- [ ] S3 images
- [ ] Sanity 
- [ ] All other non same origin requests

davidkircos (changes requested):
This breaks:

S3 images
Sanity
Helpdesk
All other non same origin requests

AyushAgrawal-A2:
@davidkircos
Sanity and S3 Images work now

AyushAgrawal-A2:
I had to change Radix UI's Avatar component to MUI as the former was not passing in crossOrigin attribute to img tag

AyushAgrawal-A2:
Changes:
- Added custom avatar component, removed @radix-ui/react-avatar
- Removed live chat
- fixed google tags cors error

quadratic-client/src/shared/components/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
...
+      {error ? (
+        <span style={{ ...stylePreset, ...style }}>
+          {typeof children === 'string' ? getLettersFromString(children) : children}
+        </span>
+      ) : (
+        <img
+          alt={alt}
+          src={error ? fallbackSrc : src ?? ''}
+          onError={() => setError(true)}
+          style={{ ...stylePreset, ...style }}
```
jimniels, line 35: @AyushAgrawal-A2 if _all_ of our avatars are going to need `crossOrigin=anonymous` why not just put it in here at the base component? Then whenever you use `<Avatar>` you don't have to remember to add that extra attribute.

In other words, make `crossOrigin=anonymous` opt-out rather than opt-in (is there anywhere in our code where we would even opt-out of this right now?)

jimniels: Actually same idea with `src`

It seems everywhere you're implementing Avatar you're also having to do

`src=getAuth0AvatarSrc(...)}`

so why not do this in the base component too?

quadratic-client/src/shared/components/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { ImgHTMLAttributes } from 'react';
+
+interface AvatarProps extends ImgHTMLAttributes<HTMLImageElement> {
+  size?: 'small' | 'medium' | 'large';
+  fallbackSrc?: string;
```
jimniels, line 5: What is `fallbackSrc`? Like are you actually using this somewhere?

quadratic-client/src/dashboard/components/DashboardSidebar.tsx
```diff
@@ -142,9 +143,19 @@ export function DashboardSidebar({ isLoading }: { isLoading: boolean }) {
...
+          <Avatar
+            src={getAuth0AvatarSrc(user?.picture)}
+            alt={user?.name}
+            crossOrigin="anonymous"
+            style={{
+              width: '24px',
+              height: '24px',
+              fontSize: '.8125rem',
+              borderRadius: '50%',
+              backgroundColor: colors.quadraticSecondary,
```
jimniels, line 155: If the `<Avatar>` component has built-in sizing, seems like we should just be using it rather than hard-coding these values (which are just duplicates of what's already in the component)?

`width`, `height`, and `fontSize` are all controlled by the `size` prop, so just use that, no? (and this is just the default size anyway, so you don't even need it)

and `borderRadius` is already set to `50%` in the base component, so no need for it here?

And I would just leave the `backgroundColor` here to be whatever the default is.

so I don't think you need any of these styles

quadratic-client/src/shared/components/Avatar.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { ImgHTMLAttributes } from 'react';
```
jimniels, line 1: I feel like we should leave a todo at the top of this file that talks about how this is should be the de-facto Avatar component going forward, so eventually all occurrences of `<Avatar>` elsewhere in the code base (like from MUI) should be swapped over to this one. 

That's the goal, right?

quadratic-client/src/dashboard/components/FilesListItemCore.tsx
```diff
@@ -52,9 +53,13 @@ export function FilesListItemCore({
...
         {creator && creator.name && (
           <TooltipPopover label={`Created by ${creator.name}`}>
-            <Avatar className="h-6 w-6 text-muted-foreground">
-              <AvatarImage src={creator?.picture} alt={creator.name} />
-              <AvatarFallback>{creator.name[0]}</AvatarFallback>
+            <Avatar
+              alt={creator.name}
+              src={getAuth0AvatarSrc(creator.picture)}
+              crossOrigin="anonymous"
+              className="h-6 w-6 text-muted-foreground"
```
jimniels, line 60: Seems like you can remove this `className` because it should just take on the base default styles

quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
```diff
@@ -218,28 +220,30 @@ ${QuadraticDocs}`,
...
                   alt={user?.name}
-                  src={user?.picture}
-                ></Avatar>
+                  crossOrigin="anonymous"
+                  style={{
+                    width: '24px',
+                    height: '24px',
+                    fontSize: '0.8125rem',
+                    borderRadius: '50%',
+                    backgroundColor: colors.quadraticSecondary,
```
jimniels, line 231: Similar comment to [this one](https://github.com/quadratichq/quadratic/pull/1677/files#r1727742005). I don't think you should pass any of these. They're just duplicates of what's already in `<Avatar>` (except `backgroundColor` and I think you should just let that be the default here, e.g. a dull gray)

quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
```diff
@@ -218,28 +220,30 @@ ${QuadraticDocs}`,
...
-                  sx={{
-                    bgcolor: 'white',
-                    width: 24,
-                    height: 24,
+                  alt="AI Assistant"
+                  style={{
+                    width: '24px',
+                    height: '24px',
                     fontSize: '0.8rem',
+                    borderRadius: '50%',
```
jimniels, line 243: Similar to the above. No need to override the style values already present in the base avatar component (width, height, fontSize, border-radius)

jimniels (commented):
All the stuff that fixes Safari looks good to me 👍 

Left some clarifying comments and also feedback related to the generalized `<Avatar>` component.

quadratic-client/src/shared/components/Avatar.tsx
AyushAgrawal-A2: @jimniels 
I am assuming that all request will not be cross origin, we can make it opt out, I am OK either ways

same with `src=getAuth0AvatarSrc(...)}`, I am OK, can be plugged directly into Avatar component, should be OK
it is required because of the intermediary gravatar link provided by Auth0

quadratic-client/src/shared/components/Avatar.tsx
AyushAgrawal-A2: I wanted to do level 1 fallback to another source, level 2 to child string / component, I removed it while debugging and forgot to add it back

quadratic-client/src/dashboard/components/DashboardSidebar.tsx
AyushAgrawal-A2: yup, we can clean these up... I kept it as a reference to preserve original values before my refactor

quadratic-client/src/shared/components/Avatar.tsx
AyushAgrawal-A2: I am not sure why we want to move away from MUI, I don't have a favourite. I am OK with whatever works
Also for easy components like this Avatar one, we can implement ourselves and avoid dependencies.
The remaining MUI <Avatar> components are the ones used with <AvatarGroup> parent component and have image stacking. I also do a custom component for that, if moving away from MUI is necessary.

quadratic-client/src/dashboard/components/FilesListItemCore.tsx
AyushAgrawal-A2: i think styles from these tailwind classes are different from default ones, it is the value originally used here.

quadratic-client/src/shared/components/Avatar.tsx
jimniels: what were you falling back to? some generic image somewhere?

quadratic-client/src/shared/components/Avatar.tsx
jimniels: The long-term plan is to move off MUI, so if we can remove it everywhere it won't be part of our bundle. I think that's a good goal to strive for, i.e. let's have the same `<Avatar>` component everywhere in the codebase.

> for easy components like this Avatar one, we can implement ourselves and avoid dependencies.

Precisely.

> The remaining MUI <Avatar> components are the ones used with <AvatarGroup>

I don't know that you have to do that in this PR. I just think it's good to leave a note in the component that "hey, this is the one we should be using going forward." And when we have a chance to move off the MUI avatar, let's do it! You can decide when is a good time based on your understanding of what it would take to move off MUI.

quadratic-client/src/dashboard/components/FilesListItemCore.tsx
jimniels: Remove it. h-6 and w-6 are just saying width/height 24px.

And the text color should be handled by the default <Avatar> component.

We don't need this having it's own one-off style that's different from the base <Avatar> component.

quadratic-client/src/shared/components/Avatar.tsx
AyushAgrawal-A2: I wanted to take 2 src argument... one would have been a foreign request... one local default src or some stable source..
It is not committed in the pr yet

quadratic-client/src/shared/components/Avatar.tsx
jimniels: I feel like our avatar solution right now is to always reference files that aren't stored on the same origin (e.g. images we get from auth0) so putting it in the base avatar component makes sense to me.

I know if I have to go use an <Avatar> somewhere else in the code to put somebody's picture on screen, I'm going to forget to do `src={getAuth0AvatarSrc()}` as well as add crossOrigin.

That's my thought for putting it in the base component, so we don't have to remember to do those two things on every single Avatar component we use elsewhere.

AyushAgrawal-A2: Sure, will do

quadratic-client/src/shared/components/Avatar.tsx
AyushAgrawal-A2: I don't want to do it in this PR. This was about fixing cross origin and it turned into a MUI / Radix UI debate...
Happy to do it later, I have put all SAB related work on hold until cross-origin is fixed properly.
I really don't want any more pile on on this PR. I will add a TODO note.

AyushAgrawal-A2: Will add a note

davidkircos (approved):
Functionality all seems to work for me!

I noticed an issue where JS `fetch` requests do not work in Safari, but I think this is a different issue. I believe those probably need to be proxied and are not today.

davidkircos:
AI Avatar doesn't load properly 
CleanShot 2024-08-26 at 20 48 11@2x

davidkircos (changes requested):
see previous comments

AyushAgrawal-A2:
>AI Avatar doesn't load properly

@davidkircos
fixed

davidkircos (approved):

PR #1682, opened 2024-08-09
fix: inter hash text clipping

AyushAgrawal-A2:


davidfig:
Which bug # does this fix?

AyushAgrawal-A2:
@davidfig 
[Screencast from 09-08-24 03_06_24 AM IST.webm](https://github.com/user-attachments/assets/2a1fe8fb-0495-4f58-a9fb-8bdcc77d6eb2)

davidfig (approved):

Issue #1673, opened 2024-08-08
Cursor not centered properly when opening sheet


luke-quadratic:
**Prod** 
![CleanShot 2024-08-08 at 09 44 57@2x](https://github.com/user-attachments/assets/74d9b584-006f-4d8d-b114-d9dd676f66d1)

**QA** 
![CleanShot 2024-08-08 at 09 43 11@2x](https://github.com/user-attachments/assets/c743372e-fb7b-4f0c-a0e1-e7d31eb7553f)

luke-quadratic:
Resolved in #1683

PR #1683, opened 2024-08-09
start viewport at + 1 to ensure the full cursor shows

davidfig:
This fixes the starting viewport so the cursor shows properly at 0,0


luke-quadratic:
confirming functionality, closes #1673

davidkircos (approved):

PR #1688, opened 2024-08-09
fix: flash of content when creating formulas

### LINKED ISSUE #1661: Flash of content when creating Formulas
Whenever you create a Formula there's a weird flash of where you last created a Formula - this has existed for a long time, had thought it was captured in an issue but the issue in mind (#1366) is different but possibly related. 

Exists in both prod and QA.

AyushAgrawal-A2:
closes #1661


davidkircos (commented):
Functionality looks good to me! 

@luke-quadratic waiting on your feedback to move forward.

luke-quadratic (approved):
This works as expected, lgtm!

PR #1692, opened 2024-08-10
fix: drag to move papercuts

### LINKED ISSUE #1671: Drag to move papercuts
- Disable drag and drop on the same location, do nothing instead of copying cell contents.
- Disable right click for drag to move a cell. Needs to be left click.

AyushAgrawal-A2:
closes #1671


davidkircos (approved):
This works great!

PR #1693, opened 2024-08-10
fix: when dragging on grid, ignore charts

### LINKED ISSUE #1686: nit: When dragging on grid, ignore charts
Pass pointer events through charts when mouse down dragging on canvas

https://github.com/user-attachments/assets/0fc2f46b-f37e-43a8-975c-9503334fd71a

AyushAgrawal-A2:
closes #1686

### DIFF
--- a/quadratic-client/src/app/gridGL/interaction/pointer/PointerHtmlCells.ts
+++ b/quadratic-client/src/app/gridGL/interaction/pointer/PointerHtmlCells.ts
@@ -142,8 +142,9 @@ export class PointerHtmlCells {
       if (target === 'body' && cell === active) {
         this.setDoubleClick();
         this.clicked = cell;
+        return true;
       }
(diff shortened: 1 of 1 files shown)


davidkircos (changes requested):
If you mouseUp over the graph the dragging isn't turned off.

I've found the most effective way to deal with this is to attach the mouseUp event to the window because then it is always triggered, even if the user drags off screen

AyushAgrawal-A2:
> If you mouseUp over the graph the dragging gets stuck on.

Fixed

> I've found the most effective way to deal with this is to attach the mouseUp event to the window because then it is always triggered, even if the user drags off screen

It is already there

davidkircos (approved):

davidkircos:
Nice fix @AyushAgrawal-A2 !

PR #1695, opened 2024-08-10
fix: number of decimal digits

AyushAgrawal-A2:


davidkircos (changes requested):
Long numbers don't render 


https://github.com/user-attachments/assets/abad1584-ef29-45da-a22d-b0c9c653c821

AyushAgrawal-A2:
>Long numbers don't render

Fixed

davidkircos (approved):
lgtm nice fixes!

PR #1700, opened 2024-08-12
feat: mssql connection

AyushAgrawal-A2:


quadratic-rust-shared/src/sql/mssql_connection.rs
```diff
@@ -0,0 +1,613 @@
...
+            .await
+            .map_err(|e| SharedError::Sql(Sql::Schema(e.to_string())))?;
+
+        let mut schema = DatabaseSchema {
+            database: self.database.to_owned().unwrap_or_default(),
+            tables: BTreeMap::new(),
+        };
+
+        for row in rows {
+            let schema_name: &str = row.get(1).unwrap_or_default();
```
ddimaria, line 193: Default might be an empty string here.  Maybe some other placeholder would be appropriate (e.g. `Unknown Schema - #num`, `Unknown Column - #num`), where `#num` is the loop index.

quadratic-rust-shared/src/sql/mssql_connection.rs
```diff
@@ -0,0 +1,613 @@
...
+                }
+            },
+        )
+    }
+}
+
+#[macro_export]
+macro_rules! convert_mssql_type {
+    ( $kind:ty, $column_data:ident ) => {{
+        <$kind>::from_sql($column_data)
```
ddimaria, line 273: Same notes here on the default value.

quadratic-rust-shared/src/sql/mod.rs
```diff
@@ -270,14 +88,12 @@ pub trait Connection {
...
-            .columns()
-            .iter()
+        let fields = Self::row_columns(&data[0])
             .enumerate()
             .map(|(index, col)| {
                 Field::new(
-                    col.name().to_string(),
-                    cols[index].data_type().to_owned(),
+                    Self::column_name(&col).to_string(),
+                    cols[index].data_type().clone(),
```
ddimaria, line 96: Nit: I'll write clones when starting a PR, but then remove them or convert to `to_owned()` to signal the intentional copy.  It's not that important here, but that's the pattern I've been doing.

davidkircos:
set default username to `admin`
CleanShot 2024-08-26 at 21 26 49

davidkircos:
Should a database name be required? I believe by default AWS RDS does not create a database

CleanShot 2024-08-26 at 21 27 50

davidkircos:
Schema viewer does not load 

CleanShot 2024-08-26 at 21 37 11

davidkircos:
SQL Syntax highlighting doesn't work for MSSQL 
CleanShot 2024-08-26 at 21 39 19

quadratic-rust-shared/src/sql/mssql_connection.rs
AyushAgrawal-A2: switched default value to ArrowType::Void

quadratic-rust-shared/src/sql/mssql_connection.rs
AyushAgrawal-A2: done

AyushAgrawal-A2:
>Should a database name be required? I believe by default AWS RDS does not create a database

it defaults to `master` if not provided, i have made it optional now

ddimaria (approved):

Issue #1678, opened 2024-08-08
Precision for Formula results resolving to huge numbers


luke-quadratic:
In prod we round properly, in QA we introduced something that is causing us to have these huge numbers by default on Formula results

To replicate:
`=AVERAGE(1,1,3)`

![CleanShot 2024-08-08 at 17 12 10](https://github.com/user-attachments/assets/308d4b55-6cf7-47c4-9a6d-60424a5f75fd)

AyushAgrawal-A2:
Precision also looks bad here, after decimal all digits should be 6 and last 7.

PR #1708, opened 2024-08-12
Remove nonsense digits for `f64 -> BigDecimal`

HactarCE:
This should provide another layer of protection against #1678

Thank you @AyushAgrawal-A2 for pushing for this! I didn't realize the extra digits were coming from Rust.

### DIFF
--- a/quadratic-core/src/values/convert.rs
+++ b/quadratic-core/src/values/convert.rs
@@ -6,6 +6,8 @@ use crate::{CodeResult, CodeResultExt, RunErrorMsg, Span, Spanned, Unspan};
 
 const CURRENCY_PREFIXES: &[char] = &['$', '¥', '£', '€'];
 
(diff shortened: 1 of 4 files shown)


AyushAgrawal-A2:
I have removed the js (renderer) patch for limiting decimal digits, as values received rust are now accurate.

AyushAgrawal-A2 (approved):

quadratic-core/src/values/convert.rs
```diff
@@ -6,6 +6,8 @@ use crate::{CodeResult, CodeResultExt, RunErrorMsg, Span, Spanned, Unspan};
 
 const CURRENCY_PREFIXES: &[char] = &['$', '¥', '£', '€'];
 
+const F64_DECIMAL_PRECISION: u64 = 16; // just enough to not lose information
```
ddimaria, line 9: @davidfig Is this precision OK with you?

ddimaria (approved):

PR #1715, opened 2024-08-14
feat: file import from New file dialog

AyushAgrawal-A2:
- [x] File import from new dialog box
- [x] Redirect page based on upload context
- [x] Drag and drop on dashboard
- [x] Import progess
- [x] Update UI
- [x] Cancel upload

Pending (future PR)
- [ ] Cancel import in core
- [ ] Grid import, without upgrade (TBD)


davidkircos (changes requested):
When importing a single file, just open it after importing.

CleanShot 2024-08-19 at 17 31 35@2x

davidkircos (changes requested):
Show import drop files here only over files side of the dashboard. Don't overlay over the nav bar
CleanShot 2024-08-19 at 17 32 20@2x

davidkircos (changes requested):
Functionality works great! 

Needs to be updated with the base branch, and rust code review.

quadratic-client/src/app/gridGL/loadAssets.ts
```diff
@@ -36,17 +36,23 @@ export function loadAssets(): Promise<void> {
...
+    addResourceOnce('python-icon', '/images/python-icon.png');
+    addResourceOnce('javascript-icon', '/images/javascript-icon.png');
 
     // Wait until pixi fonts are loaded before resolving
     Loader.shared.load(() => ensureBitmapFontLoaded(resolve));
   });
 }
+
+function addResourceOnce(name: string, url: string) {
+  if (!Loader.shared.resources[name]) {
```
ddimaria, line 55: Do we want some error or a log to show that the resource doesn't exist?

quadratic-client/src/app/helpers/files.ts
```diff
@@ -59,3 +57,29 @@ export function isGrid(file: File): boolean {
 export function isParquet(file: File): boolean {
   return file.type === 'application/vnd.apache.parquet' || hasExtensions(file.name, ['parquet', 'parq', 'pqt']);
 }
+
+export const getFileType = (file: File) => {
```
ddimaria, line 61: I think moving these here makes a lot of sense.

quadratic-client/src/router.tsx
```diff
@@ -36,7 +36,7 @@ export const router = createBrowserRouter(
               lazy={() => import('./routes/file.$uuid')}
               // We don't want to revalidate the initial file route because
               // we don't have any 2-way data flow setup for the file contents
-              shouldRevalidate={() => false}
+              shouldRevalidate={({ currentParams, nextParams }) => currentParams.uuid !== nextParams.uuid}
```
jimniels, line 39: @AyushAgrawal-A2 for my own edification, why this change? if the file uuid is changing (in the app) then you're navigating to a new place and that should be a whole document reload, no?

quadratic-client/src/shared/shadcn/ui/progress.tsx
```diff
@@ -9,11 +9,11 @@ const Progress = React.forwardRef<
 >(({ className, value, ...props }, ref) => (
   <ProgressPrimitive.Root
     ref={ref}
-    className={cn('relative h-2 w-full overflow-hidden rounded-full bg-primary/20', className)}
+    className={cn('relative h-4 w-full overflow-hidden rounded bg-[#F2F5F9]', className)}
```
jimniels, line 12: Why the hard-coded value? Were you pulling this from Figma? Feel like we should stick to `bg-primary/20`

jimniels (approved):
Reviewed the ts. Two comments but this LGTM. Will review more of the functionality when this gets merged into the base branch and we use it along with all the other functionality.

quadratic-client/src/shared/shadcn/ui/progress.tsx
AyushAgrawal-A2: @jimniels 
yup, these were direct figma values, didn't know we have tailwind equivalents

jimniels: Yeah anything in Figma is just an approximation of what the base shad/cn values are doing. So let's keep it at `bg-primary/20`

quadratic-client/src/router.tsx
AyushAgrawal-A2: @jimniels 
When creating a new file from App (i.e. open file) using new dialog box, the document needs to reload, if we want to switch to that file.
I was not reloading without this change

quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
```diff
@@ -400,55 +402,148 @@ class Core {
...
-          resolve(undefined);
-        } catch (error) {
-          // TODO(ddimaria): standardize on how WASM formats errors for a consistent error
-          // type in the UI.
-          console.error(error);
-          reportError(error);
-          Sentry.captureException(error);
-          resolve(error as string);
-        }
+    sequenceNum: number
```
ddimaria, line 407: There is a fair amount of duplication in this file that it could merit some abstractions to DRY things up.

quadratic-client/src/app/gridGL/loadAssets.ts
AyushAgrawal-A2: Nope, the issue was when opening a new file from App, resources were being loaded again, which resulting in error.
We did not have the functionality of opening a new file from the app (i.e. when another file is already open).

quadratic-client/src/shared/shadcn/ui/progress.tsx
AyushAgrawal-A2: done

quadratic-core/src/controller/execution/execute_operation/mod.rs
```diff
@@ -49,7 +49,7 @@ impl GridController {
                 }
             }
 
-            if (cfg!(target_family = "wasm") || cfg!(test)) && !transaction.is_server() {
+            if cfg!(target_family = "wasm") || cfg!(test) {
```
ddimaria, line 52: @davidfig You know more about the `transaction.is_server()` code, are the changes in these files all good?

quadratic-core/src/controller/execution/receive_multiplayer.rs
```diff
@@ -79,10 +80,15 @@ impl GridController {
...
-    pub fn server_apply_transaction(&mut self, operations: Vec<Operation>) {
+    pub fn server_apply_transaction(
+        &mut self,
+        operations: Vec<Operation>,
+        transaction_name: Option<TransactionName>,
+    ) {
         let mut transaction = PendingTransaction {
             transaction_type: TransactionType::Server,
             operations: operations.into(),
+            transaction_name: transaction_name.unwrap_or(TransactionName::Unknown),
```
ddimaria, line 91: Should we make this an error case or in some way notify via `dbgjs!` so that this un-anticipated state is known?

quadratic-core/src/controller/mod.rs
```diff
@@ -52,6 +52,10 @@ impl GridController {
         &mut self.grid
     }
 
+    pub fn into_grid(self) -> Grid {
```
ddimaria, line 55: Consider implementing the From/Into trait as it's more idiomatic.

quadratic-core/src/controller/operations/import.rs
```diff
@@ -150,6 +148,18 @@ impl GridController {
             y: row as i64 + 1,
         };
 
+        // total rows for calculating import progress
+        let total_rows = sheets
```
ddimaria, line 152: Nice 💯

quadratic-core/src/controller/user_actions/import.rs
```diff
@@ -1,26 +1,35 @@
...
+            self.server_apply_transaction(ops, Some(TransactionName::Import));
+        }
+
         Ok(())
     }
 
     /// Imports an Excel file into the grid.
     ///
-    /// Returns a [`TransactionSummary`].
+    /// Using `cursor` here also as a flag to denote import into new / existing file.
```
ddimaria, line 32: Much better doc message 🙌

quadratic-core/src/grid/file/mod.rs
```diff
@@ -67,37 +67,40 @@ impl GridFile {
...
 }
 
-fn import_json(file_contents: &str) -> Result<Grid> {
-    let file = serde_json::from_str::<GridFile>(file_contents)
-        .map_err(|e| {
-            dbg!(&e);
-            anyhow!(e)
-        })?
-        .into_latest()?;
+fn import_json(file_contents: String) -> Result<Grid> {
```
ddimaria, line 93: We previously avoided allocating to the heap.  What's the justification for changing this?

quadratic-core/src/grid/file/v1_6/file.rs
```diff
@@ -8,42 +8,42 @@ use crate::{
...
                 CodeCellLanguage::Javascript => current::CodeCellLanguage::Javascript,
-                CodeCellLanguage::Connection { kind, ref id } => {
+                CodeCellLanguage::Connection { kind, id } => {
                     current::CodeCellLanguage::Connection {
                         kind: match kind {
                             ConnectionKind::Postgres => current::ConnectionKind::Postgres,
                             ConnectionKind::Mysql => current::ConnectionKind::Mysql,
                         },
-                        id: id.clone(),
+                        id,
```
ddimaria, line 29: Nice!

quadratic-rust-shared/src/error.rs
```diff
@@ -60,7 +60,7 @@ pub enum SharedError {
     #[error("Error with Pubsub: {0}")]
     PubSub(String),
 
-    #[error("Error with Reqwest: {0}")]
+    #[error("Error with Request: {0}")]
```
ddimaria, line 63: `Reqwest` was intentional before as that's the name of the crate used.

ddimaria (commented):
Looks great, just some comments and questions, mostly non-blocking.  I like the move away from sending mutable references for most of the functions 💯

quadratic-core/src/grid/file/mod.rs
AyushAgrawal-A2: to take ownership, convert to json and drop it immediately releasing memory

quadratic-core/src/controller/mod.rs
AyushAgrawal-A2: here into_ is used in line with rust convention when we are consuming original value, here I am consuming the grid from controller, to take ownership and consume data rather than cloning, to save memory

quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
```diff
@@ -40,7 +40,7 @@ impl GridController {
         op: Operation,
     ) {
         if let Operation::AddSheetSchema { schema } = op {
-            if let Ok(sheet) = schema.into_latest() {
+            if let Ok(sheet) = schema.clone().into_latest() {
```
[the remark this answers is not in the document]
AyushAgrawal-A2: schema is also pushed to forward transactions, a clone is required here

quadratic-core/src/controller/execution/control_transaction.rs
```diff
@@ -2,26 +2,21 @@ use chrono::Utc;
...
+use crate::error_core::Result;
+use crate::grid::{CodeRun, CodeRunResult};
+use crate::parquet::parquet_to_vec;
+use crate::{Pos, Value};
 
 impl GridController {
     // loop compute cycle until complete or an async call is made
     pub(super) fn start_transaction(&mut self, transaction: &mut PendingTransaction) {
-        if cfg!(target_family = "wasm") && !transaction.is_server() {
+        if cfg!(target_family = "wasm") {
```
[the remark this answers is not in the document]
davidfig: Should be fine. Only server uses ::Server, and it doesn't need any client updates.

quadratic-core/src/controller/execution/receive_multiplayer.rs
davidfig: transaction_name is not super important. We don't even use it yet. (I was hoping to use it as part of Kernel to report on what core was doing--but never got to it.)

quadratic-core/src/controller/operations/import.rs
```diff
@@ -22,23 +20,23 @@ impl GridController {
     pub fn import_csv_operations(
         &mut self,
         sheet_id: SheetId,
-        file: &[u8],
+        file: Vec<u8>,
```
davidfig, line 23: Why the move away from slice?

quadratic-core/src/controller/user_actions/import.rs
```diff
@@ -1,26 +1,35 @@
...
 
 impl GridController {
     /// Imports a CSV file into the grid.
+    ///
+    /// Using `cursor` here also as a flag to denote import into new / existing file.
     pub fn import_csv(
         &mut self,
         sheet_id: SheetId,
-        file: &[u8],
+        file: Vec<u8>,
```
davidfig, line 15: ditto on the slice

quadratic-core/src/controller/user_actions/import.rs
davidfig, line 32: Was the overload of cursor to avoid redoing the messages? I'm okay with it, but usually better to add another variable.

quadratic-core/src/grid/file/current.rs
```diff
@@ -679,7 +679,65 @@ pub fn export_rows_size(sheet: &Sheet) -> Vec<(i64, current::Resize)> {
         .collect()
 }
 
-pub(crate) fn export_sheet(sheet: &Sheet) -> current::Sheet {
+pub fn export_rows_code_runs(sheet: &Sheet) -> Vec<(current::Pos, current::CodeRun)> {
```
davidfig, line 682: Nice change. If you're going to pull these out, it's worth a test (assuming you also pulled out import_row_code_runs.

davidfig (approved):

quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
AyushAgrawal-A2: Will do cleanup directly in jim's onboarding branch

quadratic-core/src/controller/operations/import.rs
AyushAgrawal-A2: to take ownership, use data without cloning and drop the rest asap, this saves memory usage

quadratic-core/src/controller/user_actions/import.rs
AyushAgrawal-A2: cursor will only be there when importing into existing sheet, felt redundant to add a boolean flag to denote this

quadratic-core/src/grid/file/current.rs
AyushAgrawal-A2: this was the only operation in export_sheet function, which was not separate... now it matches the rest of the exports

quadratic-rust-shared/src/error.rs
AyushAgrawal-A2: reverted

quadratic-core/src/controller/mod.rs
ddimaria: This pedantic, so we can skip.  But I would think you're going from a GridController into a Grid.  Another way to implement this is to implement the deref trait for GridController, assuming it's just a wrapper around a Grid.

ddimaria (approved):

PR #1722, opened 2024-08-18
fix: simple file storage

AyushAgrawal-A2:


AyushAgrawal-A2:
need to fix CORS at S3, this give a opaque response

PR #1726, opened 2024-08-19
fix: show sql error

### LINKED ISSUE #1705: bug: Sql cells don't show ERROR when they are in the ERROR state
CleanShot 2024-08-12 at 11 42 59@2x

AyushAgrawal-A2:
closes #1705


davidkircos (approved):
Functionality looks good! Waiting on code review

quadratic-core/src/controller/execution/control_transaction.rs
```diff
@@ -2,21 +2,16 @@ use chrono::Utc;
...
-        transaction::Transaction,
-        transaction_summary::{CELL_SHEET_HEIGHT, CELL_SHEET_WIDTH},
-        transaction_types::JsCodeResult,
-    },
-    error_core::Result,
-    grid::{CodeRun, CodeRunResult},
-    parquet::parquet_to_vec,
-    Pos, Value,
-};
+use crate::controller::active_transactions::pending_transaction::PendingTransaction;
```
ddimaria, line 5: Was this copilot, or do you just prefer an import style?

ddimaria: We should agree on a consistent format as a team to avoid these hard-to-read diffs.

ddimaria (approved):

quadratic-core/src/controller/execution/control_transaction.rs
AyushAgrawal-A2: I update rust toolchain to nightly and since then linting is different... these happen automatically on save

ddimaria: @AyushAgrawal-A2 I don't think we want to be on nightly.  Was there a specific reason for it?

AyushAgrawal-A2: @ddimaria I was trying bulk memory and rust concurrency in wasm
toolchain is local but it messed up my linting style

PR #1727, opened 2024-08-19
Proper fix for formula fetching cells from grid

HactarCE:
This is a proper fix for #1716 plus a test to ensure this doesn't happen again

### DIFF
--- a/quadratic-core/src/controller/mod.rs
+++ b/quadratic-core/src/controller/mod.rs
@@ -65,7 +65,8 @@ impl GridController {
             .map(|t| &t.forward)
     }
 
-    pub fn test_blank() -> Self {
+    #[cfg(test)]
+    pub fn new_blank() -> Self {
         Self::from_grid(Grid::new_blank(), 0)
     }
 }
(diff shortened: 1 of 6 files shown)


quadratic-core/src/controller/mod.rs
```diff
@@ -65,7 +65,8 @@ impl GridController {
             .map(|t| &t.forward)
     }
 
-    pub fn test_blank() -> Self {
+    #[cfg(test)]
```
ddimaria, line 68: 🙌

quadratic-core/src/grid/sheet.rs
```diff
@@ -241,7 +241,21 @@ impl Sheet {
...
     pub fn get_cell_for_formula(&self, pos: Pos) -> CellValue {
-        self.display_value(pos).unwrap_or(CellValue::Blank)
+        let cell_value = self
+            .get_column(pos.x)
+            .and_then(|column| column.values.get(&pos.y));
+
+        if let Some(cell_value) = cell_value {
+            match cell_value {
+                CellValue::Blank | CellValue::Code(_) => match self.code_runs.get(&pos) {
+                    Some(run) => run.get_cell_for_formula(0, 0),
```
ddimaria, line 251: Out of curiosity, why is this set to `(0. 0)`?

ddimaria (approved):

davidkircos (approved):
Lgtm! Cool to see errors moving correctly from Python > Formulas, checkable with IFERROR

quadratic-core/src/grid/sheet.rs
HactarCE: If there is a code cell _at_ the given location, then we just need the very first value in its output

PR #1742, opened 2024-08-20
Add `UNIQUE`, `SORT`, and `FILTER`

HactarCE:
- [x] `UNIQUE`
- [x] `FILTER`
- [x] `SORT`


quadratic-core/src/formulas/functions/array.rs
```diff
@@ -0,0 +1,601 @@
...
+                "FILTER(A1:C5, {FALSE; TRUE; TRUE; FALSE; TRUE})"
+            )]
+            fn FILTER(
+                span: Span,
+                array: (Spanned<Array>),
+                include: (Spanned<Array>),
+                if_empty: (Option<Value>),
+            ) {
+                // Construct this `Result`, but don't return an error yet.
+                let empty_result = if_empty.ok_or_else(|| RunErrorMsg::EmptyArray.with_span(span));
```
ddimaria, line 44: You could make this a closure since it isn't always used.

quadratic-core/src/formulas/functions/array.rs
```diff
@@ -0,0 +1,601 @@
...
+                sort_order: (Option<Spanned<i64>>),
+                by_column: (Option<bool>),
+            ) {
+                let axis = by_column_to_axis(by_column);
+
+                let index = match sort_index {
+                    None => 0, // default to first column
+                    Some(value) => {
+                        let max_index = array.size()[axis.other_axis()].get() as i64;
+                        if (1..=max_index).contains(&value.inner) {
```
ddimaria, line 123: IDK range had a `contains` function ❤️

quadratic-core/src/formulas/functions/array.rs
```diff
@@ -0,0 +1,601 @@
...
+                        _ => return Err(RunErrorMsg::InvalidArgument.with_span(value.span)),
+                    },
+                };
+
+                Array::from_slices(
+                    span,
+                    axis,
+                    array.slices(axis).sorted_by(|slice1, slice2| {
+                        compare_fn(
+                            slice1.get(index).expect("already checked bounds"),
```
ddimaria, line 146: It's been a while since I looked at formula code...are panics OK here?

quadratic-core/src/formulas/functions/array.rs
```diff
@@ -0,0 +1,601 @@
...
+fn by_column_to_axis(by_column: Option<bool>) -> Axis {
+    match by_column {
+        Some(true) => Axis::X,
+        None | Some(false) => Axis::Y,
+    }
+}
+
+#[cfg(test)]
+#[serial_test::parallel]
+mod tests {
```
ddimaria, line 227: Wow, great tests 💯

quadratic-core/src/formulas/tests.rs
```diff
@@ -62,6 +63,22 @@ pub(crate) fn check_syntax_to_err(grid: &Grid, s: &str) -> RunError {
     try_check_syntax(grid, s).expect_err("expected error")
 }
 
+/// Parses a date from a string such as `2024-12-31`.
+#[track_caller]
+pub(crate) fn date(s: &str) -> CellValue {
```
ddimaria, line 68: This functionality might already exist outside of formulas.  Are these `unwraps()` intentional?

quadratic-core/src/values/array.rs
```diff
@@ -164,6 +164,44 @@ impl Array {
...
+    }
+    fn try_from_slices<'a>(
+        axis: Axis,
+        slices: impl IntoIterator<Item = Vec<&'a CellValue>>,
+    ) -> Option<Self> {
+        let slices = slices.into_iter().collect_vec();
+        let main_len = slices.len() as u32;
+        let other_len = slices.first()?.len() as u32;
+        let size = ArraySize::new(other_len, main_len)?;
+        let a = Self::new_row_major(size, slices.into_iter().flatten().cloned().collect()).ok();
```
ddimaria, line 198: Is this cloning necessary here?

quadratic-core/src/values/cellvalue.rs
```diff
@@ -1,4 +1,6 @@
-use std::{fmt, str::FromStr};
+use std::fmt;
```
ddimaria, line 1: Good refactors in this file ❤️.  I'll be able to reuse some of this for data table filter/sort.

quadratic-core/src/values/cellvalue.rs
```diff
@@ -540,13 +545,31 @@ impl CellValue {
...
+            CellValue::Number(n) => CellValueHash::Number(n.clone()),
+            CellValue::Logical(b) => CellValueHash::Logical(*b),
+            CellValue::Instant(Instant { seconds }) => {
+                CellValueHash::Instant(seconds.to_ne_bytes())
+            }
+            CellValue::Duration(Duration { months, seconds }) => {
+                CellValueHash::Duration(*months, seconds.to_ne_bytes())
+            }
+            CellValue::Error(e) => CellValueHash::Error(e.msg.clone()),
+            _ => CellValueHash::Unknown(self.type_id()),
```
ddimaria, line 563: I'm curious about this default case.  Do we not care about hashes of the remaining CellValue variants?

ddimaria (approved):
Great PR with thorough documentation 💯   I just left some non-blocking questions/comments.

luke-quadratic (approved):
works great

(testing included UNIQUE, SORT, and FILTER; any other possible side effects of this PR if it's touching anything else was untested)

quadratic-core/src/values/cellvalue.rs
HactarCE: The hashing is just an optimization. It's technically valid to have everything hash to the same value.

The remaining `CellValue` variants are `Html`, `Code`, and `Image`, which aren't relevant here IMO. `Code` should never show up, but maybe we should handle `Html` and `Image`.

quadratic-core/src/formulas/tests.rs
HactarCE: These are just convenience functions for tests, so `.unwrap()` is fine. @davidfig may know if there's a similarly convenient way to construct these; IIRC I've mostly seen `chrono::NaiveTime::from_str()` used inline.

quadratic-core/src/values/array.rs
HactarCE: Yep. `Array::slices()` does so much rearranging of cell values that it's tricky to make it move all the `CellValue`s out of the original `Array`. A single `.clone()` on each `CellValue` at the end shouldn't be a huge deal.

quadratic-core/src/formulas/functions/array.rs
HactarCE: `.with_span()` is trivial, probably 1 or 2 assembly instructions. It'll get optimized away if not needed.

quadratic-core/src/formulas/functions/array.rs
HactarCE: No, but it's very inconvenient to handle an error here and we already checked the bounds so it shouldn't ever panic here.

PR #1743, opened 2024-08-20
fix: transaction would not save for manually resized rows having wrap

AyushAgrawal-A2:


davidfig:
Can you add test to catch the case that caused the problem

luke-quadratic:
Fix worked: approved on functionality side

davidfig (approved):

PR #1751, opened 2024-08-22
fix: cell highlighting not working from Python/JS

### LINKED ISSUE #1728: Bug: cell highlighting not working from Python/JS
At some point we regressed on cell highlighting which is now broken. Prior behavior is the cell referenced would get highlighted after running code. 

CleanShot 2024-08-19 at 16 31 48@2x

AyushAgrawal-A2:
closes #1728


luke-quadratic:
Works as intended, one blocking note is on the flash when you switch cells being highlighted, if super hard to fix can make non-blocking, otherwise should fix this

ddimaria (approved):
Code looks good!

luke-quadratic (approved):

davidkircos (approved):
@AyushAgrawal-A2 Great fix! I was missing this feature!!

PR #1752, opened 2024-08-22
fix: disable command + r while in a code editor

### LINKED ISSUE #1720: nit: disable command + r while in a code editor
It awkwardly copies the cell to the left if you do this short cut while in the code editor

AyushAgrawal-A2:
closes #1720


davidkircos:
Nice this works!

We may want a more robust way to decide which commands should or should not be allowed when on different screens. But can punt on that

davidkircos (approved):

PR #1764, opened 2024-08-22
for debugging multiplayer jwt error on QA branch

AyushAgrawal-A2:


AyushAgrawal-A2:
fixed after forced page refresh in #1765

PR #1769, opened 2024-08-23
fix: excel import in existing file

AyushAgrawal-A2:
excel import in existing file broke while refactoring, due to incorrect argument order in `quadraticCore.importFile` function


davidkircos (approved):
Functionality works!

ddimaria (approved):

PR #1776, opened 2024-08-23
fix: double-click to inline edit scrolls the sheet incorrectly

### LINKED ISSUE #1770: bug: double-click to inline edit scrolls the sheet incorrectly
Step to reproduce:
- Scroll down in a sheet.
- Double click a cell to edit.
- Sheet scroll and this cell is out of view.
- On scrolling back, same cell is in editing mode.

AyushAgrawal-A2:
closes #1770

### DIFF
--- a/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
+++ b/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
@@ -95,6 +95,7 @@ class InlineEditorMonaco {
     if (!this.editor) {
       throw new Error('Expected editor to be defined in focus');
     }
+    inlineEditorHandler.keepCursorVisible();
     this.editor.focus();
   };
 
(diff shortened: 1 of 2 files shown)


davidfig:
How do you show the cell when out of view?

quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
```diff
@@ -95,6 +95,7 @@ class InlineEditorMonaco {
     if (!this.editor) {
       throw new Error('Expected editor to be defined in focus');
     }
+    inlineEditorHandler.keepCursorVisible();
```
AyushAgrawal-A2, line 98: @davidfig 
keep cursor visible is called here,
focus is already being called in atom effects.
when we focus inline editor, we also want it to be visible, so made sense to me to just add it here

davidfig (approved):

davidkircos (approved):

PR #1781, opened 2024-08-23
Adds autocomplete for dates, times, and date times.

davidfig:
Merges into datetime for autocomplete. I kept it separate for ease of code review.


luke-quadratic:
This works as intended, can't find any bugs but didn't think of/try many edge cases

quadratic-core/src/grid/series/date_series.rs
```diff
@@ -0,0 +1,393 @@
+use chrono::{Datelike, Duration, NaiveDate};
+
+use crate::CellValue;
+
+use super::{copy_series, SeriesOptions};
+
+pub(crate) type Diff = (i32, i32, i32);
+
+/// Find the date difference between two dates.
+pub(crate) fn date_diff(start: &NaiveDate, end: &NaiveDate) -> Diff {
```
ddimaria, line 10: Several gaps in coverage in this file, it might be good to cover those branches.

quadratic-core/src/grid/series/date_series.rs
```diff
@@ -0,0 +1,393 @@
...
+        return;
+    };
+    *last = new_last;
+
+    // Adjust the day after setting year and month
+    *last += Duration::days(days as i64);
+}
+
+// Helper function to get the number of days in a given month
+fn days_in_month(year: i32, month: u32) -> u32 {
```
ddimaria, line 75: [signed_duration_since](https://docs.rs/chrono/0.4/chrono/naive/struct.NaiveDate.html#method.signed_duration_since) could be used to remove some of this code I think.

quadratic-core/src/grid/series/mod.rs
```diff
@@ -0,0 +1,166 @@
+use date_series::find_date_series;
```
ddimaria, line 1: I think if you applied the below approach, the history should be captured here and I'd only see the changes:
```shell
git mv OLD_FILE.name NEW_FILE.name
git commit -m 'Rename OLD_FILE.name to NEW_FILE.name.'
```

quadratic-core/src/grid/series/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+use date_time_series::find_date_time_series;
```
ddimaria, line 2: I'm skipping reviewing mod.rs, string_series.rs and number_series.rs since they are just copies from the deleted single file.

ddimaria (approved):

PR #1782, opened 2024-08-23
Fix console.log bug in JS

davidfig:
Properly handles a console.log where there is no return statement. 

Sample code in code editor:
```js
console.log(1)
```

Cause: the return lineNumber was -1, which caused rust to panic when unpacking because it expected an u32.


AyushAgrawal-A2 (approved):

luke-quadratic:
Works as intended!

PR #1790, opened 2024-08-24
fix: pressing arrow left next to = adds ref for left cell

### LINKED ISSUE #1758: nit: typing = then pressing left arrow should select cell to the left not move cursor over

AyushAgrawal-A2:
closes #1758

### DIFF
--- a/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorKeyboard.ts
+++ b/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorKeyboard.ts
@@ -19,7 +19,6 @@ class InlineEditorKeyboard {
 
   private handleArrowHorizontal = async (isRight: boolean, e: KeyboardEvent) => {
     const target = isRight ? inlineEditorMonaco.getLastColumn() : 2;
-    const charBeforeCursor = inlineEditorMonaco.getNonWhitespaceCharBeforeCursor();
     if (inlineEditorHandler.isEditingFormula()) {
       if (inlineEditorHandler.cursorIsMoving) {
         e.stopPropagation();
@@ -27,20 +26,16 @@ class InlineEditorKeyboard {
         keyboardPosition(e);
       } else {
         const column = inlineEditorMonaco.getCursorColumn();
-        if (column === target) {
-          e.stopPropagation();
-          e.preventDefault();
-          if (inlineEditorFormula.wantsCellRef()) {
-            if (isRight || charBeforeCursor === '=') {
-              inlineEditorHandler.cursorIsMoving = true;
-              inlineEditorFormula.addInsertingCells(column);
-              keyboardPosition(e);
-            }
-          }
-          // if we're not moving and the formula is valid, close the editor
-          else if (!(await this.handleValidationError())) {
-            inlineEditorHandler.close(isRight ? 1 : -1, 0, false);
-          }
+        e.stopPropagation();
+        e.preventDefault();
+        if (inlineEditorFormula.wantsCellRef()) {
+          inlineEditorHandler.cursorIsMoving = true;
+          inlineEditorFormula.addInsertingCells(column);
+          keyboardPosition(e);
+        }
+        // if we're not moving and the formula is valid, close the editor
+        else if (!(await this.handleValidationError())) {
+          inlineEditorHandler.close(isRight ? 1 : -1, 0, false);
         }
       }
     } else {
--- a/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorKeyboard.ts
+++ b/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorKeyboard.ts
@@ -1,6 +1,7 @@
 //! This handles the keyboard events for the inline editor. In particular, it
 //! handles when the cursorIsMoving outside of the inline formula edit box.
 
+import { events } from '@/app/events/events';
 import { sheets } from '@/app/grid/controller/Sheets';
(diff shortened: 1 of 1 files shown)


davidfig (approved):

davidkircos:
Nice! This solves the issue in the attached PR.

If it's simple we should also make left arrow work for any positon where you can reference a cell. For example if you do 

=[Left arror] + [Left arrow]
the second left arrow should also reference the cell to the left, not move the cursor within the formula 
https://github.com/user-attachments/assets/e9fb287f-0ef7-423b-9e47-8f117515c398

AyushAgrawal-A2:
>If it's simple we should also make left arrow work for any positon where you can reference a cell. For example if you do
>
>=[Left arrorw] + [Left arrow]
>the second left arrow should also reference the cell to the left, not move the cursor within the formula

@davidkircos 
this is currently disable for both left and right arrow, current code restricts the behaviour to
- Left arrow adds cell reference if pressed when cursor is next to `=` (added in this PR)
- Right arrow adds cell reference if pressed when cursor is at last position

we should enable both or none for inserting cell references when cursor in middle of text. there is existing logic to decide whether a cursor position wants a cell ref or not. should be easy to just enable it, but there must be a reason why this was disabled in the first place.

davidkircos:
> > If it's simple we should also make left arrow work for any positon where you can reference a cell. For example if you do
> > =[Left arrorw] + [Left arrow]
> > the second left arrow should also reference the cell to the left, not move the cursor within the formula
> 
> @davidkircos this is currently disable for both left and right arrow, current code restricts the behaviour to
> 
> * Left arrow adds cell reference if pressed when cursor is next to `=` (added in this PR)
> * Right arrow adds cell reference if pressed when cursor is at last position
> 
> we should enable both or none for inserting cell references when cursor in middle of text. there is existing logic to decide whether a cursor position wants a cell ref or not. should be easy to just enable it, but there must be a reason why this was disabled in the first place.

I am not sure the reason. We want to match the behavior of other spreadsheets. And they allow any arrow key to select cells at any valid position for a cell reference in the Formula.

HactarCE:
> We want to match the behavior of other spreadsheets. Other sheets allow any arrow keys to select cells at any valid position for a cell reference in the Formula.

Other spreadsheets also disable arrow key cell selection after clicking in the inline editor. Personally I find this hidden state very confusing (and inconvenient, since I like navigating text using keyboard) but we'll need to track it if we want consistency.

davidkircos (changes requested):
Implementing more of the Excel behavior makes sense in this PR. This behavior is really important for users with Formulas. 
specifically:
- All arrow keys can be used to select cells in any valid cell reference position in a formula.

AyushAgrawal-A2:
>If it's simple we should also make left arrow work for any positon where you can reference a cell. For example if you do
>
>=[Left arrorw] + [Left arrow]
>the second left arrow should also reference the cell to the left, not move the cursor within the formula

@davidkircos 
done

davidkircos (approved):
This seems to work great! Big improvement!

HactarCE (changes requested):
It is now impossible to edit a formula using only the keyboard unless you type the whole thing all at once. I find this inconvenient personally, but I think it's also an accessibility issue.

Left arrow next to the initial `=` is ok because there's no reason to have your cursor to the left of `=`, but now it's basically impossible to move the cursor within the inline formula editor.

Anecdotally, it feels very inconsistent: occasionally left arrow moves within the formula, other times it inserts a cell reference, and usually it just closes the formula editor entirely.

davidkircos:
> It is now effectively impossible to edit a formula using only the keyboard without retyping the whole thing. I find this inconvenient personally, but I think it's also an accessibility issue.
> 
> Left arrow next to the initial `=` is ok because there's no reason to have your cursor to the left of `=`, but now it's basically impossible to move the cursor within the inline formula editor.
> 
> Anecdotally, it feels very inconsistent: occasionally left arrow moves within the formula, other times it inserts a cell reference, and usually it just closes the formula editor entirely. I prefer requiring up/down arrows to insert cell references, because those are not used for single-line text navigation anyway so there's no conflict.

How does our behavior differ from other spreadsheets? I didn't have any issues editing formulas with just the keyboard.

HactarCE:
Excel has the behavior of this PR (or close to it) when initially creating a formula (by double-clicking a blank or non-formula cell and typing `=`, or by selecting any cell and pressing <kbd>=</kbd>). There are three actions that break out of this mode:
- Clicking once in the inline editor while it is already open
- Double-clicking on an existing formula cell to edit it
- Pressing <kbd>F2</kbd> to edit a cell (whether it is blank or contains a formula)
After any of those three actions, arrow keys navigate within the inline editor and it becomes impossible to use the arrow keys to insert a cell reference.

Google Sheets is even more confusing; it is similar to Excel, but it's sometimes possible to re-enter the "arrow keys to insert cell references" state. If I type `=B1+C1` into a cell, then close and reopen the formula editor and press backspace twice to delete `C1`, then _whether I can use arrow keys to insert a cell reference depends on the timing between my two backspace keypresses_.

IMO both of these behaviors are [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) (hidden state! and why is <kbd>F2</kbd> different from double-clicking?) and inconvenient (I can't insert cell references when editing an existing formula?) Personally, I like Quadratic's current behavior because it is always possible to navigate within the inline editor (using left/right) and insert cells (using up/down), although inserting a reference to a left/right cell requires the counterintuitive key sequence <kbd>↑</kbd> <kbd>↓</kbd> <kbd>←</kbd>. I think replicating Excel's behavior (or any hybrid between what Excel does and what we currently do) would also be acceptable.

PR #1791, opened 2024-08-24
Allow getCells to populate Connections

davidfig:
- [x] replace {{x, y, sheetName?}}  in connection query with the value at that location
- [x] show an error if the sheet name does not exist 
- [x] update cells_accessed with all get cell calls 
- [x] allow inserting cells with a button in the connection editor
- [x] support relative references
- [x] show cursor when code editor loses focus (not perfect yet)
- [x] cell accessed highlighting


luke-quadratic:
This is rly awesome - no bugs found running through basic tests of feature

quadratic-client/src/app/ui/menus/CodeEditor/CodeEditorBody.tsx
```diff
@@ -143,6 +143,43 @@ export const CodeEditorBody = (props: Props) => {
       monaco.editor.defineTheme('quadratic', QuadraticEditorTheme);
       monaco.editor.setTheme('quadratic');
 
+      // this adds a cursor when the editor is not focused (useful when using insertCellRef button)
```
ddimaria, line 146: I feel like we've been putting this type of code in hooks files to reduce clutter here.

quadratic-client/src/app/ui/menus/CodeEditor/CodeEditorRefButton.tsx
```diff
@@ -23,19 +24,24 @@ export const CodeEditorRefButton = () => {
   const [disabled, setDisabled] = useState(true);
   useEffect(() => {
     const checkDisabled = () => {
-      // we do not yet support multiple multiCursors for inserting cell references
-      if (
-        (sheets.sheet.cursor.multiCursor && sheets.sheet.cursor.multiCursor.length > 1) ||
-        sheets.sheet.cursor.columnRow !== undefined
-      ) {
-        setDisabled(true);
+      // for connections, we only support one cursor position
```
ddimaria, line 27: 🙌

quadratic-core/src/controller/execution/run_code/run_connection.rs
```diff
@@ -1,10 +1,85 @@
...
 impl GridController {
+    /// Attempts to replace handlebars with the actual value from the grid
+    fn replace_handlebars(
+        &self,
+        transaction: &mut PendingTransaction,
+        sheet_pos: SheetPos,
+        code: &str,
+        default_sheet_id: SheetId,
+    ) -> Result<String, String> {
+        let re = Regex::new(
```
ddimaria, line 19: You can put this regex in a lazy_static (or lazy_cell) to avoid recompiling every time.

quadratic-core/src/controller/execution/run_code/run_connection.rs
```diff
@@ -1,10 +1,85 @@
...
+            r#"\{\{\s*(?:(relative)\s*:\s*)?(-?\d+)\s*,\s*(-?\d+)\s*(?:,\s*"?([^"]*)"?\s*)?\}\}"#,
+        )
+        .map_err(|err| format!("Regex compilation failed: {}", err))?;
+        let mut result = String::new();
+        let mut last_match_end = 0;
+
+        for cap in re.captures_iter(code) {
+            let whole_match = cap.get(0).unwrap();
```
ddimaria, line 27: Unsafe unwrap.

quadratic-core/src/controller/execution/run_code/run_connection.rs
```diff
@@ -15,15 +90,31 @@ impl GridController {
...
+                        sheet_pos.sheet_id.to_string(),
+                        replaced_code,
+                        kind,
+                        id.to_owned(),
+                    );
+                }
+                Err(msg) => {
+                    let error = RunError {
+                        span: None,
+                        msg: RunErrorMsg::PythonError(msg.to_owned().into()),
```
ddimaria, line 108: Should this be a `PythonError`?  It seems like we should have one for a connection here.

quadratic-core/src/controller/execution/run_code/run_connection.rs
```diff
@@ -15,15 +90,31 @@ impl GridController {
...
+                    };
+                    transaction.current_sheet_pos = Some(sheet_pos);
+                    let _ = self.code_cell_sheet_error(transaction, &error);
```
ddimaria, line 111: I think you have the `_` here to quiet the linter, but this effectively throws away errors from the result.

quadratic-core/src/controller/execution/run_code/run_connection.rs
```diff
@@ -32,3 +123,193 @@ impl GridController {
         self.transactions.add_async_transaction(transaction);
     }
 }
+
+#[cfg(test)]
+mod tests {
```
ddimaria, line 128: Nice tests 💯

quadratic-core/src/position.rs
```diff
@@ -371,6 +371,14 @@ pub struct SheetRect {
 }
 
 impl SheetRect {
+    pub fn new(x0: i64, y0: i64, x1: i64, y1: i64, sheet_id: SheetId) -> SheetRect {
```
ddimaria, line 374: Nice addition, I bet there's several places where this could be applied to reduce boilerplate.

quadratic-client/src/app/ui/menus/CodeEditor/CodeEditorBody.tsx
davidfig: this entire file needs a refactor. not a great way of separating it out without the refactor, but i agree with your comment.

quadratic-core/src/controller/execution/run_code/run_connection.rs
davidfig: We're going to generalize this to CodeError down the road.

quadratic-core/src/controller/execution/run_code/run_connection.rs
davidfig: i'm okay with that. not sure what else to do with it

quadratic-core/src/position.rs
davidfig: i've been adding this type of new statement to all our primitives. much easier to use; especially in tests

davidkircos:
@davidfig is this safe from SQL injection?

davidfig:
No, not yet. It's a much worse experience b/c the library requires that we know the types. 

For our current implementation and permissions, any user who has access to the SQL doesn't need injection since they already have full ability to change the SQL. In the future, when we allow users to run pre-existing SQL but not change the SQL (but they can change values that feed into the SQL), then we will need to be much better about injection protections.

davidkircos (approved):
@davidfig looks great to me! Fix merge conflicts and it's good to go

davidkircos:
Blocked by waiting for switch to A1

PR #1792, opened 2024-08-24
fix: long loading time due to multiplayer connection

### LINKED ISSUE #1730: nit: don't wait for python to load before opening a file

AyushAgrawal-A2:
closes #1730


davidkircos:
What does this change do?

I still notice this issue. If I:
1. Open a file
2. Reload the file (it's fast)
3. Make a small change and reload the file right after (it's significantly slower)

AyushAgrawal-A2:
@davidkircos 
Multiplayer connection has a timeout of 5 secs, if it connects it will sync otherwise it retries again...
previously it would attempt multiple retries, now files opens after first timeout, synced or not
Snackbar will show unsynced status on file open.

>Make a small change and reload the file right after (it's significantly slower)

this is due to 
 - 5sec timeout for first attempted multiplayer connection
 - the multiplayer is syncing, (can be checked in networks tab in dev console). This will take time depending on the no. / size of transactions and connection speed.

Root cause is multiplayer connection.

This PR removes the multiple attempted retries of MP connection by setting both loading states to false when first attempt fails.

davidfig (approved):

davidkircos:
> * 5sec timeout for first attempted multiplayer connection
> * the multiplayer is syncing, (can be checked in networks tab in dev console). This will take time depending on the no. / size of transactions and connection speed.

I am curious why the first attempt fails, it doesn't seem like it should fail. The service is up and the file is valid. 5 second retry is pretty slow. Maybe we should implement a retry backoff? 1s, 3s, 5s, etc...

Syncing a single small transaction is almost instant, it seems like it's something else.

luke-quadratic:
Approved on functionality - solves issue as described by drastically speeding up time to initially load any file

luke-quadratic (approved):

davidkircos (changes requested):
I belive this issue still exists:

To reproduce:

Open a file
Reload the file (it's fast)
Reload the file (it's fast)
Make a small change and reload the file right after (it's very slow)

luke-quadratic:
True, but it is still a big improvement to some cases and since it's a one line merge is worth sending - we can leave the issue open and describe what cases are still slow? @davidkircos

davidkircos (approved):
This is much faster!

davidfig (approved):

PR #1793, opened 2024-08-25
fix: early keystrokes are dropped when editing a cell

### LINKED ISSUE #1706: Early keystrokes are dropped when editing a cell
Steps to reproduce:
1. Select a cell (but do not press enter or double-click to edit the contents).
2. Press several keyboard keys at the same time. It's best if they're in the same row of the keyboard (e.g., `JKL;`), to avoid ghosting.

Expected behavior: All inputted characters appear in the cell.

Actual behavior: Some characters may be dropped.

Here I press `NEIO` (Colemak equivalent of QWERTY's `JKL;`) at the same time. Keys are dropped unless I double-click the cell first.

https://github.com/user-attachments/assets/ce6d7d30-9b9f-43ce-a9ef-70b5deafca39

AyushAgrawal-A2:
closes #1706

### DIFF
--- a/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorHandler.ts
+++ b/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorHandler.ts
@@ -32,6 +32,8 @@ class InlineEditorHandler {
   private open = false;
   private showing = false;
 
+  private initialValue = '';
+
   x = 0;
   y = 0;
   width = 0;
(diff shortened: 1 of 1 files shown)


davidfig (approved):

HactarCE:
This doesn't fix the issue for me

luke-quadratic:
This seems to fix the issue for me @HactarCE - only finding it doesn't if I mash like 5 or more characters at once which isn't something I presume anyone needs/wants to do, for 3 characters or less seems to always register.

HactarCE (approved):

PR #1795, opened 2024-08-25
Remove negative space + UI improvements

davidfig:
This PR: 
* removes negative space: we are disabling it, but keeping it as part of the grid for now. This allows people to move their content away from the negative space but not add new content there (or possibly we allow it for margin work?)
* add a sheet size to the grid to limit the parts of the sheet that are visually usable. The sheet size will grow as content is added.

Decisions:
* what do we do with our coordinate system? do we revert to Excel or keep the (x, y). At the minimum, we should maintain the relative (x,y). We also have already expanded it, eg, (col=5). We could change that to (col=A)

- [x] gridlines
- [x] headings
- [x] remove axis lines
- [ ] setSheetSize (limiting the number of columns and rows)
- [ ] automatically expand sheet size when set and less than the content
- [ ] decide what to do with cell references (revert to Excel or keep (x,y))
- [x] background fill w/grid lines
- [ ] viewport snapping
- [ ] remove mouse and keyboard interaction (except drop) when in negative space
   - [ ] mouse movement
   - [ ] clicking on cell
   - [ ] autocomplete
   - [ ] keyboard movement
- [ ] handle dropping and moving into negative space (dialog warning that this will offset the entire sheet)
- [ ] bounce when too far into negative space (except drop)
- [ ] ensure x,y are always positive


davidfig:
Moved to remove-negative-2

PR #1796, opened 2024-08-26
feat: team file list avatars (& misc. improvements)

jimniels:
### Fixes 

- For grid view, avatars are moved to far right with the secondary actions triggered moved to top right
  - Everything stays the same on list view (avatar/secondary actions on the far right)
- Avatars always show on team list view, even if they are your files.
- Dates older than 24hrs switch from relative ("2h ago") to showing the date (Aug 24, 2024)
- Fixes a bug where tooltips could display behind other content
- Changes sort order label from "Last updated" to "Last modified" to match the file's meta which says "Modified ___ ago"


davidkircos (approved):
@jimniels Awesome! Looks very clean.

PR #1799, opened 2024-08-26
fix: numbers not getting rendered after resizing other columns

### LINKED ISSUE #1797: Numbers not getting rendered after resizing other columns
CleanShot 2024-08-26 at 10 05 59@2x

(Msg in slack for sheet to replicate)

AyushAgrawal-A2:
closes #1797


luke-quadratic:
Blocking 
CleanShot 2024-08-26 at 12 19 50

luke-quadratic (approved):

quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellLabel.ts
```diff
@@ -222,10 +222,10 @@ export class CellLabel {
     if (!this.isNumber()) return false;
 
     const clipLeft = Math.max(this.cellClipLeft ?? -Infinity, this.AABB.right - (this.nextLeftWidth ?? Infinity));
-    if (this.actualLeft < clipLeft) return true;
+    if (clipLeft - this.actualLeft > 0.001) return true;
 
     const clipRight = Math.min(this.cellClipRight ?? Infinity, this.AABB.left + (this.nextRightWidth ?? Infinity));
-    if (this.actualRight > clipRight) return true;
+    if (this.actualRight - clipRight > 0.001) return true;
```
davidkircos, line 228: What does the magic number `0.001` do @AyushAgrawal-A2 ?

AyushAgrawal-A2: @davidkircos 
Both the values being compared here are suppose to have the same value, It differs slightly due to floating point handling in javascript. It results in false positive because of the least significant digit is different.
0.001 is the minimum difference added here to counter this and place a minimum threshold difference. - `epsilon comparison`

davidkircos: Should this depend on the precision the user requests as display or is it just based on the precision of floats in JS?

Is there no built in way to compare two floats in JS this way?

AyushAgrawal-A2: @davidkircos 

>based on the precision of floats in JS

this, no relation to `the precision the user requests as display`

>Is there no built in way to compare two floats in JS this way?

no

davidfig: We should make a global precision float comparison function. This is used in a few places.

davidfig (approved):

AyushAgrawal-A2:
>We should make a global precision float comparison function. This is used in a few places.

@davidfig 
done

davidfig (approved):

PR #1801, opened 2024-08-26
fix: add Shift+/ shortcut for show code cell menu

AyushAgrawal-A2:


davidkircos (approved):
I am not sure how to test that this fixes the French keyboard issue. But this looks good to me.

PR #1811, opened 2024-08-27
Data validation fixes

### LINKED ISSUE #1798: Data validation improvements
- [ ] Scrollable dropdowns 
- [ ] Message bleeding

Example in image where number of options goes past the limit of the sheet, should limit size of shown options and make it scrollable 
CleanShot 2024-08-26 at 10 10 08@2x

Messages bleed over and don't wrap
CleanShot 2024-08-26 at 10 22 38@2x

davidfig:
- [x] clamps the size of a dropdown list to the smaller of 50vh and the bottom of the viewport when opened
- [x] ensures that all text is wrapped within messages 

Closes #1798


luke-quadratic (approved):
Works exactly as issue describes

ddimaria (approved):
Needs UI review.

PR #1814, opened 2024-08-28
fix: dashboard stuck in loading state on pressing back from app

AyushAgrawal-A2:
Fixes:
- Dashboard route stuck on loading, when back button is pressed during navigation to App (/files) route
- New file dialog not closing - same criteria as above
- Console error from new Avatar component due to missing forwardref
- Fix mobile scroll


davidfig (approved):

davidkircos (changes requested):
I was still able to get stuck in loading state

1. Open blank file
2. Click browser Back button
3.  Stuck loading
CleanShot 2024-08-28 at 16 05 50@2x

AyushAgrawal-A2:
@davidkircos 

Now, I am not able to get it stuck on loading state, pls verify

p.s.: I know it is important, that is why I took it on priority, suo moto.

luke-quadratic (approved):
Am personally no longer getting into this state but @davidkircos should double-confirm since he's the one that experienced it so many times

davidkircos (changes requested):
On the inital signup flow only, this problem still exists. This is probably due to the redirect when a user is first created.

While logged out
1. Go to app
2. Create a fresh account
3. Create a blank file
4. Click back
Takes you to Auth0 instead of dashboard.


https://github.com/user-attachments/assets/123f76e0-d3e7-4847-9a9d-b631214a8809

davidkircos (approved):
Fixed!

PR #1819, opened 2024-08-29
feat: when users have 0 teams, make them create one

jimniels:
- [ ] Ask company name
- [ ] Can you code or not?

When users try to access `/` but they have 0 teams, this navigates them to a special page (`team-onboarding`) where they can create a new team by giving it a name.

Generally this will be hit by new users to the app, but would also apply to users who are part of a team and leave it (then have 0 teams).

CleanShot 2024-08-29 at 11 48 13@2x

It uses the same form component we use when the user clicks "new team" in the dashboard.

CleanShot 2024-08-29 at 11 51 40@2x

### DIFF
--- a/quadratic-client/src/routes/teams.create.tsx
+++ b/quadratic-client/src/routes/teams.create.tsx
@@ -20,40 +20,28 @@ import { useEffect, useState } from 'react';
 import { useForm } from 'react-hook-form';
 import { ActionFunctionArgs, redirect, useFetcher, useLocation, useNavigate, useParams } from 'react-router-dom';
 import z from 'zod';
-type ActionData = {
-  name: string;
-  picture?: string;
-};
+
+const CreateTeamFormSchema = z.object({
+  name: TeamSchema.shape.name,
+});
 
 export const action = async ({ request }: ActionFunctionArgs) => {
-  const data: ActionData = await request.json();
+  const data = await request.json();
 
-  const { uuid } = await apiClient.teams.create(data);
-  if (uuid) {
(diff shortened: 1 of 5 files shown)


quadratic-client/src/routes/teams.create.tsx
```diff
@@ -66,57 +54,72 @@ export const Component = () => {
...
+  ) : null;
+};
+
+export const CreateTeamForm = ({ onCancel }: { onCancel?: () => void }) => {
+  const { addGlobalSnackbar } = useGlobalSnackbar();
+  const fetcher = useFetcher();
+  const form = useForm<z.infer<typeof CreateTeamFormSchema>>({
+    resolver: zodResolver(CreateTeamFormSchema),
+    defaultValues: {
+      name: '',
```
ddimaria, line 78: Is this necessary?

quadratic-client/src/routes/teams.create.tsx
```diff
@@ -66,57 +54,72 @@ export const Component = () => {
...
   useEffect(() => {
     if (fetcher.state === 'idle' && fetcher.data && fetcher.data.ok === false) {
       addGlobalSnackbar('Failed to create team. Try again', { severity: 'error' });
     }
   }, [fetcher, addGlobalSnackbar]);
 
-  const onSubmit = async (data: z.infer<typeof FormSchema>) => {
-    fetcher.submit(data, { method: 'POST', encType: 'application/json' });
+  const onSubmit = async (data: z.infer<typeof CreateTeamFormSchema>) => {
+    fetcher.submit(data, { action: '/teams/create', method: 'POST', encType: 'application/json' });
```
ddimaria, line 90: You may be able to get the `action` from routes file?

PR #1835, opened 2024-09-04
fix: second file creation will result in indefinite loading

AyushAgrawal-A2:
fixes
[Second file creation will result in indefinite loading](https://www.loom.com/share/949d96f8f7944aea937d47cabb77af3c)
[Command Palette New option doesn't create a new file with "Untitled"](https://www.loom.com/share/4a4475442a764abd89e5924005b9aabf)
[Duplicate file doesn't not produce exact copy and it is without (COPY) in the title name](https://www.loom.com/share/cc99ba76511c490d99e7216e893bdf55)

### DIFF
--- a/amplify.yml
+++ b/amplify.yml
@@ -15,3 +15,5 @@ frontend:
           value: "same-origin"
         - key: "Cross-Origin-Embedder-Policy"
           value: "require-corp"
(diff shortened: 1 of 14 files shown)


davidkircos (approved):
lgtm!

PR #1845, opened 2024-09-05
chore: actions refactor

AyushAgrawal-A2:
- [x] Move keyboard shortcuts to TS from Zod
- [x] Get keyboard shortcut display from default shortcuts
- [x] Zoom menu


quadratic-client/src/app/actions/actions.ts
```diff
@@ -0,0 +1,67 @@
+export enum Action {
```
davidfig, line 1: Do we still need the string version of this? Can't we refer to it directly as the enum in the code?

davidfig: There's a way to turn TS enums into strings, I think. We'll need something like that for user-defined keyboard shortcuts.

quadratic-client/src/app/gridGL/interaction/keyboard/keyboardPosition.ts
```diff
@@ -323,110 +324,110 @@ function moveCursor(deltaX: number, deltaY: number) {
 
 export function keyboardPosition(event: KeyboardEvent): boolean {
   // Move cursor up
-  if (matchShortcut('move_cursor_up', event)) {
+  if (matchShortcut(Action.MoveCursorUp, event)) {
```
davidfig, line 327: Just an observation: I don't understand why we defined these as user-definable. They should just be constant. But nothing to do here :)

davidfig (approved):

quadratic-client/src/app/actions/actions.ts
AyushAgrawal-A2: Enum to string defaults to 0, 1, 2, ... , I am just keying in snake case string values, I don't want to add a dependency for converting it to string.
Having string values will make conversion to json simple, i think, as I am using this as key in default KeyboardShortcuts object

jimniels (approved):
This looks good to me 👍

PR #1855, opened 2024-09-07
feat: add claude-3.5-sonnet to ai assistant in code editor

AyushAgrawal-A2:
requires adding anthropic api key to deployment, in quadratic-api env variables
```
ANTHROPIC_API_KEY=
```


davidkircos:
Added to Heroku staging, and production.

It's worth noting that going forward all external APIs in every service should be optional for self hosting. If a user self-hosts and doesn't include an ANTHROPIC_API_KEY, it should be disabled in the app. Same with OPEN_AI and every other external service

davidkircos:
Quick inital feedback.

The first time I used Claude in Python, it gave me Javascript code.

The margin between Avatar and the text is different.
CleanShot 2024-09-07 at 09 09 36@2x

quadratic-api/src/routes/ai/openai.ts
```diff
@@ -0,0 +1,75 @@
...
+      messages,
+      temperature: 0,
+      stream: true,
+    });
+
+    for await (const chunk of completion) {
+      response.write(`data: ${JSON.stringify(chunk)}\n\n`);
+    }
+
+    response.write('[DONE]\n\n');
```
davidkircos, line 62: Why write this to the stream?

quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
```diff
@@ -65,8 +59,40 @@ If the code was recently run here is the result:
 ${JSON.stringify(consoleOutput)}\`\`\`
 This is the documentation for Quadratic: 
 ${QuadraticDocs}`,
-    },
-  ] as AiMessage[];
+    [consoleOutput, editorContent, schemaJsonForAi, selectedCell.x, selectedCell.y, mode]
+  );
+
+  const cellContext = useMemo<AIMessage>(
```
davidkircos, line 65: Is this sent with every message?

Should we also add the editor content in addition to the console output?

quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
```diff
@@ -249,6 +235,20 @@ ${QuadraticDocs}`,
           <div id="ai-streaming-output-anchor" key="ai-streaming-output-anchor" />
         </div>
       </div>
+
+      <div className="flex flex-col items-end gap-2 px-3 pb-2">
+        <select
```
davidkircos, line 240: use shadcn dropdown https://ui.shadcn.com/docs/components/dropdown-menu

quadratic-client/src/app/ui/menus/CodeEditor/AiAssistant.tsx
AyushAgrawal-A2: ai chat api is stateless, entire prompt history, include even quadratic docs in sent in every api request
editor content is already being sent here

quadratic-api/src/routes/ai/openai.ts
AyushAgrawal-A2: works without it, I think it was part of some old api
https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream

waveminded:
Add a tooltip to the button to change AI models? All other buttons in the app have one.

davidkircos:
> Add a tooltip to the button to change AI models? All other buttons in the app have one.

In other tools it usually shows the model name, you click on it and you can change it. I think that would be better than the logo. More self explanatory 
CleanShot 2024-09-09 at 11 07 19

davidkircos (approved):
This looks good to me!

davidfig (approved):

PR #1866, opened 2024-09-13
Add snowflake connection

ddimaria:
Closes https://github.com/quadratichq/quadratic/issues/1804

- [x] Connect to Snowflake
- [x] Query Snowflake
- [x] Return the database schema from Snowflake
- [x] Basic testing
- [x] Test mocks


davidkircos:
Missing logo in the cell type menu 
CleanShot 2024-09-18 at 11 49 26@2x

ddimaria:
> Missing logo in the cell type menu CleanShot 2024-09-18 at 11 49 26@2x

@davidkircos this is fixed

Screenshot 2024-09-18 at 7 26 47 PM

jimniels:
The logos for Microsoft sql and snowflake are a little big in relationship to the others

CleanShot 2024-09-23 at 16 10 35@2x

Something more like this would make sense:

CleanShot 2024-09-23 at 16 10 24@2x

I probably need to resize these in the source Figma file and then export back in here, as this is the kind of thing that's just a designer's judgement. Not every logo can fill the canvas size for these buttons, as they're all different visual weights.

quadratic-rust-shared/src/sql/snowflake_connection.rs
```diff
@@ -0,0 +1,496 @@
...
+            "TEST".into(),
+            "TEST".into(),
+            None,
+            "ALL_NATIVE_DATA_TYPES".into(),
+            None,
+            None,
+        )
+    }
+
+    // async fn _seed(
```
davidfig, line 277: ?

davidfig (approved):

quadratic-rust-shared/src/sql/snowflake_connection.rs
ddimaria: I originally was trying to run snowflake in docker via LocalStack, but they required a special license, so just created this DB in snowflake and used for testing.  I don't know where else to put this query so that it can be recreated if needed?

PR #1871, opened 2024-09-16
Implement `Duration` type and arithmetic ops for dates/times/datetimes/durations in formulas

HactarCE:
Durations can be written as `1 year, 6 months, 4 weeks, 3 days, 12 hours, 30 minutes, 5 seconds, 16 milliseconds, 666 microseconds, 667 nanoseconds` or the compact form `1y6mo4w3d12h30m5s16ms666us667ns` or anything in between. In a formula, they can be written using a string and will be coerced when used with common operators. For example, `"8 months" * 2` results in `1y 4mo`.

All basic arithmetic operations (`+`, `-`, `*`, `/`) should work in the most reasonable way with dates, times, durations. When an operation is not allowed, a good error is given, including a suggestion to use a duration type if the user seems to be trying to do something that should use one.

Known issues/limitations that can be addressed in the future:
- [ ] No formatting customization (e.g., `1y` vs. `1 year` or `1 week` vs. `7 days`)
- [ ] `SUM` and `AVERAGE` do not work with the new types

Tasks remaining:
- [x] Seconds are printed with too much precision, exposing floating-point error
- [x] Parse durations when typed into the sheet
- [x] Convert to Python/pandas


quadratic-core/src/formulas/functions/operators.rs
```diff
@@ -108,6 +182,7 @@ mod tests {
 
         assert_eq!(
             (1 * -6 + -2 - 1 * (-3_i32).pow(2_u32.pow(3))).to_string(),
+            // TODO: `-3 ^ 2` should be parsed as `-(3 ^ 2)`, not `(-3) ^ 2`
```
ddimaria, line 185: Should we create an issue for this TODO?

quadratic-core/src/values/time.rs
```diff
@@ -67,42 +74,170 @@ impl Ord for Instant {
...
+
+        let mut is_first = true;
+        for (mut quantity, unit) in units_to_display {
+            match is_first {
+                true => is_first = false,
+                false => write!(f, " ")?,
+            }
+
+            // Replace `-0.0` with `0.0`.
+            if quantity == 0.0 {
```
ddimaria, line 154: TIL: `0.0 == -0.0` in Rust.

quadratic-core/src/values/time.rs
```diff
@@ -115,17 +250,236 @@ impl Ord for Duration {
...
         }
         ret
     }
 }
 
+impl Add for Duration {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        // TODO: this should always be checked addition
```
ddimaria, line 265: Create an issue for these checked values?

quadratic-core/src/values/time.rs
```diff
@@ -115,17 +250,236 @@ impl Ord for Duration {
...
+    }
+
+    /// Returns whether the duration represents zero time.
+    fn is_zero(self) -> bool {
+        self == Self::ZERO
+    }
+
+    /// Returns whether the duration represents an integer number of days.
+    pub fn is_integer_days(self) -> bool {
+        (self.seconds / Self::DAY.seconds).fract() == 0.0
```
ddimaria, line 438: I think you may be doing upstream checks, but this could be a divide-by-zero error.

quadratic-core/src/values/time.rs
```diff
@@ -134,3 +488,177 @@ pub fn map_local_result<T: chrono::TimeZone + Display>(
...
+            ("1y ,3m", Err(())),                     // space before comma
+            ("4:30", Err(())),                       // time, not duration
+            ("16", Err(())),                         // no unit
+            ("3,y10d", Err(())),                     // comma in a bad spot
+            ("10year,3 mo2w16d,, 4000min", Err(())), // double comma
+            ("1 year month", Err(())),               // two units in a row
+            ("0.5 year", Err(())),                   // fractional year is not allowed
+            ("3 mo, 10year", Err(())),               // nondecreasing unit order
+        ] {
+            println!("Parsing duration {input:?}");
```
ddimaria, line 633: Remove this?

quadratic-core/src/values/time.rs
```diff
@@ -134,3 +488,177 @@ pub fn map_local_result<T: chrono::TimeZone + Display>(
...
+            let result = input.parse::<Duration>();
+            match expected {
+                Ok(expected_ok) => {
+                    assert_eq!(result.unwrap().to_string(), expected_ok);
+
+                    // Test idempotency
+                    println!("Parsing duration {expected_ok:?}");
```
ddimaria, line 640: This could be removed as well.

quadratic-kernels/python-wasm/quadratic_py/utils.py
```diff
@@ -61,6 +62,29 @@ def normalize_bool(value: str) -> bool:
     return value
 
 
+def parse_duration(value: str) -> relativedelta:
+    ret = relativedelta()
+    for m in re.compile(r'([-\d\.]+)([a-zµ]+)').finditer(value.lower()):
+        count = float(m.group(1))
+        unit = m.group(2)
+        print(count,unit)
```
ddimaria, line 70: Remove?

ddimaria (approved):
Great idiomatic code ❤️    I left several non-blocking comments.

quadratic-core/src/formulas/functions/operators.rs
HactarCE: Good idea! [Done](https://github.com/quadratichq/quadratic/issues/1906)

quadratic-core/src/values/time.rs
HactarCE: IEEE 754 moment

quadratic-core/src/values/time.rs
HactarCE: [Done!](https://github.com/quadratichq/quadratic/issues/1907)

quadratic-core/src/values/time.rs
HactarCE: `Self::DAY` is a known nonzero constant. It's also not `-1` so overflow is impossible.

quadratic-core/src/values/time.rs
HactarCE: I like to leave `println!()`s in tests in case they break in the future, since the traceback alone isn't enough to figure out which input failed.

quadratic-kernels/python-wasm/quadratic_py/utils.py
HactarCE: Now that one's my bad lol

quadratic-core/src/values/time.rs
HactarCE: Same as above

PR #1882, opened 2024-09-18
fix: cannot go from multi-cell selection to single-cell selection

### LINKED ISSUE #1880: bug: cannot go from multi-cell selection to single-cell selection
If you start dragging and make a multi-cell selection, you can never drag back to a single cell selection. Dragging back to the cell you started with always keeps two cells selected. Here's quadratic:

https://github.com/user-attachments/assets/f97b57bb-b17d-45dc-ac7d-709d1246d330

Whereas I would expect it to work like Sheets (minus the buggy hand cursor part ha):

https://github.com/user-attachments/assets/2dd3ea04-ed0f-4d3b-a38e-328d84b1dfd2

AyushAgrawal-A2:
closes #1880


davidfig (approved):

jimniels (approved):
nice, I love one line changes!

PR #1864, opened 2024-09-12
feat: AI Analyst

AyushAgrawal-A2:
closes #1201 

Review
- [x] Infra / security code review @davidkircos 
- [x] API and Rust review @ddimaria 
- [x] Client review @jimniels 
- [x] Merge!

Feat
- [x] Only open AI Chat automatically on empty files @AyushAgrawal-A2 
- [x] AI Analyst can return data tables directly to the sheet
~~Select programming language~~
- [x] Update buttons for code that's already run

PR #1833, opened 2024-09-03
feat: app redesign

jimniels:
### Todos

- [x] Sidebar with actions
- [x] File menu with actions (needs to be wired up)
- [x] Formatting bar with actions (needs to be wired up)
- [x] Get rid of floating context menu and have a right-click context menu
- [x] Command palette actions/icons
  - Not all of these got wired up, we'll do this slowly over time
- [x] Permissions
- [x] Mixpanel events for new menus
  - [x] Formatting bar
  - [x] File menus
  - [x] Sidebar
- [x] Responsiveness
- [x] Fix all `TODO: (jimniels)` in PR code
- [x] Map `app/ui/icons` to `shared/components/icons`
  - `app/ui/icons` were marked as deprecated and we'll slowly replace these over time

### Nice to haves

- [x] User menu in upper right (sign out)
- [x] Swap out icons on dashboard to new ones

### Didn't Do

- Formatting button renders real-time update depending on the active formatting applied to the current cell

## Outstanding Bugs
- [x] Clicking on code editor in the side bar with a newly opened file doesn't properly set the location. And code is lost after running and closing.
- [x] User primary color for resize controls
- [x] Show sidebar on tablet
- [x] Kernel menu not opening
- [x] Hot-swapping code cells doesn't work
  - 
- [x] Code peak triggers when menus are open (@davidfig)
  - 
- [x] Avatars don't align in sheet for multiplayer situations ([img](https://github.com/user-attachments/assets/35a6c7fb-495c-4dbb-83b7-f76a10d81ba8))

PR #1884, opened 2024-09-19
fix: react re-renders

AyushAgrawal-A2:
has everything in https://github.com/quadratichq/quadratic/pull/1864 except AI Side Panel

based on https://github.com/quadratichq/quadratic/pull/1823 which is based on https://github.com/quadratichq/quadratic/pull/1833


jimniels (approved):
This is a great set of performance changes. TY @AyushAgrawal-A2

PR #1889, opened 2024-09-19
Rust cleanup: `cargo fmt`; fix clippy lints, and 2024 edition migrations

HactarCE:


quadratic-rust-shared/src/pubsub/redis.rs
```diff
@@ -97,8 +97,7 @@ impl super::PubSub for RedisConnection {
         value: &[u8],
         _active_channel: Option<&str>,
     ) -> Result<()> {
-        self.multiplex.publish(channel, value).await?;
```
ddimaria, line 100: I just did a `let () self.multiplex.publish(channel, value).await?;` in my branch rather than returning the result, which might be hard to understand when reading the code.

HactarCE: Oops yep, I meant to do that. Returning the result was the first thing I tried, and later I figured out `let () = `.

PR #1895, opened 2024-09-19
fix(safari): global keyboard shortcuts with modifiers don't work

### LINKED ISSUE #1894: bug (safari): global keyboard shortcuts with modifiers don't work
Global modifier key shortcuts don't seem to work, such as:

- Command palette `CMD + P`
- Go to `CMD + G`
- Presentation mode `CMD + .`

AyushAgrawal-A2:
closes #1894


luke-quadratic (approved):
fixes issue

jimniels (changes requested):
None of the global modifier shortcuts work for me.

Also, looking at the code, if we're gonna enumerate them we're gonna need a much bigger list. For example, in addition to the ones already listed, none of these work:

CleanShot 2024-09-19 at 21 03 26@2x

PR #1901, opened 2024-09-21
chore: cleanup clones for upgrade/import/export file for v1.7

AyushAgrawal-A2:


quadratic-core/src/grid/file/serialize/code_cell.rs
```diff
@@ -13,60 +13,63 @@ use super::{
...
                 current::OutputValueSchema::Array(current::OutputArraySchema { size, values }) => {
                     Value::Array(crate::Array::from(
                         values
                             .chunks(size.w as usize)
-                            .map(|row| row.iter().map(import_cell_value).collect::<Vec<_>>())
+                            .map(|row| {
+                                row.iter()
+                                    .map(|value| import_cell_value(value.to_owned()))
+                                    .collect::<Vec<_>>()
+                            })
```
AyushAgrawal-A2, line 40: not able to figure out how to get owned value here, row chuck is a slice
`value.to_owned()` can be avoided if we can directly use owned value

quadratic-core/src/grid/file/serialize/code_cell.rs
```diff
@@ -79,7 +82,9 @@ pub(crate) fn export_rows_code_runs(
                             },
                             values: array
                                 .rows()
-                                .flat_map(|row| row.iter().map(export_cell_value))
+                                .flat_map(|row| {
+                                    row.iter().map(|value| export_cell_value(value.to_owned()))
```
AyushAgrawal-A2, line 86: same

davidfig (approved):

quadratic-core/src/grid/file/serialize/code_cell.rs
HactarCE: I pushed this one

quadratic-core/src/grid/file/serialize/code_cell.rs
HactarCE: I don't think there's an easy way to avoid this one. It's possible by collecting each row into a `Vec<()>` before transforming it, or by just manually building up a `Vec<Vec<()>>` element-by-element in this function. The latter is the most clone-free way but it's not pretty.

PR #1922, opened 2024-09-25
A1: Remove negative coordinates and change cell references to A1

davidfig:
TODOs: https://app.quadratichq.com/file/c8aab96e-f26c-4c3b-afdc-66be0a77a0b3

### Doc updates
- [ ] remove all negative and (0, 0) references from docs
- [ ] add page that describes why we moved away from (x, y) and negative offsets (include link in snackbar message)
- [ ] language docs updates
- [ ] upgrade all examples to 1.7.1 and ensure all cell references work properly
- [ ] ensure all snippets use new positioning


davidkircos:
I think the can't zoom into negative space is a nicer experience than the can zoom into gray space but can't pan into gray space only out.

PR #1939, opened 2024-10-01
Fix Match formula

### LINKED ISSUE #1916: Match formula not working
Not sure cause, but replicating in other sheets returns results correctly. 

See sheet in Slack for recreation.

HactarCE:
Fixes #1916

### DIFF
--- a/quadratic-core/src/formulas/functions/lookup.rs
+++ b/quadratic-core/src/formulas/functions/lookup.rs
@@ -461,21 +461,24 @@ fn lookup<V: ToString + AsRef<CellValue>>(
         LookupMatchMode::Exact => std::cmp::Ordering::Equal,
         LookupMatchMode::NextSmaller => std::cmp::Ordering::Less,
         LookupMatchMode::NextLarger => std::cmp::Ordering::Greater,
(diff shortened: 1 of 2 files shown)


quadratic-core/src/formulas/functions/lookup.rs
```diff
@@ -461,21 +461,24 @@ fn lookup<V: ToString + AsRef<CellValue>>(
...
+                    LookupSearchMode::LinearReverse => {
+                        lookup_regex(regex, haystack.iter().rev()).map(fix_rev_index)
+                    }
+                    LookupSearchMode::BinaryAscending | LookupSearchMode::BinaryDescending => {
+                        internal_error!(
+                            "invalid match_mode+search_mode combination \
+                             should have been caught earlier in XLOOKUP",
+                        );
+                    }
+                });
```
HactarCE, line 478: This block is unchanged, just indented

ddimaria (approved):

HactarCE:
The lines lacking coverage are internal errors that should be unreachable. It's just a bugfix + regression test, so no user test required. Merging now.

PR #1945, opened 2024-10-03
fix: focus

AyushAgrawal-A2:


quadratic-client/src/app/ui/menus/TopBar/TopBarMenus/EditMenubarMenu.tsx
```diff
@@ -1,12 +1,19 @@
...
     <MenubarMenu>
       <MenubarTrigger>Edit</MenubarTrigger>
-      <MenubarContent className="pointer-move-ignore">
+      <MenubarContent
+        className="pointer-move-ignore"
+        onCloseAutoFocus={(e) => {
+          e.preventDefault();
+          focusGrid();
+        }}
+      >
```
jimniels, line 16: Minor, take it or leave it: I don't know if we'll ever have another use for this menu, in which case you could just put this on the underlying `MenubarContent` shadcn component.

I did that with `onEscapeKeyDown` so I wouldn't have to put it on every use of the component in the file menu.

jimniels (approved):
On minor comment, but otherwise LGTM 👍

quadratic-client/src/app/ui/menus/TopBar/TopBarMenus/EditMenubarMenu.tsx
AyushAgrawal-A2: done

davidkircos (approved):

Issue #1938, opened 2024-10-01
Strange sheet tab button behavior 


luke-quadratic:
![CleanShot 2024-10-01 at 13 54 01](https://github.com/user-attachments/assets/fdc80447-5665-4372-a571-60512988d885)

AyushAgrawal-A2:
@luke-quadratic 
not able to reproduce, fixed I think, please confirm

luke-quadratic:
Can confirm, fixed

PR #1947, opened 2024-10-03
fix: sheet bar dropdown

jimniels:
Fixes the sheet bar dropdown to use shadcn components (instead of szh-menu), so it now supports dark mode and brings it inline with other UI elements. Also fixes it so after sheet actions run the grid receives focus.


jimniels:
This needs a fix for #1938 which causes really weird dragging issues. Would also be nice to fix it so choosing a color isn't part of the "dragging" interface for the dropdown.

https://github.com/user-attachments/assets/14319110-58b5-40cc-a615-f63f1167de8b

cc @davidfig or @AyushAgrawal-A2 could you take a look?

AyushAgrawal-A2:
>This needs a fix for #1938 which causes really weird dragging issues. Would also be nice to fix it so choosing a color isn't part of the "dragging" interface for the dropdown.

done

davidkircos (changes requested):
Rename button on the menu doesn't work. It loses focus immediately

Move left and move right also don't work

davidkircos (approved):
lgtm!

PR #1952, opened 2024-10-04
feat: new design + bug fixes for multiplayer avatars

### LINKED ISSUE #1951: bug: make additional document users accessible
If I want to follow someone in the sheet who is hidden behind the "+n" control, I can't click on that and see the list of users and choose someone.

image

jimniels:
Design enhancements to fix #1951 

- Users who can be followed get a colored dot
- Highlight for users who you follow
- Additional users dropdown when there are a lot of users in the document


jimniels:
Feedback from live session after standup:

- [ ] Anonymous user gets a dropdown for their avatar, even though they're not logged in
  - image
- [ ] Dropdown "following you" shouldn't break to two lines
  -  image
- Triple dot tooltip: turn it into "+1" number of avatars
- Other users menu disappears if someone you're following is moving around (or you move the grid)
- People who are following you: don't disable them, just indicate they're following you on hover (no extra treatment)
- Following icon or some extra indication for people who are following you
- Multiplayer colors
- Bottom border for following is missing

davidkircos (approved):
This looks awesome! Really solid improvement

PR #1964, opened 2024-10-11
fix: getting started buttons wrapping issues

jimniels:
Button text shouldn't wrap, but buttons inside container should (when screens get narrow)

image


davidkircos (approved):
Appears to work great!

Issue #1965, opened 2024-10-12
Trouble importing CSV files.


paulwrath1223:
On Ubuntu 24.04, and latest stable chrome, I am unable to drag CSV files in to the grid. (especially annoying when importing through the insert tab just tells you to drag and drop the file)

However, I can still select a CSV as the source when creating a new sheet and this works... Except it doesn't parse my CSV correctly. My CSV uses semicolons as the separator, and granted, that's pretty stupid when it's literally called "Comma Separated Values", but this is the standard for PCB component bill of materials, being exported that way in Fusion 360 and KiCad.

In any case, google sheets is able to parse it fine with automatic detection, and even lets the user specify the delimiter.

If it's not already implemented in the dev branch, I can take a crack at the automatic delimiter detection and put in a PR. My idea is to try to parse it using commas first, but if the rows have different lengths, try semicolon and tab (\t).

I could also look into drag and drop on Linux, but this is my first week of using Linux, so I don't think I would be much help.

davidkircos:
Which browser are using where drag and drop doesn't work?

AyushAgrawal-A2:
@paulwrath1223 
Hey, I am on Ubuntu 24.04 and I am able to drag/drop CSV files. For me it works on Google Chrome and Brave browser
Below is the screen cast showing a drag / drop of a 100k row sample CSV file. I use this daily as a test for performance regression for my PRs.

[Screencast from 2024-10-12 12-38-21.webm](https://github.com/user-attachments/assets/d1dd9f79-cab1-4697-9dcc-e1dea373b5f8)

Happy to debug and solve the issue you are facing, but I need more details on how to replicate it.

paulwrath1223:
Hey guys, Im actually having a more insidious problem now. Dragging and dropping is still hit-and-miss, and I don't know how to replicate the state that makes it fail, but see attached video.

[Bug video](https://youtu.be/6fjoxOd7-nQ) (too big for GitHub attachment)

I swear I'm not copying and pasting.

Chrome Version 129.0.6668.100 (Official Build) (64-bit)

paulwrath1223:
Can confirm that text typed in a password field in another tab is still picked up but converted to `••••••••`

AyushAgrawal-A2:
@paulwrath1223 
I'll need more help from you to debug this:
1. For drag/drop failures, when it stops working can you please check the console for errors and please share with us. It is possible that the app kernel panicked or there is some other error thrown there.
[...]

PR #1969, opened 2024-10-14
fix: prevent default paste on middle click in Linux

AyushAgrawal-A2:
ref https://github.com/quadratichq/quadratic/issues/1965


davidkircos (approved):
I am not able to test this directly but the code looks right

davidfig (approved):

PR #1979, opened 2024-10-17
feat: filter files by creator

jimniels:
Filter field allows inputting the file name or the creator's name and it will filter/highlight on that value

You can also click on the avatar of the person on the file and it will populate the filter field with that person's name for filtering.

Filter priority:

- File name
- User name
- User email


davidkircos (approved):
nice! looks great to me!

PR #1980, opened 2024-10-17
feat: show logged in user context

jimniels:
On the dashboard or in the app, there are times you can hit a route that results in a screen where something went wrong.

Today it's not always clear that you might be hitting that error based on who you're logged in as.

For example, I go to `/teams/team-id-1` and it says "You can't access this team" and I think "I should be able to, that's weird..."

But the reality is that I can't access that team because I'm not logged in as the user I think I am when I hit that screen.

So this PR makes it a lot easier to infer why you might be hitting these screens, as it will show you you're currently logged in status (if you're logged in — if you're not logged in, e.g. an anon user, it won't show this).

Examples:

**You don't have permission to see team**

CleanShot 2024-10-17 at 13 42 37@2x

**Team not found**

CleanShot 2024-10-17 at 13 42 29@2x

**Unexpected error**

CleanShot 2024-10-17 at 13 44 32@2x


davidkircos (approved):
Nice improvement!

PR #1984, opened 2024-10-21
feat: force logout via route search params

jimniels:
Allow visiting `/logout` in the URL bar to log you out of the app.


jimniels:
Gonna close this for now...

PR #1986, opened 2024-10-22
devops: use aws s3 transfer acceleration

AyushAgrawal-A2:


davidkircos:
This seems to work for me! Let's save for the next QA so we aren't modifying too much dev ops related stuff in on QA branch

davidkircos:
I enabled Transfer Acceleration in production. So it should be as simple as merging this code

davidkircos (approved):

PR #1990, opened 2024-10-24
Search dialog triggers when manually changing sheets

### LINKED ISSUE #1989: Changing sheets does not reset the Find dialog
Correct behavior: leave find dialogue the same but apply it to whatever sheet is selected instead of just keeping attached to previous sheet.

davidfig:
closes #1989

When searching on one sheet, if you manually change sheets, then the search dialog will search that sheet with the same options. When searching all sheets, then the behavior is the same as before.


quadratic-client/src/app/ui/components/Search.tsx
```diff
@@ -36,23 +36,26 @@ export function Search() {
...
-      if (found.length) {
-        setResults(found);
-        setCurrent(0);
-        moveCursor(found[0]);
-        events.emit(
-          'search',
-          found.map((found) => ({ x: Number(found.x), y: Number(found.y), sheetId: found.sheet_id.id }), 0)
-        );
-        return;
+  const onChange = useCallback(
```
ddimaria, line 39: GH should be better at diffs.  Nothing changed here except moving the function signature down a line.

davidkircos (approved):
lgtm!

PR #2004, opened 2024-10-30
feat: new file dialog replacement

jimniels:
Changes:

- "New file" buttons in sidebar create file immediately
- "New file" button in the upper right of dashboard creates file immediately
- "Import" dropdown communicates the other ways to start a new file with data in it
  - Local file (opens browser file dialog and imports user-selected files)
  - API (opens a new file with a python example code to query api)
  - Examples (navigates to examples)
  - Connections (opens a new file with an empty query to the selected connection)

Also modifies the schema browser to more clearly indicate the action you take with the widget. You can 1) drill into individual tables to explore the schema, or 2) insert a query to the selected table OR create a new file querying a specific table if you're on the dashboard

CleanShot 2024-11-18 at 13 44 45@2x

TODOs:

- [x] code todos
- [x] remove old newfiledialog code


davidkircos:
I generally think that two buttons in the top: `New File` and `Import Data` is the simplest way to split it. New file is blank new file. Import Data has the remainder of menu options.

What are you thinking for the default new user state?
CleanShot 2024-10-31 at 13 38 14@2x

jimniels:
@davidkircos 

> I generally think that two buttons in the top: New File and Import Data is the simplest way to split it. New file is blank new file. Import Data has the remainder of menu options.

That's actually the way I had it in the beginning! But since I introduce two more "new file" buttons (in the sidebar, and in the files list) I didn't think it needed a _third_ way to create a new blank file. But happy to go back to that one.

> What are you thinking for the default new user state? 

I was thinking that the empty state would just be empty and that first item would stay, e.g. like this

CleanShot 2024-10-31 at 21 17 22@2x

I agree that if we split it into "Import data..." and "New file" in the top, a _third_ way to create a new file might be a bit much.

The advantage of having a "New file" in the list view with the dashed border makes it clear that you can drag a file over that area. If we only have that as the zero state, it's easy to miss that functionality.

I'll do a pass with all three "new file" and we can see if it feels too much. If so, we can cut out the one in the file list.

jimniels:
- "New file with selected table" and make it blue like the new file button

davidkircos (approved):
functionality lgtm!

PR #2005, opened 2024-10-30
fix: support theming in app bar

jimniels:
When you add install Quadratic as a standalone app, for example using Chrome:

CleanShot 2024-10-30 at 15 08 22@2x

It gets its own app bar. This needs to adjust to the theme color when the theme changes. This adds a `<meta>` tag that browsers use as a cue for this, so when you change it in the UI, the browser will update the app bar color too.

https://github.com/user-attachments/assets/10c3e6fd-9c57-44e1-8812-3243cad0aca9


davidkircos (approved):
lgtm

PR #2033, opened 2024-11-06
Prioritize permission denied

ddimaria:
Fixes this issue from QA Wolf

Would removing a user from a team satisfy conditions 1 or 2?
The flow is this:
- newUser1 creates an acct, newUser2 creates an acct
- newUser1 creates teamFile
- newUser1 creates a team, then invites newUser2 to the team with "can edit" permission
- newUser2 is able to see the teamFile
- newUser1 removes newUser2 from team
- newUser2 clicks into the teamFile, but is met with the "License Revoked" screen.

### DIFF
--- a/quadratic-api/src/licenseClient.ts
+++ b/quadratic-api/src/licenseClient.ts
@@ -8,6 +8,7 @@ import z from 'zod';
 import dbClient from './dbClient';
 import { LICENSE_API_URI, LICENSE_KEY } from './env-vars';
 import { hash } from './utils/crypto';
+import { ApiError } from './utils/ApiError';
 
 type LicenseResponse = z.infer<typeof LicenseSchema>;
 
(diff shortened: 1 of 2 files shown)


quadratic-api/src/licenseClient.ts
```diff
@@ -26,6 +27,7 @@ export const licenseClient = {
     } catch (err) {
       if (err instanceof Error) {
         console.error('Failed to get the license info from the license service:', err.message);
```
davidkircos, line 29: Does throwing an `ApiError` automatically also console.error it?

ddimaria: I traced through the code and can't see where it does this?  Also, I never see logs in API, but so see the console.error I've added.  In the long term, we should have better logging (e.g. winston), though might be worth tabling if we decide to rewrite this service in Rust.

davidkircos (approved):

PR #2034, opened 2024-11-06
design tweaks for fix buttons

jimniels:


quadratic-client/src/app/ui/components/FixSpillError.tsx
```diff
@@ -67,17 +67,27 @@ export const CodeEditorFixSpillError = () => {
...
+          <SpillErrorMoveIcon className="mr-2" /> Move down to nearest free space
+        </DropdownMenuItem>
+        <DropdownMenuItem onClick={() => handleModeCodeCellRight(false)}>
+          <SpillErrorMoveIcon className="mr-2 -rotate-90" />
+          Move right to nearest free space
+        </DropdownMenuItem>
+      </DropdownMenuContent>
+    </DropdownMenu>
+  );
+};
```
jimniels, line 88: I want to have one component for both places where this button renders. Is that possible?

PR #2037, opened 2024-11-07
fix: hyperlinks show up in text with a colon

### LINKED ISSUE #1992: bug: hyperlinks show up in text with a colon
Type anything into a text cell with a colon and it recognizes it as hyperlink, e.g. `Jim:nielsen` will be highlighted as a URL

CleanShot 2024-10-28 at 13 56 23@2x

AyushAgrawal-A2:
closes #1992

### DIFF
--- a/quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellLabel.ts
+++ b/quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellLabel.ts
@@ -54,6 +54,8 @@ const STRIKE_THROUGH_OFFSET = 32;
 export const FONT_SIZE = 14;
 export const LINE_HEIGHT = 16;
 
+const URL_REGEX = /^(https?:\/\/|www\.)[^\s<>'"]+\.[^\s<>'"]+$/i;
+
 export class CellLabel {
   private cellsLabels: CellsLabels;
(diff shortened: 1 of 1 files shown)


luke-quadratic (approved):
works as intended

quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellLabel.ts
```diff
@@ -54,6 +54,8 @@ const STRIKE_THROUGH_OFFSET = 32;
 export const FONT_SIZE = 14;
 export const LINE_HEIGHT = 16;
 
+const URL_REGEX = /^(https?:\/\/|www\.)[^\s<>'"]+\.[^\s<>'"]+$/i;
```
davidkircos, line 57: It may be better to use the built-in URL function in JS

```
function isValidUrl(url) {
  try {
    new URL(url);
    return true;
  } catch (e) {
    return false;
  }
}
```

AyushAgrawal-A2: @davidkircos 
This is already there and was causing this bug in the first place, using regex to filter out false positives.

davidkircos (approved):
nice!

PR #2046, opened 2024-11-07
fix: dark mode issues

### LINKED ISSUE #2006: bug: dark mode issues
- [ ] rename file - can't read in dark mode, has white background
- [ ] grid text - click into a cell in dark mode, text is also dark and can't read until clicking out of that particular cell

AyushAgrawal-A2:
closes #2006


jimniels (approved):
LGTM 👍 

(I increased the contrast for the selection too, as it was _really_ dark in dark mode — still not optimal, but an improvement)

quadratic-client/src/app/ui/loading/QuadraticLoading.tsx
```diff
@@ -25,6 +25,7 @@ export function QuadraticLoading() {
   return (
     <>
       {/* ToBarLoading allows window to be moved while loading in electron */}
+      <ThemeAppearanceModeEffects />
```
jimniels, line 28: What does this fix? [This component is already mounted in the root-most component](https://github.com/quadratichq/quadratic/blob/qa/quadratic-client/src/routes/_root.tsx#L36), so this would be mounting it twice?

jimniels: Idk if it's related, but per [David's comment](https://github.com/quadratichq/quadratic/issues/2006#issuecomment-2465284171), it does seem like [the TopBarLoading component has a hard-coded background color value](https://github.com/quadratichq/quadratic/blob/qa/quadratic-client/src/app/ui/components/TopBarLoading.tsx#L10) which we should get rid of and just add `bg-background` class

AyushAgrawal-A2: QuadraticLoading is also a route fallback, 

When the app opens, we see QuadraticLoading twice, once as route fallback and another time when app is syncing offline / multiplayer transactions. I was seeing the first loading screen as white and 2nd one as black.

This effect is contained inside QuadraticLoading, which is a temporary, it will not conflict / execute twice with the other effect

AyushAgrawal-A2: Regarding the topbar, it is an extra component, I added dark-mode-hack to that as well.
I am not sure why this component exists or is it even need now.

jimniels: Yeah it was because there was an electron app at one point but there hasn't been in a looooooong time. AFAIK that's the only reason that component exists

jimniels (approved):

quadratic-client/src/app/ui/loading/QuadraticLoading.tsx
davidkircos: Jim is right that is why it was created^

davidkircos (approved):

PR #2047, opened 2024-11-07
fix: disable few default keybindings in inline monaco editor

### LINKED ISSUE #1996: Cmd+g inside a cell opens monaco menu in that cell
CleanShot 2024-10-29 at 09 50 52@2x

AyushAgrawal-A2:
closes #1996


luke-quadratic (approved):
Fixes cmd+g bug

luke-quadratic:
Noting that cmd+g from within cell opens browser's find on page. Fix is still improvement but if max nitpicking then maybe cmd+g from inside cell should open the go-to prompt in sheet?

davidfig (approved):

davidkircos:
> Noting that cmd+g from within cell opens browser's find on page. Fix is still improvement but if max nitpicking then maybe cmd+g from inside cell should open the go-to prompt in sheet?

Yeah, it's the same with Command + F when inline. It opens the browser's find menu, and it should open to sheet find menu.

PR #2048, opened 2024-11-07
fix: tab key sometimes inserts 2 spaces

### LINKED ISSUE #1993: Tab key sometimes inserts 2 spaces
1. Type `a` into A1
2. Press enter to move to A2
3. Press enter to start editing A2
4. Type `a` into A2
5. Press tab

Expected behavior: move to B2

Actual behavior: 2 spaces inserted

happens on https://quadratic-git-ayush-aipanel.vercel.quadratic-preview.com/

AyushAgrawal-A2:
closes #1993

### DIFF
--- a/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
+++ b/quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
@@ -427,7 +427,7 @@ class InlineEditorMonaco {
     monaco.languages.registerCompletionItemProvider('inline-editor', {
       provideCompletionItems: (model, position) => {
         const lowerCase = this.get().toLowerCase();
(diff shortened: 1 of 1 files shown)


HactarCE (approved):
Fix works as expected!

The code looks dubious. If empty completions are the issue, then should we add a `t.length > 0` check instead?

davidfig (approved):

davidkircos (approved):

HactarCE:
> @HactarCE
> 
> the completion should be longer than the text already there in the editor, otherwise there is nothing more to add when pressing tab. Only spaces are apppended.
> 
> Sorry, I don't get the `dubious code` or `empty string` bit

Ah ok, that makes sense. Sounds good!

Don't worry about it. "dubious" here means "strange" -- I didn't understand what was going on or why the fix worked, and now I do.

PR #2054, opened 2024-11-09
Open recent files from within a sheet

davidfig:
This adds an Open recent file > submenu to the file menu. This stores the local file list in local storage. 

- [x] add Open recent file submenu
- [x] add file that opens successfully to the top of the recent list
- [x] remove file that fails to open from the recent list
- [x] rename file updates the recent list   
- [x] delete file removes the file from the recent list

Future:
- [ ] Include the team name if it's a team file (this requires an API change)
- [ ] Move from localStorage to DB storage, so a user always has their latest files regardless of what computer they're using
- [ ] Add a normal Open... option to the file menu so you can get a list of files w/o returning to the dashboard


davidkircos:
Recent files seems like it should come from the API

davidfig:
Agreed. But for a v1, this seems to work w/o having to change APIs.

quadratic-client/src/app/actions.ts
```diff
@@ -101,6 +102,7 @@ export const deleteFile = {
       try {
         const data = getActionFileDelete({ userEmail, redirect });
         submit(data, { method: 'POST', action: ROUTES.API.FILE(uuid), encType: 'application/json' });
+        updateRecentFiles(uuid, '', false);
```
jimniels, line 105: This should happen [in `quadratic-client/src/routes/api.files.$uuid.ts` in the delete action](https://github.com/quadratichq/quadratic/blob/qa/quadratic-client/src/routes/api.files.%24uuid.ts#L33-L41) 

If you put it in this file, then if a user deletes a file from the dashboard you won't see that reflected in your files list.

quadratic-client/src/app/ui/menus/TopBar/TopBarMenus/updateRecentFiles.ts
```diff
@@ -0,0 +1,48 @@
...
+  } catch (e) {
+    console.warn('Unable to update recent files', e);
+  }
+};
+
+// Clears the recent files list in localStorage
+export const clearRecentFiles = () => {
+  localStorage.removeItem(RECENT_FILES_KEY);
+  window.dispatchEvent(new Event('local-storage'));
+};
```
jimniels, line 48: I feel like this file should be put in `quadratic-client/src/shared` somewhere, because its functionality that really is shared between the app and the dashboard, because if you delete a file on the dashboard you'll want this code to run (which is what [this comment is about](https://github.com/quadratichq/quadratic/pull/2054/files#r1848829375)).

So this should live outside of `src/app` since it's not just specific to the app, but shared between the app and the dashboard.

quadratic-client/src/dashboard/components/FilesListItem.tsx
```diff
@@ -126,12 +127,14 @@ export function FilesListItemUserFile({
     // Update on the server and optimistically in the UI
     const data: FileAction['request.rename'] = { action: 'rename', name: value };
     fetcherRename.submit(data, fetcherSubmitOpts);
+    updateRecentFiles(uuid, value, true, true);
```
jimniels, line 130: Similar feedback here, if you move these into the routes file, you won't have to find every place where you rename or delete a file across both the dashboard and the app.

davidkircos (approved):
lgtm!

jimniels (approved):
Added one small change to use ROUTES constant for file redirect, but otherwise LGTM 👍

PR #2058, opened 2024-11-10
perf: use plotly cdn for python charts

AyushAgrawal-A2:
todo: we should rerun examples having python charts, this will reduce file size and creating new files from examples will be alot faster.


davidkircos (approved):
This appears to work, even in Safari

davidfig (approved):

PR #2060, opened 2024-11-10
fix: ERROR: TypeError: this.pyodide.globals.get(...) is not a function

AyushAgrawal-A2:


quadratic-client/src/app/web-workers/pythonWebWorker/worker/python.ts
```diff
@@ -167,7 +167,7 @@ class Python {
     if (!this.pyodide) {
       console.warn('Python not loaded');
     } else {
-      const output = await this.pyodide.globals.get('inspect_python')(pythonCode);
+      const output = await this.pyodide.runPythonAsync(`inspect_python(${JSON.stringify(pythonCode)})`);
```
davidkircos, line 170: Why do you have to stringify the pythonCode? It should be a string already

ddimaria:
I ran some examples and it seemed fine.  I'm not sure why we need to stringify, but it's not a lot of overhead as code is typically small.

ddimaria (approved):

PR #2061, opened 2024-11-11
chore: lock file maintenance (dependencies update and dedupe)

AyushAgrawal-A2:
before:

image

after:

image


davidkircos (approved):
branch appears to work normally

quadratic-api/package.json
```diff
@@ -82,7 +81,7 @@
     "prisma": "^5.5.2",
     "ts-node": "^10.9.1",
     "ts-node-dev": "^2.0.0",
-    "typescript": "^5.4.5"
+    "typescript": "^5.6.3"
```
jimniels, line 84: If typescript is specified in the root `package.json`, we shouldn't have to also include it in `quadratic-api` and `quadratic-client` and `quadratic-shared` — no? Isn't that the purpose of the root `package.json`: to define (and coordinate) deps shared across multiple child packages?

quadratic-client/package.json
```diff
@@ -158,7 +157,7 @@
     "sass": "^1.69.5",
     "serve": "^14.2.0",
     "tailwindcss": "^3.3.5",
-    "typescript": "^5.4.5",
+    "typescript": "^5.6.3",
```
jimniels, line 160: Same as https://github.com/quadratichq/quadratic/pull/2061/files#r1838949694

quadratic-client/public/pyodide/package.json
```diff
@@ -35,7 +35,7 @@
     "ts-mocha": "^9.0.2",
     "tsd": "^0.24.1",
     "typedoc": "^0.25.1",
-    "typescript": "^4.6.4",
+    "typescript": "^5.6.3",
```
jimniels, line 38: I'm not sure that we want to update this...or, at least that updating it is doing anything? Like this package.json isn't actually doing anything is it? Aren't all the files in `public/` just static files that get copied over and there's no `npm i` going on with anything in here? 

Ultimately if this file doesn't actually do anything, then it doesn't really matter if you update it, other than it serves as a historical marker for when the `pyodide` package was copied in here (and what it's declared deps were when it got put here)

quadratic-shared/package.json
```diff
@@ -8,7 +8,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
-    "typescript": "^5.6.2",
+    "typescript": "^5.6.3",
```
jimniels, line 11: Same as https://github.com/quadratichq/quadratic/pull/2061/files#r1838949694

PR #2083, opened 2024-11-18
feat: request proxying for JS & fix auth header override for both JS and Python requests

### LINKED ISSUE #1785: Authorization headers in proxy being overwritten in proxy calls from the client
In `quadratic-client/src/app/web-workers/pythonWebWorker/worker/python.ts`, we're overwriting the Authorization header (or at least trying to).  This will cause problems for anyone hitting the proxy from Python that needs to have that header in place.

 - [ ] In the client, pack all user-defined headers into a body var called `user-headers`
 - [ ] In the connections service (`proxy.rs`), place all `user-headers` in the headers of the external reqwest

### LINKED ISSUE #1899: feat: request proxying for JS & fix auth header override for both JS and Python requests
Python will proxy requests so you don't hit CORS issues, but JavaScript doesn't. We should add that.

Additionally, at the same time as adding the JS proxy we should fix both proxies to not override authorization headers. See Slack convo for recommended approach to fix. 

image

AyushAgrawal-A2:
closes #1899 
closes #1785


quadratic-connection/src/proxy.rs
```diff
@@ -24,12 +26,22 @@ pub(crate) async fn axum_to_reqwest(
...
         .into_iter()
         .filter(|(name, _)| !headers_to_ignore.contains(&name.as_str()))
+        .map(|(name, value)| {
+            if name.to_string().starts_with(PROXY_HEADER_PREFIX) {
+                (
+                    name.to_string().replace(PROXY_HEADER_PREFIX, ""),
+                    value.to_str().unwrap().to_string(),
+                )
+            } else {
+                (name.to_string(), value.to_str().unwrap().to_string())
```
ddimaria, line 42: unsafe unwrap

quadratic-connection/src/proxy.rs
```diff
@@ -24,12 +26,22 @@ pub(crate) async fn axum_to_reqwest(
...
 
     for (name, value) in req
         .headers()
         .into_iter()
         .filter(|(name, _)| !headers_to_ignore.contains(&name.as_str()))
+        .map(|(name, value)| {
+            if name.to_string().starts_with(PROXY_HEADER_PREFIX) {
+                (
+                    name.to_string().replace(PROXY_HEADER_PREFIX, ""),
+                    value.to_str().unwrap().to_string(),
```
ddimaria, line 39: unsafe unwrap

quadratic-connection/src/proxy.rs
ddimaria, line 42: `value.to_str().unwrap().to_string()` is duplicated in both branches, consider assigning to a variable before the conditional.

luke-quadratic (approved):
marking approved on functionality, ran it through: 
7 different API calls, 5 Python and 2 JS, requests with and without headers, all worked great

ddimaria:
> This patch works, but does not close #1785 as it doesn't conform to the design:
> 
> > In the client, pack all user-defined headers into a body var called user-headers
> 
> This design will avoid future collisions.

@AyushAgrawal-A2 After reviewing again, I think the prefix approach works. Let me know when the other comments are addressed and I'll re-review.

quadratic-connection/src/proxy.rs
AyushAgrawal-A2: done

ddimaria (approved):

PR #2085, opened 2024-11-18
feat: update code editor header to include panel positioning control

jimniels:
Put the control for changing the panel positioning in a single place, so it's in the same place regardless of where the panel is docked.

https://github.com/user-attachments/assets/80e60e2e-580d-4d73-8a5c-ea21f873e08a


davidkircos (approved):
lgtm! much nicer UI

Issue #2077, opened 2024-11-15
AI Chat: Small Fixes


davidkircos:
- [x] Open AI chat by default for every file. Once user has closed and opened AI chat with sidebar 3 times, AI Chat open by default only on files with <=1 transaction (so file imports still see it). @jimniels #2090 
- [x] Focus grid on new file load, not chat. @AyushAgrawal-A2 #2090  
- [x] Move the chat history button to the center button, with the new chat on the left. Make new more obvious. Blue? @jimniels #2090 
- [ ] Fix with AI on the sheet hover hover should prompt left chat instead of right chat. Keep the Fix with AI button in the code editor opening in the code editor chat. Wait for A1.
- [x] While a chat response is running, the user can "send" another chat. But it just disappears. This should be prevented until the previous chat finishes. #2090 
- [x] Cmd + K to open chat #2090 
- [x] AI Chat often tries to say the answer that it has returned to the sheet, but the answer in the sheet is correct and in the AI chat it is wrong; AI chat can try to explain but it shouldn't try to also say the answer. #2090
- [x] Prompt AI to not answer questions in the chat, answer in the sheet. Often gives one answer on the sheet and one in the chat that are different. The one on the sheet is correct and the one in the chat is incorrect. #2090
- [x] Prompt AI to not use a conditional return. Quadratic does not support conditional returns. #2090
- [ ] Not AI: Code editor show collapsed by default instead of expanded.
- [x] Show actual API error to user, instead of a generic status 400. #2090
- [ ] Show Console as context in Code Editor Chat
- [ ] AI responses don't scale with text size changes in editor 

Needs scope
- [x] Make the initial state a little more "What can I help with?". @jimniels #2090
- [x] Apply flow in Chat for Code Editor is wonky. Proposal: Separating Apply and Run? @jimniels #2086 
- [x] Apply button in right chat should show up as disabled while loading code. Possibly collapse code by default and apply just shows it above? @jimniels #2086
- [x] Move thumbs up and thumbs down to be more obvious, and don't hide run button. @jimniels #2086

luke-quadratic:
Vast majority fixed - ones that are not fixed will be opened as scoped independent single issues.

PR #2090, opened 2024-11-19
fix: ai tweaks

jimniels:
Fixes for #2077, including:

- [x] Tracks the number of times a user has opened the AI Analyst and cues off that determine whether to show it by default
  - The logic is: if the user has _manually opened_ the AI chat 3 or more times, we cue off whether the sheet is empty or not. Otherwise we always show the AI Analyst open by default.
  - It's worth noting that:
    - The state of whether the AI is open or not is not a user preference and it is therefore lost on a page refresh (this is the same as the code editor, if you have the code editor open and you refresh the page, it's now closed which is the default state)
    - This requires users manually open the AI analyst _at least 3 times_ by themselves, before we will stop showing it by default to them. We don't pay any heed to whether they've ever _closed_ the AI analyst. Only opened it.
- [x] Add "what can I help with?" to empty state
- [x] Move history/new chat button
- [x] Keep focus on grid on initial load @AyushAgrawal-A2 
- [x] Cmd + k to open left side panel
- [x] While a chat response is running, the user can "send" another chat. But it just disappears. This should be prevented until the previous chat finishes.
- [x] Show actual API error to user, instead of a generic status 400.
- [x] Prompt AI to not answer questions in the chat, answer in the sheet. Often gives one answer on the sheet and one in the chat that are different. The one on the sheet is correct and the one in the chat is incorrect.
- [x] Prompt AI to not use a conditional return. Quadratic does not support conditional returns.

CleanShot 2024-11-20 at 14 41 37

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -92,5 +92,8 @@
   },
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "[html]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   }
 }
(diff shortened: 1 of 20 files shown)


davidkircos (approved):
Nice, lgtm!

PR #2093, opened 2024-11-21
fix: code cell outlines

AyushAgrawal-A2:


luke-quadratic (approved):
fixes bug on functionality yay

quadratic-core/src/controller/active_transactions/pending_transaction.rs
```diff
@@ -345,6 +345,53 @@ impl PendingTransaction {
             offsets_modified.insert((None, Some(row)), size.unwrap_or(0.0));
         }
     }
+
+    pub fn add_updates_from_transaction(&mut self, transaction: PendingTransaction) {
+        self.generate_thumbnail |= transaction.generate_thumbnail;
+
+        self.validations.extend(transaction.validations);
+
+        for (sheet_id, dirty_hashes) in transaction.dirty_hashes {
```
ddimaria, line 354: Not for now, but I would think this could be a reusable fn.

ddimaria (approved):

PR #2099, opened 2024-11-22
fix: sticky headers in js proxy request

### LINKED ISSUE #2098: bug: javascript network requests fail in a loop
It seems the proxy headers are duplicating on consecutive calls unless you're defining them inside the loop itself.

## This doesn't work

Code, where the `options` is defined at the top level of the function, then called inside a loop

CleanShot 2024-11-22 at 14 33 37@2x

Request headers for the 1st time `fetch` runs:

CleanShot 2024-11-22 at 14 34 24@2x

Request headers for the 2nd time `fetch` runs in the loop:

CleanShot 2024-11-22 at 14 34 58@2x

The 1st call succeeds. The 2nd call fails with a 401 Unauthorized error.

## This does work

Putting the options definition right alongside the `fetch` call

CleanShot 2024-11-22 at 14 36 06@2x

This makes both the 1st and 2nd requests work.

AyushAgrawal-A2:
closes #2098


jimniels (approved):
Works for me. Fixes the issue I was encountering 👍

davidfig (approved):

PR #2101, opened 2024-11-25
fix: new users sent to empty file

### LINKED ISSUE #2100: Feat: First time a user opens the app, open a blank file
@jimniels wdyt?

jimniels:
Resolves #2100

When a user lands on the dashboard and has 0 teams, create a new team and then instead of sending them to the homepage for that team, create a new file for them and send them to the file

(Note: most of the time this will be because they are a _new_ user who hasn't been associated with a team yet, but it could also be for existing users who left the last team they were a part of).


davidkircos (changes requested):
redirect to new file on desktop and dash on mobile

davidkircos (approved):
great!

PR #2120, opened 2024-12-10
Clear borders and formatting monstrosity

davidfig:


quadratic-core/src/clear_option.rs
```diff
@@ -0,0 +1,121 @@
...
+            None => ClearOption::Clear,
+        }
+    }
+}
+
+impl<T> ClearOption<T> {
+    pub fn unwrap(self) -> T {
+        match self {
+            ClearOption::Some(value) => value,
+            ClearOption::Clear => panic!("ClearOption is Clear"),
```
ddimaria, line 36: Is this panic necessary?

quadratic-core/src/grid/sheet/borders_a1/borders_col_row.rs
```diff
@@ -15,19 +15,43 @@ impl BordersA1 {
...
 
     pub fn copy_column(&self, column: i64) -> Option<BordersA1Updates> {
         let updates = BordersA1Updates {
-            left: self.left.copy_column(column),
-            right: self.right.copy_column(column),
-            top: self.top.copy_column(column),
-            bottom: self.bottom.copy_column(column),
+            left: self
+                .left
+                .copy_column(column)
```
ddimaria, line 41: You could remove a lot of boilerplate here by deriving From/Into for whatever the return value is for copy_column.  The same applies for the remove_row below.  There are also other ways to keep this DRY.

ddimaria (approved):

PR #2121, opened 2024-12-11
feat: new file with prompt

jimniels:
## Summary

Allows people to land on the app and have a file created with a prompt run automatically.

- Any request to `/files/create` from outside of the app will create a new file for a user.
- If an included `prompt` search param is included, it will start a file with an AI prompt, e.g. `files/create?prompt=generate+some+python+code`

Under the hood, this works by this logic:

- See if the user belongs to a team
  - If yes, create a _private_ file in that team
  - If no, create a team first, then create a _private_ file in that team

Also: this PR includes a piece of data on the `/acknowledge` endpoint which will tell you when a new user was created on that call.

## To Test

There are three cases we should test:

1. **You're logged in to the app already.** It should go from marketing site to app and run prompt in a new _private_ file
2. **You're not logged in to the app, but have an account.** It should go from marketing site, to login screen, to app and run prompt in a new _private_ file
3. **You don't have an account yet**. It should go from marketing site, to login page, to sign up page, to app and run prompt in a new _private_ file


quadratic-client/src/routes/login-result.tsx
```diff
@@ -11,7 +11,20 @@ export const loader = async () => {
     if (isAuthenticated) {
       // Acknowledge the user has just logged in. The backend may need
       // to run some logic before making any other API calls in parallel
-      await apiClient.users.acknowledge();
+      const { userCreated } = await apiClient.users.acknowledge();
+
+      // Special case for first-time users
+      if (userCreated) {
+        console.log('First-time user created');
```
davidkircos, line 18: Remove

davidkircos (approved):

PR #2139, opened 2024-12-20
feat: persist active sheet in URL

jimniels:
Per [this conversation in Slack](https://quadratichq.slack.com/archives/C03KXQTH5TP/p1734709691784069), this updates the app so the current active sheet is stored in the URL (and therefore part of the share process).

Functionality:

- When the user changes to any sheet (other than the first one), the URL updates to include a `sheet` param indicating which sheet is currently active, e.g. `?sheet=<uuid>`
- When the user changes to the first sheet, the `sheet` param is removed

To test:

- Change the active sheet in the sheet bar (or command palette) and see the active sheet show up in the URL bar.
- Share the link to a sheet and see that it contains the active sheet in the URL
- Load a file with a sheet name that doesn't and see that it just defaults to the first sheet

---

This also resolves [an older issue](https://github.com/quadratichq/quadratic/issues/732) by introducing the functionality into the app.


cole-at-quadratic:
CleanShot 2025-01-16 at 17 21 52
Additionally, we could offer another option in the right-click menu for "Copy link to this tab" (not essential though)

jimniels:
Updated.

Now when the user changes tabs, the URL updates with the current sheet UUID. However "Copy link" from the share tab just copies the base file URL

davidkircos (approved):

PR #2154, opened 2025-01-03
More Formulas: VAR, STDEV, PMT

luke-quadratic:


HactarCE (changes requested):
Looks good! Before merging:
- Run `cargo fmt`
- There is already a function `crate::util::assert_f64_approx_eq()` that works similarly to `assert_close()` but doesn't take a message. Could we use that, and possibly add a `message: &str` parameter to it?

luke-quadratic:
> * add a `message: &str` parameter to it

@HactarCE Awesome, thanks, good idea. 

Required editing where that function was used in other places as well (trig and math functions).

HactarCE (approved):
Looks good to me!

PR #2155, opened 2025-01-05
add try_from for Rgba

golok727:
allows to try from `#rgb`, `#rgba`, `#rrggbb`, `#rrggbbaa`, `rgb(r, g, b)`, or `rgba(r, g, b, a)`


quadratic-core/src/color.rs
```diff
@@ -99,10 +174,7 @@ mod tests {
     #[parallel]
     fn new() {
         let color = Rgba::new(0x22, 0x44, 0x66, 0xff);
-        assert_eq!(color.red, 0x22);
-        assert_eq!(color.green, 0x44);
-        assert_eq!(color.blue, 0x66);
-        assert_eq!(color.alpha, 0xff);
+        assert_eq!(color, Rgba::new(0x22, 0x44, 0x66, 0xff));
     }
```
AyushAgrawal-A2, line 178: this will always pass, both the arguments passed into `assert_eq!` are effectively the same

golok727: I made the change for consistency, but I can revert it if you'd prefer.

AyushAgrawal-A2: for consistency ? what is the purpose of this test now... it was checking individual struct fields before

golok727: I changed it to align with the other tests I added—it's effectively the same.

golok727: Ah sorry! I was looking at the wrong one.

davidkircos:
Hey @golok727, thanks for the contribution!

What functionality does this add to the application?

davidkircos:
This functionality is not needed

PR #2156, opened 2025-01-05
Option::unwrap in const is stable now use it

golok727:
https://doc.rust-lang.org/stable/core/option/enum.Option.html#method.unwrap


quadratic-core/src/values/array_size.rs
```diff
@@ -50,13 +50,10 @@ impl TryFrom<(u32, u32)> for ArraySize {
...
 
-// TODO(ddimaria):`[][0]` is now being detected by clippy, fix this
-#[allow(clippy::out_of_bounds_indexing)]
 impl ArraySize {
-    #[allow(unconditional_panic)]
-    pub const _1X1: Self = match NonZeroU32::new(1) {
-        Some(one) => ArraySize { w: one, h: one },
-        None => [][0], // workaround for `.unwrap()` being unstable in const context
+    pub const _1X1: Self = ArraySize {
+        w: NonZeroU32::new(1).unwrap(),
```
ddimaria, line 55: Can you replace these `unwraps()` with `expect("Unable to create _1X1 ArraySize")`

golok727: Done

ddimaria (approved):

PR #2163, opened 2025-01-06
feat: prompt logging & AI feature preferences

jimniels:
**Feature**

- Allow users to turn AI-related features on or off at the team level (on by default)
  - Only can be turned on/off by team owners
- Log user prompts

**TODO**

- [x] Need a link to a community post that talks about prompt logs (the one we can link to from the UI)
  - Links to https://docs.quadratichq.com/security#ai-data-privacy-setting
- [x] Add "zero data retention" on Teams page @jimniels 
- [x] Add mix panel events when thumbs up/down is clicked.
- [x] Don't record thumbs up/down if feature is turned off (support both client and server)

**To Test**

- Team owners of a team can turn the setting on/off
- Team viewers/editors cannot see the setting
- ...

---

Other things included in this PR:

- Preparations to [update react-router to major version 7](https://reactrouter.com/upgrading/v6). This primarily means upgrading to the latest 6.x version and adding all the necessary feature flags. We can ship this, and in a future release, assuming we don't get any bugs, [do the rest of what's necessary to upgrade to version 7](https://reactrouter.com/upgrading/v6#upgrade-to-v7). Note: react-router v7 will require node 20, so we'll have to make sure to make that update. While that really just means updating to node 20 for building the app, we also use node on the server and should upgrade that too while we're at it.
- Support visiting `/logout` directly to log out of the app
- Remove `TEAM_DELETE` and `TEAM_BILLING_EDIT` so now `TEAM_MANAGE` is inclusive of both of these, plus turning AI features on/off
- Update 'Chat' label to 'Sheet chat' and 'Code chat' in relevant places


davidkircos:
If a user Thumbs down a prompt, it would be great to be able to collect a text box of feedback

quadratic-api/prisma/migrations/20250108212737_add_team_preference_save_prompt/migration.sql
davidkircos, line 1: Squash migrations

AyushAgrawal-A2: done

davidkircos (approved):
funcitonality lgtm!

PR #2166, opened 2025-01-07
Add `EDATE()` and `EOMONTH()` formula functions

HactarCE:
Quadratic's implementation of `EDATE()` doesn't use serial numbers for dates because [even the Excel devs can't remember their own semantics correctly](https://github.com/user-attachments/assets/f665e565-22d0-4311-8d8d-e7f9c985d599), and we have a proper type system anyway.

Excel docs:
- [`EDATE()`](https://support.microsoft.com/en-us/office/edate-function-3c920eb2-6e66-44e7-a1f5-753ae47ee4f5)
- [`EOMONTH()`](https://support.microsoft.com/en-us/office/eomonth-function-7314ffa1-2bc9-4005-9d66-f49db127d628)


ddimaria (approved):
This code looks good!  @luke-quadratic what needs to be done in the docs to reflect these additions?

PR #2167, opened 2025-01-07
feat: support duplicating example files from outside the app

### LINKED ISSUE #2147: Update website examples to create a duplicate when clicked
Currently when a user opens up an Example file from https://www.quadratichq.com/examples, the user is unable to interact with the spreadsheet because it is read-only. We want users to be able to interact with these examples so they can see the power of Quadratic and instantly gain value from it. When a user opens an example template from within the app, a duplicate is created so that they can edit it right away. The same should be true for users coming from our marketing website.

Users should be able to duplicate read-only files that have no connections, but that's a separate issue.

jimniels:
Fixes #2147 

**To test**

Visit `/files/create?example=<example-file-url-in-prod>`

Example:

https://quadratic-git-duplicate-example-files.vercel.quadratic-preview.com/files/create?example=https://app.quadratichq.com/file/b90a8c28-4573-4073-bfaa-5bf14e808f51

Implemented by: https://github.com/quadratichq/quadratic-website/pull/240

### DIFF
--- a/quadratic-client/src/routes/files.create.tsx
+++ b/quadratic-client/src/routes/files.create.tsx
@@ -1,15 +1,36 @@
 import getActiveTeam from '@/dashboard/shared/getActiveTeam';
 import { apiClient } from '@/shared/api/apiClient';
-import { ROUTES } from '@/shared/constants/routes';
+import { ROUTES, SEARCH_PARAMS } from '@/shared/constants/routes';
 import { LoaderFunctionArgs, redirect } from 'react-router-dom';
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
   const url = new URL(request.url);
-  const prompt = url.searchParams.get('prompt');
 
+  // Get the active team
   const { teams } = await apiClient.teams.list();
   const { teamUuid } = await getActiveTeam(teams, undefined);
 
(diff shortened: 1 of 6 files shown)


quadratic-client/src/routes/files.create.tsx
```diff
@@ -1,15 +1,36 @@
...
+  // (They may have write access to another team, but not the 'active' one.)
+  const team = teams.find(({ team }) => team.uuid === teamUuid);
+  if (!team?.userMakingRequest.teamPermissions.includes('TEAM_EDIT')) {
+    return redirect(
+      `/?${SEARCH_PARAMS.SNACKBAR_MSG.KEY}=${encodeURIComponent('Failed to create file. You can only view this team.')}`
+    );
+  }
+
+  // Are they trying to duplicate an example file? Do that.
+  const example = url.searchParams.get('example');
```
ddimaria, line 23: Do they need `TEAM_EDIT` to get to this point?

ddimaria (approved):
Code looks good, I just had one question.

quadratic-client/src/routes/files.create.tsx
jimniels: It's definitely an edge case that I think will _rarely_ happen, especially since this is a route only we know about and only we use.

PR #2168, opened 2025-01-07
Add `IFNA()` formula function

HactarCE:
[Excel docs](https://support.microsoft.com/en-us/office/ifna-function-6626c961-a569-42fc-a49d-79b4951fd461)


quadratic-core/src/error_run.rs
```diff
@@ -105,8 +103,8 @@ pub enum RunErrorMsg {
     Overflow,
     DivideByZero,
     NegativeExponent,
-    NotANumber,
-    Infinity,
+    /// NaN or ±Infinity
```
ddimaria, line 106: Out of curiosity, why are we combining these?

ddimaria (approved):
Code checks out, I just have one non-blocking question.

PR #2179, opened 2025-01-10
feat: remove dependence on mui for spinner

jimniels:
Replace all instances of `<CircularProgress>` with our own custom spinner. This removes our dependence on the MUI package.

The new `<SpinnerIcon>` uses the `progress_activity` from our icon set and adds a spinning animation on it.

CleanShot 2025-01-09 at 16 47 29@2x

### DIFF
--- a/quadratic-client/src/app/gridGL/HTMLGrid/codeRunning/CodeRunning.tsx
+++ b/quadratic-client/src/app/gridGL/HTMLGrid/codeRunning/CodeRunning.tsx
@@ -3,7 +3,8 @@ import { sheets } from '@/app/grid/controller/Sheets';
 import type { CodeRun } from '@/app/web-workers/CodeRun';
 import { LanguageState } from '@/app/web-workers/languageTypes';
 import { MultiplayerUser } from '@/app/web-workers/multiplayerWebWorker/multiplayerTypes';
-import { CircularProgress } from '@mui/material';
+import { SpinnerIcon } from '@/shared/components/Icons';
+import { cn } from '@/shared/shadcn/utils';
 import { useEffect, useState } from 'react';
 import './CodeRunning.css';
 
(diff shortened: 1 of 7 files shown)


quadratic-client/src/app/gridGL/HTMLGrid/codeRunning/CodeRunning.tsx
```diff
@@ -134,14 +135,19 @@ export const CodeRunning = () => {
...
-            color={code.color === 'black' ? 'primary' : undefined}
-            size={`${CIRCULAR_PROGRESS_SIZE}px`}
-            key={index}
-            sx={{ position: 'absolute', left: code.left, top: code, color: code.color }}
-          />
-        ))}
+        .map((code, index) => {
+          return (
+            <span
+              className="-translate-x-[2px] -translate-y-[4px] scale-75"
```
jimniels, line 141: FWIW: this is just a little futzing with the positioning by nudging things around until they looked right (since we're shirking the icon with `scale-75` we have to get its position juuuuuuuuust right.

quadratic-client/src/app/gridGL/HTMLGrid/codeRunning/CodeRunning.tsx
```diff
@@ -134,14 +135,19 @@ export const CodeRunning = () => {
...
+              style={{ position: 'absolute', left: code.left, top: code.top }}
+            >
+              <SpinnerIcon
+                className={cn(code.color === 'black' && 'text-primary')}
+                style={code.color !== 'black' ? { color: code.color } : {}}
```
jimniels, line 146: From my testing, it doesn't look like `code.color` was actually doing anything. It was always returning `'black'` but I still left the logic the same (for now).

PR #2180, opened 2025-01-10
feat: trigger file picker from "Insert > Data > From file"

jimniels:
When a user goes to "Insert -> Data -> From file", now the system file picker will be triggered (rather than what it currently does, which is trigger a toast that says "drag and drop a file").

When a file is chosen, it will insert data at the current cursor position.

This functionality is uniform between the "Insert..." menu and the "Import data" popup that shows when the sheet is empty.

CleanShot 2025-01-10 at 10 13 31@2x


davidkircos (approved):
Approved functionality! Nice change

PR #2185, opened 2025-01-13
fix: proxy calls are cached

### LINKED ISSUE #2138: bug: proxy calls are cached
Per [this conversation in Slack](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1734468535140319), proxy calls can get cached by the browser, so that each request returns the same data and the only way to make them work as expected is to open the developer tools and check the "disabled cache" box. Per Ayush:

> I think this is happening due to proxy header prefix change I did in https://github.com/quadratichq/quadratic/pull/2083 , caches of request are getting mixed somehow...

AyushAgrawal-A2:
closes #2138

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4478,7 +4478,7 @@ dependencies = [
 
 [[package]]
 name = "quadratic-connection"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "arrow 51.0.0",
  "arrow-schema 51.0.0",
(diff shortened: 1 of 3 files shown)


ddimaria (approved):

jimniels:
LGTM 👍  API calls that were previously caching seem to be properly updating anytime I change the URL

jimniels (approved):

PR #2189, opened 2025-01-13
Add `SUMPRODUCT()` formula

HactarCE:
This also fixes a relevant parsing bug. The `SUMPRODUCT()` tests act as a regression tests for the parsing bug.


ddimaria (approved):

luke-quadratic (approved):
Works great

PR #2197, opened 2025-01-15
try o1

AyushAgrawal-A2:


luke-quadratic:
I think we can probably close this for now - I think rough conclusion from testing is we should add o1-mini for the time being

PR #2202, opened 2025-01-16
fix: long running lint job on ci

AyushAgrawal-A2:


davidkircos (approved):
Nice is significantly faster for unit tests and lint!

Issue #2187, opened 2025-01-13
Bug: Drag to move a block of cells doesn't work if selection is made from bottom up 


davidkircos:
![Image](https://github.com/user-attachments/assets/98abe29c-9005-4be9-bbda-9ac64b3e697a)

Issue #1412, opened 2024-06-05
Data tables 


luke-quadratic:
**Data Tables v1: Name Data Tables**

- [ ] Rust code to support CodeRunResult metadata
  - [ ] Create new CellValue::Import variant
  - [ ] A DataTable is a struct that has a name
  - [ ] A DataTable enter to be added to `CodeCellValue` and `Import`
- [ ] Column Names (default to numbered).  How do we know if existing data has column headers?
- [ ] UI to support adding/editing names
- [ ] Convert parquet and csv imports into CodeRunResults automatically
- [ ] Update Python and JS APIs to support table names
- [ ] AI updates

**Data Tables v2: sorting and filteirng**
Output CSVs, requests, connection responses, and all other dataset functions as data tables that can be sorted, ordered, etc. 

#408 to be solved at same time for consuming these named tables/ranges in code, formulas, etc. 

**Data table initial feature set:** 
- [ ] Sort ascending, descending by column 
- [ ] Hide columns or rows 
- [ ] Reorder columns or rows 
- [ ] Change number of total rows displayed 
- [ ] Filtering by conditions

PR #2216, opened 2025-01-17
fix: move cells in multi selection only works from sourcecell

golok727:
closes: https://github.com/quadratichq/quadratic/issues/2187


davidkircos:
will be fixed in #1412

PR #2226, opened 2025-01-18
Add an AI planning mode

luke-quadratic:
AI planning mode puts model in a mode where the first prompts in planning mode outline a plan to be taken. 

The model then outlines its plans and lets users refine the plan before executing. 

Once done planning, the model walks through the execution with the user.


luke-quadratic:
After playing with this a bunch it feels much more relevant in a data notebook than a spreadsheet, closing for now with plans to revisit in future.

PR #2227, opened 2025-01-19
Edit docker file to match docs

luke-quadratic:
Open Q on whether it's this update or the docs that need updated. The docs instruction "npm run docker:up" doesn't work but "npm run docker:dev" does, but in dev localstack doesn't start so local doesn't work. 

If this isn't the right change then there's something else I don't understand that will require a docs change to update how docker should be started correctly.


luke-quadratic:
Closed, will be expressed in reverse: docs change, not docker change

PR #2229, opened 2025-01-20
Add AI rules

luke-quadratic:


luke-quadratic:
> Move settings to labs

Addressed

PR #2250, opened 2025-01-30
fix: save active sheet state in URL

jimniels:
Fixes [this issue](https://quadratichq.slack.com/archives/C03KXQTH5TP/p1738189574052989) where the active sheet state was put in the URL as `?sheet=<uuid>`, but wasn't reloading that sheet when you reloaded the page

To test, open a file and create multiple sheets. Then:

- When you switch between sheets, or create a new one, or delete the one you're looking at, see that the URL updates to `?sheet=:id` **_except_ the first one.** The first sheet will always omit `?sheet=:id` from the URL
- When you reload the page with a `?sheet=:id` param present, that sheet should be active when the app reloads
- If you load the page with a sheet that doesn't exist, e.g. `?sheet=foo`, it should show the first sheet (and wipe the `sheet` param from the URL)
- If you drag to reorder sheets in the sheet bar, you should see the `?sheet=:id` param update in the URL 
  - e.g. move to a sheet that's not first. see that `?sheet=...` is in the URL, then drag it to be first, and the `?sheet` param should disappear.


davidkircos (changes requested):
invalid sheet name causes rust_client panic 
CleanShot 2025-01-30 at 17 03 03@2x

jimniels:
Ah duh. I moved this from being in React to being in the sheets class, so it needs to check. Will fix

jimniels:
Fixed. Ready for review again.

davidkircos (approved):

PR #2251, opened 2025-01-30
fix: unsafe unwraps

AyushAgrawal-A2:


quadratic-rust-shared/src/sql/mssql_connection.rs
```diff
@@ -101,7 +101,7 @@ impl Connection for MsSqlConnection {
         if let Some(username) = &self.username {
             config.authentication(AuthMethod::sql_server(
                 username,
-                self.password.as_ref().unwrap(),
+                self.password.as_deref().unwrap_or_default(),
```
ddimaria, line 104: I think we can return an error here (just like we do in `port` above).

AyushAgrawal-A2: done

ddimaria (approved):

PR #2254, opened 2025-01-31
fix: update file updated date when a new checkpoint is made

### LINKED ISSUE #2088: Last updated not tracking secondary sheets
Dashboard is only updating when the first sheet in a file is changed 

If you change any of the other sheets the dashboard doesn't update your file as the last one edited

jimniels:
Fixes #2088

To test:

- Create multiple files in the dashboard
- For the oldest file in the dashboard, open it, and navigate to the last sheet (or any sheet that's not the first one in the list)
- Make an edit in that sheet
- Navigate to the dashboard
- The file you just edited should now be the first one in the list


quadratic-api/src/routes/internal/checkpoint.$fileUuid.PUT.ts
```diff
@@ -39,7 +39,7 @@ router.put(
 
     const result = await dbClient.$transaction(async (prisma) => {
       // Retrieve the latest checkpoint
-      const lastTransactionQuery = await dbClient.file.findUniqueOrThrow({
+      const lastTransactionQuery = await prisma.file.findUniqueOrThrow({
```
jimniels, line 42: Since this is happening inside a transaction, it should use `prisma` here rather than `dbClient` — even thought for this context they're likely the same thing. Should help prevent future possible bugs though.

davidkircos (changes requested):
Looking at this logic, I think that it could all be combined into a single database call.

Endpoint 
- Takes params from Files service, validates them.
- Creates a checkpoint, and updates date on file in postgres db in a single query.
- Validates the checkpoint is greater than the last one saved, if not reverts
- Returns the newly created checkpoint.

```typescript
router.put(
  '/file/:uuid/checkpoint',
  validateM2MAuth(),
  requestValidationMiddleware,
  async (req: Request, res: Response) => {
    
    const errors = validationResult(req);
    if (!errors.isEmpty()) {
      return res.status(400).json({ errors: errors.array() });
    }

    try {
      const fileUuid = req.params.uuid;
      
      const result = await dbClient.$transaction(async (prisma) => {
        const newCheckpoint = await prisma.fileCheckpoint.create({
          data: {
            sequenceNumber: req.body.sequenceNumber,
            s3Bucket: req.body.s3Bucket,
            s3Key: req.body.s3Key,
            version: req.body.version,
            file: {
              connect: { uuid: fileUuid },
              update: { updatedDate: new Date() }
            }
          },
          select: {
            sequenceNumber: true,
            version: true,
            s3Key: true,
            s3Bucket: true,
            file: {
              select: {
                FileCheckpoint: {
                  orderBy: { sequenceNumber: 'desc' },
                  take: 2,
                }
              }
            }
          }
        });

        // The first checkpoint will be our newly created one
        // The second (if it exists) would be the previous highest
        const [_, previousCheckpoint] = newCheckpoint.file.FileCheckpoint;
        
        if (previousCheckpoint && previousCheckpoint.sequenceNumber >= req.body.sequenceNumber) {
          throw new Error('SEQUENCE_NUMBER_INVALID');
        }

        return newCheckpoint;
      });

      return res.status(200).json({
        fileUuid: fileUuid,
        lastCheckpoint: {
          sequenceNumber: result.sequenceNumber,
          version: result.version,
          s3Key: result.s3Key,
          s3Bucket: result.s3Bucket,
        },
      });
    } catch (error) {
      if (error.message === 'SEQUENCE_NUMBER_INVALID') {
        return res.status(400).json({ 
          error: 'Invalid sequence number: must be greater than the latest checkpoint' 
        });
      }
      if (error.code === 'P2025') {
        return res.status(404).json({ error: 'File not found' });
      }
      console.error('Checkpoint creation error:', error);
      return res.status(500).json({ error: 'Internal server error' });
    }
  }
);
```

jimniels:
Idk, database calls are not my strongpoint. If I try to use the code above, I get errors. If I ask about it cursor, it says prisma doesn't support these kinds of operations and tries to rewrite it to multiple calls (like it is today).

CleanShot 2025-02-03 at 11 21 05@2x

This might be something I would need your help on @davidkircos (or maybe @ddimaria )

ddimaria:
> Idk, database calls are not my strongpoint. If I try to use the code above, I get errors. If I ask about it cursor, it says prisma doesn't support these kinds of operations and tries to rewrite it to multiple calls (like it is today).
> 
> This might be something I would need your help on @davidkircos (or maybe @ddimaria )
It looks like Prisma supports transactions: https://www.prisma.io/docs/orm/prisma-client/queries/transactions

I'm not sure if @davidkircos wanted the calls merged for atomic reasons, or if it was just an efficiency thing?

jimniels:
@ddimaria I'm pretty sure @davidkircos wanted them merged just as an efficiency thing. But I can't figure out how to combine a create and an update in a single query with Prisma.

FWIW: this entire thing is already happening inside a transaction. [According to their docs](https://www.prisma.io/docs/orm/prisma-client/queries/transactions#example), which show an example of doing separate updates within a single transaction:

> [both] queries run within a database transaction. When the application reaches the end of the function, the transaction is committed to the database.

I'm not a db expect, but that sounds like what David K. wanted in the first place? e.g. a single db transaction.

The provided code is a pretty substantial rewrite of what this endpoint is doing and idk if I feel totally confident making that change because I didn't write this endpoint.

So I say:

1. Merge this PR as-is. It's two separate `prisma.X` calls, but they're happening inside a single DB transaction which seems to be what prisma recommends. If it becomes a bottleneck, we can deal with it.
2. We sit down and talk about how we want to rewrite this endpoint to match David K.'s expectations and I work with David D. updating it, because I'm not familiar enough with file checkpoints to feel good making changes here.
3. Consider a different approach, where we no longer use `updated_date` on the `File` table but instead always look at the file's checkpoint anytime we want to know the last time it was updated. 

I don't think we want to do no. 3 because then we have to solve for a different set of problems in the API, like when metadata around a file is updated (like the file name, or the file thumbnail) then we have to decide if that means "the file was updated" even though it's not a file checkpoint.

davidkircos (approved):
lg to merge as is

PR #2271, opened 2025-02-11
chore: track ai usage

AyushAgrawal-A2:


quadratic-shared/ai/models/AI_MODELS.ts
```diff
@@ -25,6 +17,10 @@ export const MODEL_OPTIONS: {
     canStreamWithToolCalls: true,
     enabled: true,
     provider: 'openai',
+    rate_per_million_input_tokens: 2.5,
```
davidkircos, line 20: Is this the price for each model?

Seems like we shouldn't store this in our code base as it will presumably change over time and be a chore to update

AyushAgrawal-A2: done

davidkircos (approved):
lgtm! Merging into billing branch. 

Thanks @AyushAgrawal-A2 !

PR #2281, opened 2025-02-14
feat: allow duplicate file outside context of "current" team

### LINKED ISSUE #2252: feat: Add an option Share Readonly, Allow Duplicate (for content creators / marketing)
Scope:
- [ ] Add a checkbox to the share dialog for allow duplicate for public share link (or we just make this the default?) Needs scope @jimniels 
- [ ] Add ui for logged out users and logged in users to duplicate a readonly file easily in this bar. I'm thinking for anmoyous users we have two buttons "Duplicate" which redirects to signup?next=duplicate_file and "Login / Signup"

Image

jimniels:
Allow users to duplicate files that are in teams they don't have access to.

Logged out users will see:

CleanShot 2025-02-19 at 17 11 29@2x

Logged in users will see:

CleanShot 2025-02-19 at 17 11 52@2x

When you duplicate a file from the sheet side of the app (or from one of the permission messages you see above), it will duplicate the file to your active team. If you have access to more than one team, it will route you to a page and ask where you want to duplicate the file.

CleanShot 2025-04-04 at 15 02 40@2x

Resolves #2252 

### Duplicating a file

Notes on testing around this feature:

- "Duplicate file" from the app will always duplicate the file to your private collection of team files ("My files").
- If you can access the file, you can duplicate it. That includes:
    - File is on a team you belong to (as viewer or editor)
    - File was shared with you via public link (edit or view access)
    - You were invited to the file via email (viewer or editor)

As a logged in user:

- [x] Duplicate a file in a team you belong to
- [x] Duplicate a file in a team you DO NOT belong to

As a logged out user:

- [x] Duplicate a file in a team you belong to
- [x] Duplicate a file in a team you DO NOT belong to

As a new user:

- [x] Duplicate a file (into a new team that gets created when you sign up)

Ways to duplicate:

- [x] File menu: File -> Duplicate
- [x] Command palette: File -> Duplicate
- [x] Read-only file alert (logged in but can't edit a file)
- [x] Welcome to Quadratic file alert (not logged in)

Ensure that any connections in a file cannot be run (or their schema viewed) if the connections belong to another team. To test this, do the following:

- [x] Create a file, duplicate it to another team you belong to, then ensure that the connection cannot be run or viewed.
- [x] Create a file, share it with another Quadratic account, have that other account duplicate the file, then ensure that the connection cannot be run or viewed.

### Routes

> [!NOTE]  
> This PR changes a lot of the routing, we should ensure all routes work as expected. This includes fixes to exisiting special handling for new users.

This includes URL routes for users coming to the app from the marketing site. We should make sure these work as expected **for both existing users and new users after sign up**.

- [x] [`/files/create`](https://duplicate-file.quadratic-preview.com/files/create) - create a new file
- [x] [`/files/create?prompt=draw+a+chart`](https://duplicate-file.quadratic-preview.com/files/create?prompt=draw+a+chart)
- [x] [`/files/create?example`](https://duplicate-file.quadratic-preview.com/files/create?example=https://app.quadratichq.com/file/08f9dc30-6c5f-4f13-bb97-6284ad8195ab) - create a file from an example
- [x] [`/education/enroll`](https://duplicate-file.quadratic-preview.com/education/enroll) - check enrollment status in education program
- [x]  [`/file/:uuid/duplicate`](https://duplicate-file.quadratic-preview.com/file/bf7c339f-ee2e-425f-a9d9-62c62c59c214/duplicate) - duplicates a file immediately (if you have access — try having two teams and you'll see the team picker when you access this route)
- [x] `/team/*` - links directly to pages on team, such as:
    - [`/team/members`](https://duplicate-file.quadratic-preview.com/team/members)
    - [`/team/settings`](https://duplicate-file.quadratic-preview.com/team/settings)
    - [`/team/connections`](https://duplicate-file.quadratic-preview.com/team/connections)

### Connection permissions

> [!NOTE]  
> This PR changed the server-side API in relation to connections, so that functionality should be tested

- [x] Create a connection
- [x] Update a connection
- [x] Delete a connection
- [x] Run a connection in a file


jimniels:
Latest status at the time of leaving this off:

- Preview deploy isn't working the same way as local. You can duplicate a file locally but doing it on the preview build results in a "file validation error" in the UI with a 4xx error from the file service.
- UI probably needs a re-think to make it clear how/where duplicate is happening.

quadratic-api/src/routes/v0/files.$uuid.checkpoints.GET.ts
```diff
@@ -65,7 +69,10 @@ async function handler(req: Request, res: Response<ApiTypes['/v0/files/:uuid/che
   );
 
   return res.status(200).json({
-    name,
+    file: { name },
+    team: {
+      uuid: teamUuid,
+    },
```
jimniels, line 75: I had to update this endpoint for the "File history" UI.

Previously that UI was relying on the implicit "active team" on the client, which means there's a case where you could have an active team (as set in the dashboard) and duplicating from the file history would duplicate to that team (rather than the one its in).

This changes the file history UI to explicitly only allow duplicating previous versions of a file to the same team as the file is in (which needs to be returned from the API)

quadratic-client/src/auth/auth.ts
```diff
@@ -46,26 +47,27 @@ export const authClient: AuthClient = getAuthClient();
...
+  // If the user is authenticated, make sure we have a valid token
+  // before we load any of the app
+  await authClient.getTokenOrRedirect();
 
-    return loaderFn(loaderFnArgs);
-  };
+  // If the user is authenticated, make sure we have a team to work with
+  const activeTeamUuid = await getOrInitializeActiveTeam();
+
+  return { activeTeamUuid };
```
jimniels, line 70: Previously, this was a function that wrapped a react router `loader` and ensured the user was logged in.

However, it was only used in one place (a single route that wrapped all "protected" routes). That pattern didn't actually work as expected and protect all its child routes.

What was happening was all matched routes would fire their loaders in parallel, so if you hit `/` it would fire multiple loaders, including the "protected" ones which means API calls would be made for non-logged in users (they returned 4xx errors, but they were still unnecessary and might've been causing some other weird side effects we were seeing in the API).

This fixes the issue. You should be able to go to the app (any route, e.g. `/` or `/files/create` etc) and see that no API calls are made. Instead, it will figure out that you are not authenticated before any of those API calls fire because `requireAuth()` is used as the 1st line of every route that is protected.

(the `throw` in here will throw up to the loaders where they're used and therefore return the redirect to login)

quadratic-client/src/dashboard/components/TeamSwitcher.tsx
```diff
@@ -41,8 +42,8 @@ export function TeamSwitcher({ appIsLoading }: Props) {
   return (
     <DropdownMenu>
       <DropdownMenuTrigger className={cn(`gap-2 py-1 text-sm font-semibold`, sidebarItemClasses.base)}>
-        <div className="mx-0.5 flex h-5 w-5 flex-shrink-0 items-center justify-center rounded bg-foreground capitalize text-background">
-          {activeTeamName.slice(0, 1)}
+        <div className="mx-0.5">
+          <TeamAvatar name={optimisticActiveTeamName} />
```
jimniels, line 46: Componentized this because we use it on the duplicate file team picker now — but also this was technically a small bug where we weren't displaying the activeTeamName when a team was being renamed — so fixed!

quadratic-client/src/routes/education.enroll.ts
```diff
@@ -0,0 +1,13 @@
...
+import { redirect, type LoaderFunctionArgs } from 'react-router-dom';
+
+/**
+ * Checks the user's status, then sends them to the dashboard with the education dialog
+ */
+export const loader = async (loaderArgs: LoaderFunctionArgs) => {
+  await requireAuth(loaderArgs);
+  await apiClient.education.refresh();
+  return redirect(`/?${SEARCH_PARAMS.DIALOG.KEY}=${SEARCH_PARAMS.DIALOG.VALUES.EDUCATION}`);
+};
```
jimniels, line 13: This was previously defined inline in the router, but now has its own file route definition.

quadratic-client/src/routes/file.tsx
```diff
@@ -6,9 +6,9 @@ import * as Sentry from '@sentry/react';
 import mixpanel from 'mixpanel-browser';
 import { useEffect } from 'react';
 import { Outlet } from 'react-router-dom';
-import { Empty } from '../../shared/components/Empty';
+import { Empty } from '../shared/components/Empty';
 
-export function BrowserCompatibilityLayoutRoute() {
+export function Component() {
```
jimniels, line 11: Since this was just a route component, we moved it into `routes/`

quadratic-client/src/routes/logout.tsx
```diff
@@ -3,13 +3,13 @@ import { redirect } from 'react-router-dom';
...
-  return redirect('/');
-};
+export const loader = logout;
 
 // We log out in a "resource route" that we can hit from a fetcher.Form
-export const action = async () => {
+export const action = logout;
+
+async function logout() {
+  localStorage.clear();
```
jimniels, line 12: Added this here, so when a user logs out it clears localstorage so they start afresh

quadratic-client/src/routes/team.$.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Shortcut route to get to a route for whatever the 'active' team is, e.g.
+ *
+ * /team/settings -> /teams/:teamUuid/settings
+ */
+export const loader = async (loaderArgs: LoaderFunctionArgs) => {
+  const { activeTeamUuid } = await requireAuth(loaderArgs);
+  const { params } = loaderArgs;
+  return redirect(ROUTES.TEAM(activeTeamUuid) + '/' + params['*']);
+};
```
jimniels, line 14: Similar to others, this used to be defined in the router file, but moved it here so it has its own file like other routes.

quadratic-client/src/routes/teams.$teamUuid.files.create.tsx
```diff
@@ -15,7 +15,13 @@ const getFailUrl = (path: string = '/') => {
   return path + '?' + params.toString();
 };
 
-export const loader = async ({ request, params }: LoaderFunctionArgs) => {
+export const shouldRevalidate = () => false;
```
jimniels, line 18: Similar to others, was defined in the router file previously but everything else about it is defined in this file. So we define it here too.

quadratic-client/src/shared/utils/activeTeam.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // Otherwise, they’re _likely_ a new user
+    // (they could be a user who was removed from a team and now has 0 teams)
+    // In either case, we'll create a new team for them automatically and use it.
+    const newTeam = await apiClient.teams.create({ name: 'My Team' });
+    const newTeamUuid = newTeam.uuid;
+    localStorage.setItem(KEY, newTeamUuid);
+    // (because we use this in loaders, we can throw a redirect)
+    throw isMobile ? redirect(ROUTES.TEAM(newTeamUuid)) : redirect(ROUTES.CREATE_FILE(newTeamUuid));
```
jimniels, line 81: It's worth noting that this piece of logic for handling new users has been in the app for a while and it will override any particular action they take from outside the app.

So, for example, if a user comes from the marketing site via prompt (`/files/create?prompt...`) that won't work if they go through the signup flow because this will pick up that they're a new user and override it.

We might need some more thought on supporting that case — but I haven't touched it yet in this PR as it's a problem that exists in prod.

jimniels: cc @davidkircos worth mentioning to you. I'm not sure if we've tested this, but if you submit a prompt from the marketing site and then signup, you won't get taken to a new file with that prompt. you'll just end up in a new file (if you're not on mobile).

To have special handling for new users, we'd have to also sniff out whether they're coming to a specific URL (and not just the root `/`) and then allow that action to happen (e.g. coming to `/files/create?prompt=...` — I'm not sure if there are others...maybe new users coming for the education enrollment check?)

quadratic-client/src/router.tsx
```diff
@@ -1,115 +1,117 @@
...
-            <Route index element={<Navigate to="/" replace />} />
-            <Route path=":uuid/history" lazy={() => import('./routes/file.$uuid.history')} />
-            <Route
-              path=":uuid"
-              id={ROUTE_LOADER_IDS.FILE}
-              lazy={() => import('./routes/file.$uuid')}
-              shouldRevalidate={({ currentParams, nextParams }) => currentParams.uuid !== nextParams.uuid}
-            />
-          </Route>
+        {/**
```
jimniels, line 22: Note: a large part of this file has been refactored to accommodate [the case outlined in this comment](https://github.com/quadratichq/quadratic/pull/2281#discussion_r2029436800)

It's mostly shuffling routes around and instead of having one wrapper route for "protected routes"

```
<Route loader={protectionFn}>
  <Route path="teams/:uuid/connections" />
  <Route path="teams/:uuid/members" />
</Route>
```

That doesn't work like you think it does. It will still run the loaders for any deeply-matched routes, which means they make API calls before the redirect to login happens.

Now all route protections happen _with every route that needs to be protected_. Essentially this:

```
<Route path="teams/:uuid/connections" loader={protectionFn} />
<Route path="teams/:uuid/members" loader={protectionFn} />
```

jimniels: Also added a bunch of comments in this file to make it easier to digest what's happening with all the routes in the app and how they're structured.

quadratic-client/src/app/ui/menus/TopBar/TopBarMenus/FileMenubarMenu.tsx
```diff
@@ -85,12 +84,14 @@ export const FileMenubarMenu = () => {
...
             {createNewFileAction.label}
+            <ExternalLinkIcon className="ml-auto !h-4 !w-4 text-center !text-xs text-muted-foreground opacity-50" />
           </MenubarItem>
         )}
         {duplicateFileAction.isAvailable(isAvailableArgs) && (
-          <MenubarItem onClick={() => duplicateFileAction.run({ fileUuid, submit })}>
+          <MenubarItem onClick={() => duplicateFileAction.run({ fileUuid })}>
             <FileCopyIcon />
             Duplicate
+            <ExternalLinkIcon className="ml-auto !h-4 !w-4 text-center !text-xs text-muted-foreground opacity-50" />
```
jimniels, line 94: File -> New / Duplicate will open in a new tab

CleanShot 2025-04-07 at 13 43 19@2x

quadratic-api/src/middleware/getTeamConnection.test.ts
```diff
@@ -64,12 +84,33 @@ describe('getConnection()', () => {
...
+          teamUuid: '11111111-1111-1111-1111-111111111111',
+          userId,
+        });
+      } catch (error) {
+        const e = error as ApiError;
+        expect(e).toBeInstanceOf(ApiError);
+        expect(e.status).toBe(404);
+      }
+    });
+  });
```
jimniels, line 115: This is the new test that will specifically check as to whether a connection that's being requested is accessible by the user _in the given team_.

quadratic-api/src/middleware/getTeamConnection.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  // Make sure the user has access to the connection _via the specified team_
+  const team = await getTeam({ uuid: teamUuid, userId });
+  if (!team.userMakingRequest.permissions.includes('TEAM_VIEW')) {
+    throw new ApiError(403, 'You don’t have access to this connection');
+  }
+
+  return { connection, team };
+}
```
jimniels, line 54: This file was renamed from `getConnection` to `getTeamConnection` to make clear we're getting a connection specifically in the context of a team.

luke-quadratic:
Nits but only things I could find: 

- Menu overlaps editor 
CleanShot 2025-04-10 at 09 46 14@2x
- (Idk how related to this PR this is) If you open a view-only file that has a connection and open the editor for that connection it takes you to auth - you see the editor opened for a split second and then it shoots you to auth 
Give a shout if you need help recreating this

jimniels:
Both issues [noted by Luke](https://github.com/quadratichq/quadratic/pull/2281#issuecomment-2794324515) have been fixed in the latest build.

quadratic-api/src/middleware/getTeamConnection.test.ts
AyushAgrawal-A2: I don't understand why this test will fail, also this try catch here is wrong, if the await in try never throws an error test will still pass

AyushAgrawal-A2: ^ for this `connectionUuid` and `teamUuid` pair, this test should pass, which it does by never throwing an error, so the `expect` does not trigger at all

quadratic-client/src/app/web-workers/quadraticCore/worker/coreConnection.ts
```diff
@@ -14,7 +14,8 @@ declare var self: WorkerGlobalScope &
       sheetId: string,
       code: string,
       connector_type: ConnectionKind,
-      connection_id: String
+      connection_id: String,
+      teamUuid: string
```
AyushAgrawal-A2, line 18: does not match the params of `this.sendConnection` or `rustCallback.sendConnection`

quadratic-api/src/middleware/getTeamConnection.test.ts
jimniels: Great catch. I wrote this test poorly and it was giving me a false positive.

I updated the connection/team UUID pairs to be the proper combination of "this shouldn't work" and I updated the test to explicitly fail after the await. Now it should be as expected

jimniels:
@AyushAgrawal-A2 great catches, I've resolved both of those. Ready for 👀  again.

AyushAgrawal-A2 (approved):

luke-quadratic (approved):
Forgot to add approval when I went through this before - followed Jim's exhaustive list of testing

PR #2293, opened 2025-02-25
feat: add 3.7 sonnet

AyushAgrawal-A2:
Checklist:
- Old chat history should work, even with new models
- Hot swapping of models should work in between chat


davidkircos (approved):
lgtm, let's merge into grok to :ship: tomrrow

PR #2300, opened 2025-02-26
Hide hoverCell when the cell has the inline editor open

davidfig:
Replicate: try to edit a formula with an error. You should not be able to hover over where the red triangle was and see the error information


quadratic-client/src/app/gridGL/HTMLGrid/hoverCell/HoverCell.tsx
```diff
@@ -142,6 +143,14 @@ export function HoverCell() {
 
   useEffect(() => {
     const addCell = (cell?: JsRenderCodeCell | EditingCell | ErrorValidation) => {
+      // don't show hover cell if the inline editor is showing at the same location
```
ddimaria, line 146: Great comment, will keep anyone from casually deleting.

ddimaria (approved):

PR #2303, opened 2025-02-26
Fix rename mouse

davidfig:
When renaming table or column name, you can use the mouse to edit.


quadratic-client/src/app/gridGL/HTMLGrid/contextMenus/Base.tsx
```diff
@@ -45,6 +45,7 @@ export const ContextMenuBase = ({ children }: { children: React.ReactNode }) =>
     };
 
     const handleMouseDown = (event: MouseEvent) => {
+      if (event.target instanceof HTMLInputElement) return;
```
ddimaria, line 48: A comment might be helpful here

ddimaria (approved):

davidkircos (approved):
functionality lgtm!

PR #2307, opened 2025-02-27
fix: correct context menu for JS charts

jimniels:
Fixes https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1740617319494609


davidkircos (approved):
good fix

PR #2308, opened 2025-02-27
fix: get cells

AyushAgrawal-A2:


jimniels (approved):
I can't speak to the code, but this is definitely an improvement on how the sheet was functioning for me with a particular sheet. So I'll 👍  the functionality.

quadratic-client/src/app/web-workers/pythonWebWorker/worker/pythonCore.ts
davidfig, line 1: How are you handling errors in getCells if you don't send the lineNumber?

davidfig (approved):

quadratic-client/src/app/web-workers/pythonWebWorker/worker/pythonCore.ts
davidfig: Rust will cancel the execution and record the error if it fails at getCells

AyushAgrawal-A2: it will be handled by the py / js worker itself, as it does for any other error apart from getCells ones

AyushAgrawal-A2: > Rust will cancel the execution and record the error if it fails at getCells
this was the bug, it leaves the py / js worker hanging and the worker later responds back with another result

getCells does not short circuit now, I reponds to the py / js worker only, does not change the execution flow

davidfig (approved):

PR #2340, opened 2025-02-27
feat: move theme options out of labs

### LINKED ISSUE #2283: Take Themes/Dark mode out of Labs
Last issue outstanding - #2105

jimniels:
Resolves #2283 

CleanShot 2025-02-27 at 15 38 50@2x

CleanShot 2025-02-27 at 14 57 44@2x

---

Notes on implementation:

1. Theme options no longer live in "Labs" but are now under "Profile & preferences".
    1. We can consider moving this to a separate place in the future, like making "Profile" and "Preferences" two different places. Just doesn't feel like we have enough of a reason to make that distinction today? So was just planning on keeping them combined for now.
2. Theme options remain in the same spot when you're in the spreadsheet.
3. "Labs" is no longer in the sidebar because there are 0 things in the lab (so it disappears from the menu)
4. Users who never turned this on in labs will have theming turned on automatically. Defaults are blue/light.
5. Moved the css invert style to individual elements that need it, rather than the entire grid wrapper (as this was causing visual bugs where things were white when they should be black)


AyushAgrawal-A2 (changes requested):
can you please fix the drag/drop modal dark mode

[Screencast from 2025-02-28 04-48-33.webm](https://github.com/user-attachments/assets/cab1c42f-a188-45fe-bfcb-e0c91e41d907)

jimniels:
Nice, good catch @AyushAgrawal-A2  — it's a straightforward fix, swapped some hard-coded colors to theme variables. Screenshots:

CleanShot 2025-02-27 at 16 26 07@2x

CleanShot 2025-02-27 at 16 26 14@2x

AyushAgrawal-A2 (approved):

quadratic-client/src/app/gridGL/HTMLGrid/HTMLGridContainer.tsx
```diff
@@ -118,10 +121,10 @@ export const HTMLGridContainer = (props: Props): ReactNode | null => {
       </div>
 
       <GridFileInput />
-
       <Following />
-
+      <ImportProgress />
       <EmptyGridMessage />
+      <Search />
```
jimniels, line 127: Moved these just to try and put all these elements in the same place. They were in QuadraticGrid before, but ultimately render in the same place. So I'm sticking them here where everything else is.

quadratic-client/src/app/gridGL/HTMLGrid/SuggestionDropdown.tsx
```diff
@@ -148,7 +148,7 @@ export const SuggestionDropDown = () => {
   return (
     <div
       className={cn(
-        'border.gray-300 pointer-events-auto absolute cursor-pointer overflow-y-auto border bg-white text-gray-500',
+        'pointer-events-auto absolute cursor-pointer overflow-y-auto rounded-sm border border-border bg-background text-muted-foreground',
```
jimniels, line 151: CleanShot 2025-02-28 at 10 30 05@2x

Using theme colors here rather than hard-coded colors.

quadratic-client/src/app/gridGL/QuadraticGrid.tsx
```diff
@@ -37,7 +35,6 @@ export default function QuadraticGrid() {
   return (
     <div
       ref={containerRef}
-      className="dark-mode-hack"
```
jimniels, line 40: Moved this from the parent because it was inverting _everything_ under it. This cause HTML-rendered elements like the empty sheet message to invert to white in dark mode:

CleanShot 2025-02-28 at 12 40 25@2x

Instead, we now apply `dark-mode-hack` to any individual element that should be inverted. Hopefully over time we can strip these out one-by-one and apply proper theme colors.

quadratic-client/src/app/ui/components/ImportProgress.tsx
```diff
@@ -14,22 +14,21 @@ export const ImportProgress = () => {
   const abortController = files[currentFileIndex].abortController;
 
   return (
-    <div className="fixed bottom-16 left-8 z-50 h-[92px] w-[403px] select-none border border-slate-200 bg-white pb-2 pl-4 pr-4 pt-2 tracking-tight shadow-[0_2px_5px_0px_rgba(0,0,0,0.15)]">
-      <div className="flex justify-between">
+    <div className="absolute bottom-4 left-4 z-50 w-96 select-none rounded border border-border bg-background pb-2 pl-4 pr-4 pt-2 tracking-tight shadow-lg">
+      <div className="mb-2 flex items-center justify-between">
```
jimniels, line 18: This element had a bunch of hard-coded values. Causing it to not work well in dark mode.

CleanShot 2025-02-28 at 12 46 44@2x

Changed all them to be theme colors.

davidkircos (changes requested):
Context menu is disconnected from right click position

davidkircos (approved):
lgtm

PR #2349, opened 2025-02-28
Allow mouse selection when scrolling the viewport at the edges

davidfig:
https://github.com/user-attachments/assets/cf08cc42-2d5b-4d47-ad78-7bc6ecc0282c


davidkircos (approved):
awesome! this also makes it possible to scroll with trackpad while selecting with a second mouse (which I do all the time). Great!

davidkircos (changes requested):
When making a selection the selected area updates much more slowly then it used to, even without scrolling

https://github.com/user-attachments/assets/de459bf0-2c61-41af-b309-920f5dce72e4

davidfig:
Will revisit with a different solution.

davidkircos (approved):
functionality works great for me

AyushAgrawal-A2 (approved):

PR #2354, opened 2025-02-28
Rework code/formula translation & name-updating

HactarCE:
This PR overhauls sheet/table/column renaming and coordinate translation in formulas and code cells.

## Intended behavior

These actions should preserve all cell references:
- Cutting & pasting code cells

These actions should update only **relative** references in code & formulas to **all** sheets:
- Copying & pasting code cells

These actions should update **all** references in code & formulas to the sheet where they occur:
- Inserting/deleting columns
- Renaming a sheet
- Renaming a table
- Renaming a column within a table

References should be updated even in cells that are not in the current sheet. References (including invalid ones) should remain unmodified.

Let me know via comment here or on Slack if I've made any mistakes in the intended behavior.

## Code review

For code review, check that there are no extraneous calls to `Grid::a1_context()`.


quadratic-core/src/a1/a1_context/mod.rs
```diff
@@ -154,6 +154,13 @@ impl A1Context {
     pub fn table_mut(&mut self, table_name: &str) -> Option<&mut TableMapEntry> {
         self.table_map.get_mut(table_name)
     }
+
+    /// Constructs an A1 context with only a single sheet.
+    pub fn with_single_sheet(sheet_name: &str, sheet_id: SheetId) -> Self {
+        let mut ret = Self::default();
```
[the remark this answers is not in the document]
HactarCE: "return value"

I avoid "res" or "result" because `Result` has special meaning in rust

I'm happy to change this variable if you have another name in mind

quadratic-core/src/a1/sheet_cell_ref_range.rs
```diff
@@ -3,47 +3,73 @@
...
+    /// Returns whether the reference needs a sheet name in order to be unambiguous.
+    fn needs_sheet_name(&self, default_sheet_id: Option<SheetId>) -> bool {
+        self.explicit_sheet_name
+            || match self.cells {
+                CellRefRange::Sheet { .. } => {
+                    default_sheet_id.is_none_or(|sheet_id| sheet_id != self.sheet_id)
+                }
+                CellRefRange::Table { .. } => false, // table names are unique per file
+            }
+    }
```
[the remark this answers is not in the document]
HactarCE: It can't read. If `explicit_sheet_name` is true then the function returns true.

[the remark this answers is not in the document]
ddimaria: @HactarCE what do you think ☝️?

quadratic-core/src/a1/sheet_cell_ref_range.rs
```diff
@@ -76,6 +101,44 @@ impl SheetCellRefRange {
...
+                sheet_id: self.sheet_id,
+                cells: self.cells.saturating_adjust(adjust)?,
+                explicit_sheet_name: self.explicit_sheet_name,
+            })
+        } else {
+            Some(self)
+        }
+    }
+
+    pub fn replace_table_name(&mut self, old_name: &str, new_name: &str) {
```
ddimaria, line 134: nit: doc comments

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -465,7 +476,9 @@ impl GridController {
 
         let mut cursor = clipboard
             .selection
-            .translate(cursor_translate_x, cursor_translate_y)?;
+            .clone()
```
ddimaria, line 479: Is this `clone()` necessary?

quadratic-core/src/grid/code_cell.rs
```diff
@@ -64,99 +69,221 @@ impl CodeCellValue {
...
         &mut self,
-        old_name: &str,
-        new_name: &str,
-        default_sheet_id: &SheetId,
-        a1_context: &A1Context,
+        old_a1_context: &A1Context,
+        new_a1_context: &A1Context,
+        pos: SheetPos,
     ) {
-        if old_name != new_name && self.is_code_cell() {
```
ddimaria, line 113: Was the removal of `if old_name != new_name` intentional?

quadratic-core/src/grid/code_cell.rs
```diff
@@ -64,99 +69,221 @@ impl CodeCellValue {
...
+                self.replace_q_cells_a1_selection(pos, a1_context, |mut cell_ref| {
+                    cell_ref.replace_column_name(table_name, old_name, new_name);
+                    Ok(cell_ref.to_a1_string(Some(pos.sheet_id), a1_context))
+                });
+            }
+        }
+    }
+}
+
+/// Adjustment to make to the coordinates of cell references in code cells.
```
ddimaria, line 170: Great docs 💯

quadratic-core/src/grid/file/shift_negative_offsets.rs
```diff
@@ -94,9 +94,13 @@ pub fn shift_negative_offsets(grid: &mut Grid) -> HashMap<String, (i64, i64)> {
                 }
 
                 // Translate all ranges and collect into new HashSet
-                *ranges = ranges
-                    .iter()
-                    .filter_map(|r| r.translate(x_shift, y_shift).ok())
+                *ranges = std::mem::take(ranges)
+                    .into_iter()
+                    .filter_map(|r| {
+                        // This is actually wrong -- we should translate differently
```
ddimaria, line 100: @AyushAgrawal-A2

quadratic-core/src/util.rs
```diff
@@ -102,13 +102,34 @@ macro_rules! row {
...
+/// # Examples
+///
+/// ```
+/// # use quadratic_core::{pos, Pos, grid::SheetId};
+/// assert_eq!(pos![A1], Pos::new(1, 1));
+/// assert_eq!(pos![C418], Pos::new(3, 418));
+///
+/// // With a sheet ID (identifier)
+/// let my_sheet = SheetId::new();
+/// assert_eq!(pos![my_sheet!A1], Pos::new(1, 1).to_sheet_pos(my_sheet));
```
ddimaria, line 118: 💯

ddimaria (approved):
@AyushAgrawal-A2 just one comment to address before merging

quadratic-core/src/grid/file/shift_negative_offsets.rs
HactarCE: I talked with Ayush; I was mistaken and the existing code (and the code that is currently in this PR) is correct

quadratic-core/src/grid/code_cell.rs
HactarCE: Removed because:
- we no longer have access to `old_name` and `new_name` (since we create the `A1Context` objects much earlier)
- if `old_name` and `new_name` are the same, then it's a waste of time to loop over all the code cells anyway
I'll add the check at a much higher level, which is where it should be IMO.

quadratic-core/src/a1/sheet_cell_ref_range.rs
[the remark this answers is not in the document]
HactarCE: I have taken the suggestion for readability's sake even though the logic is the same

quadratic-core/src/controller/operations/clipboard.rs
HactarCE: `adjust()`/`translate()`/etc. take owned values because usually that avoids extra allocation. Here is a rare case where it might require an extra allocation, but it's just once per copy so it'll have minimal perf impact.

PR #2357, opened 2025-02-28
fix: add csp header to allow iframe embed

AyushAgrawal-A2:


davidkircos (approved):

davidkircos:
The only way to test is production

Issue #2244, opened 2025-01-27
Ensure Visible off on wide cells


davidkircos:
- [x] When dragging to select large cells where the far right side is off screen, each cell you select moves the view port and it's impossible to select what you want.
- [x] When reading a list of cells where the right edge is off screen, everytime you press down the viewport jumps to the bottom right of the selected cell, making it so you have to scroll back left on every cell.

We should really only be calling ensure visible when the cursor is fully off screen and when the user is using the keyboard.

If a user is selecting cells with the mouse, we should not move the viewport 

(To successfully close this issue, the problems with users getting jumped around in the sheet should be fully solved. All navigating should work as expected; this bug is by far the biggest culprit but there might be other scenarios that are odd behavior)

PR #2361, opened 2025-02-28
Fix table selection

davidfig:
- [x] do not automatically use table ref
- [x] allow table ref when user clicks on table name or column (only)
- [x] allow extension of table ref by clicking and dragging from table name and columns elsewhere
- [x] allow table ref when using keyboard to move into the table name or column name cell (without shift)
- [x] ensureVisible is not called when using mouse to select cells
- [x] properly handle case where cell width/height is larger than screen size (default to the left side of the cell) 
- [x] column names can properly be selected for copy/paste


davidkircos (commented):
This functionality feels much more intuitive. Want to spend some more time on it, and get feedback from other team members. 

Core tests are failing

davidfig:
This also fixes: #2244--the fix makes it work comparable to Excel's wide cells.

davidkircos (approved):
this feels like a huge improvement in table selections
i haven't found any new issues

PR #2362, opened 2025-02-28
Fix floating column headers height

davidfig:
Fixes this bug:

https://github.com/user-attachments/assets/744c2f3e-af65-46dc-ae3b-87a3df643327


ddimaria (approved):

AyushAgrawal-A2 (changes requested):
Column header does not resize when show_name is false

[Screencast from 2025-03-01 03-20-24.webm](https://github.com/user-attachments/assets/aef440a9-e259-4883-8fb9-4e9374b4fb6e)

AyushAgrawal-A2 (approved):

davidkircos (approved):
functionality looks great to me!

Issue #2320, opened 2025-02-27
Marching ants inside of tables is drawn poorly


luke-quadratic:
Two issues with marching ants in tables: 
1. When you copy an entire table the marching ants only get applied to the data, not the entire table
2. General drawing of the ants isn't perfect to the selection

PR #2364, opened 2025-02-28
Copy a table shows marching ants around the full table

ddimaria:
Addresses https://github.com/quadratichq/quadratic/issues/2320

marching-ants


AyushAgrawal-A2 (approved):

davidkircos (changes requested):
The ants have a different margin on the left, bottom, and right sides. I think they should all have the same margin as the right one, really close to the edge

CleanShot 2025-03-01 at 09 25 47@2x

davidkircos (approved):
lgtm! Nice

PR #2365, opened 2025-02-28
chore: remove mui avatars

jimniels:
This changes the avatars to use shadcn colors as well as removes usage of the MUI `<Avatar>` and `<AvatarGroup>` on the team page. Instead we rely on our own internal `<Avatar>` component which is now sensitive to light/dark mode.

CleanShot 2025-02-28 at 16 34 22@2x

CleanShot 2025-02-28 at 16 34 10@2x

Part of an on-going effort to remove MUI components and standardize on shadcn.

### DIFF
--- a/quadratic-client/src/shared/components/Avatar.tsx
+++ b/quadratic-client/src/shared/components/Avatar.tsx
@@ -1,4 +1,5 @@
 import { getAuth0AvatarSrc } from '@/app/helpers/links';
+import { cn } from '@/shared/shadcn/utils';
 import type { ImgHTMLAttributes } from 'react';
 import React, { forwardRef } from 'react';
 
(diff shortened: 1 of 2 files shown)


quadratic-client/src/shared/components/Avatar.tsx
```diff
@@ -7,41 +8,46 @@ interface AvatarProps extends ImgHTMLAttributes<HTMLImageElement> {
...
-  };
+    const stylePreset = {
+      width: size === 'small' ? '24px' : size === 'medium' ? '32px' : size === 'large' ? '40px' : '24px',
+      height: size === 'small' ? '24px' : size === 'medium' ? '32px' : size === 'large' ? '40px' : '24px',
+      fontSize: size === 'small' ? '0.75rem' : size === 'medium' ? '1rem' : size === 'large' ? '1.125rem' : '0.8125rem',
+      borderRadius: '50%',
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+    };
```
jimniels, line 23: A lot of reformatting/indentation going on in this file. Primary changes are removing hard-coded `color` and `backgroundColor` values and using shadcn token values. Also now we use token colors in `className` on the <span>` below.

AyushAgrawal-A2 (approved):

davidkircos (approved):
nice, lgtm

PR #2381, opened 2025-03-03
Move column and row

### LINKED ISSUE #1072: Move columns/rows
(Scoped but needs discussion before building to make sure we all agree this is the correct approach)

(Should also include moving table columns/rows)

### Single column/row move 
- Cuts/pastes all formatting
- Can be inserted left/right or above/below any column/row respectively
- Formulas/code do not shift references relatively, unless the data they are referencing gets shifted 
      (e.g. SUM(C9:C12), if C9:12 get moved to E9:E12, then it becomes SUM(E9:E12)); but if a column
      got inserted to the left of the SUM formula but the data location stayed the same, reference
      shouldn’t change
- Columns/rows that aren’t being moved will shift location only when the element being shifted
      comes from other side (see GIF, hard to explain)

Image

### Multi-column/row move 
- Cuts/pastes all formatting 
- Contiguous columns highlighted are default moved together, non-contiguous columns do not get moved 
- Can be inserted left/right or above/below any column/row respectively
- Formulas/code do not shift references relatively, unless the data they are referencing gets shifted 
      (e.g. SUM(C9:C12), if C9:12 get moved to E9:E12, then it becomes SUM(E9:E12)); but if a column
      got inserted to the left of the SUM formula but the data location stayed the same, reference
      shouldn’t change
- Columns/rows that aren’t being moved will shift location only when the element being shifted
      comes from other side (see GIF, hard to explain)

Image

davidfig:
Closes #1072

- [x] move column(s) and row(s) by deleting them and inserting them - dragging happens from the heading bar
- [x] move column(s) and row(s) by cutting them (but leaving the columns/rows) - dragging happens from the sides of the selection box

Note: move column/row within data table will be done in a separate PR

https://github.com/user-attachments/assets/423bcf06-3bb7-4f82-88cb-db2984cffd5a


luke-quadratic (approved):
This feels great. 

I tested: tons of combinations of moving data, code, tables, etc around the sheet; behaved as expected.

luke-quadratic:
Note from David F: 
"Delete columns and rows were slightly tweaked. Also, undo and multiplayer should be tested."

AyushAgrawal-A2 (approved):

davidfig:
Added functionailty behind the debug flag. Won't work without &debug in the url

davidkircos (approved):

PR #2389, opened 2025-03-03
Hide table dropdown icon when not active

davidfig:


davidkircos (approved):
nice, this looks great!

PR #2397, opened 2025-03-04
AI shouldn't try to explain the data or code its generating in the chat 

luke-quadratic:
AI shouldn't try to explain the data or code its generating in the chat. 

When the AI tries to repeat back the data or code it's generating it's often incorrect. We've gotten a lot of user reports where people have been confused by the discrepancy. 

**Prompts for testing (these prompts try to say the values nearly 100% of the time in prod)** 
"Create a sample product analytics dataset with monthly active users, conversion rate, revenue, session duration, and churn rate for 2024, then create a dual-axis chart showing the relationship between active users and revenue growth."
"Create some sample sales data and create a chart plotting it."


davidkircos (approved):
makes sense to me

PR #2416, opened 2025-03-05
Programmatically set sheet name

davidfig:
This changes the default sheet name to `Sheet 1`. We can revert this change by changing a const and rerunning the `build:file:blank` script.


luke-quadratic:
What should we test? Unclear to me what to be looking out for or what this changes other than "is the default sheet 'Sheet 1'"

AyushAgrawal-A2 (approved):

PR #2424, opened 2025-03-05
fix: font loading

jimniels:
## Problem

When the icon font loads slowly, sometimes you see the element's innerText

image

## Solution

Hide the icons by default. Then, once the font loads via Google’s CDN, reveal them by removing a global document attribute.

## To Test

Use network request blocking to simulate the font not loading (by blocking `fonts.gstatic.com`)

CleanShot 2025-03-05 at 13 44 15@2x


quadratic-client/src/app/ui/components/InitialPageLoadError.tsx
```diff
@@ -1,19 +0,0 @@
...
-  };
-
-  return (
-    <QuadraticSnackBar open={open} onClose={onClose} autoHideDuration={8000}>
-      <Alert severity="error" onClose={onClose} elevation={6}>
-        Failed to load the file specified in URL.
-      </Alert>
-    </QuadraticSnackBar>
-  );
-}
```
jimniels, line 19: Deleted this file because found during the course of this PR that this file isn't being used anywhere.

PR #2427, opened 2025-03-05
Combine data table operations in set cell values

### LINKED ISSUE #2437: AI hidden at too high of window width
AI gets hidden at way too wide of browser width. 

Once browser is too small and you can't open AI when clicking AI we should pop something warning that the screen width needs to be larger to open AI.

ddimaria:
Closes https://github.com/quadratichq/quadratic/issues/2327
Closes #2437 

- [x] When adding one or more cell values to the immediate right or bottom, expand the table
- [x] Add tests
- [x] Housekeeping/refactors 

ai-insert-multiple-rows-and-cols


quadratic-core/src/values/cellvalue.rs
```diff
@@ -556,6 +562,77 @@ impl CellValue {
         }
     }
 
+    /// Convert string to a cell_value and generate necessary operations
```
ddimaria, line 565: I didn't write this function in this PR, I just moved it to a more appropriate place.

luke-quadratic:
I got into a state that I can't recreate in prod so I assume it was introduced in this branch. 

Basically when the AI took an action to replace a python code table with a data table it deleted the data but a lot was left over but can't be deleted. (my button press isn't getting registered but in this video I'm pressing delete and nothing is happening) 

CleanShot 2025-03-10 at 17 00 45

not sure if relevant but this existed in console 
CleanShot 2025-03-10 at 17 02 16@2x

ddimaria:
@luke-quadratic I removed the "Requested Changes" label since this issue isn't related to this PR.  @AyushAgrawal-A2 is resolving in a separate PR.

luke-quadratic:
> @luke-quadratic I removed the "Requested Changes" label since this issue isn't related to this PR. @AyushAgrawal-A2 is resolving in a separate PR.

Sounds good, thanks. Re-reviewing.

luke-quadratic:
The undo operation here should both delete the value and reduce the table size. These shouldn't be separate transactions.
CleanShot 2025-03-11 at 09 43 06

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -1,165 +1,179 @@
...
-                        x: sheet_pos.x + x as i64,
-                        y: sheet_pos.y + y as i64,
-                    };
+                    let pos = Pos::new(sheet_pos.x + x as i64, sheet_pos.y + y as i64);
+                    let current_sheet_pos = SheetPos::from((pos, sheet_pos.sheet_id));
 
                     let is_code = matches!(cell_value, CellValue::Code(_));
                     let is_data_table =
                         matches!(cell_value, CellValue::Code(_) | CellValue::Import(_))
                             || sheet.has_table_content(pos, true);
```
AyushAgrawal-A2, line 72: `sheet.has_table_content` creates and checks output_rects of all data tables, we can use the vec of data table bounds here as well

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -1,165 +1,179 @@
...
 
-                    match is_data_table {
-                        true => {
-                            ops.extend(self.set_data_table_operations_at(sheet_pos, value));
-                        }
-                        false => {
+                    // (x,y) is within a data table
+                    if is_data_table {
+                        let data_table_pos = sheet.first_data_table_within(pos)?;
```
AyushAgrawal-A2, line 76: `first_data_table_within` same as `sheet.has_table_content`, creates and check output rects of all tables, we can combine these two

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -1,165 +1,179 @@
...
-                            // also re-enable the test_expand_data_table_column_row_on_setting_value test
-                            // data_table_ops.extend(self.set_cell_value_data_table_operations(
-                            //     SheetPos::from((pos, sheet_pos.sheet_id)),
-                            //     value,
-                            // ));
+                        } else {
+                            data_table_cell_values[x][y] = Some(cell_value);
+
+                            if !format_update.is_default() {
+                                ops.push(Operation::DataTableFormats {
```
AyushAgrawal-A2, line 87: this will create one format op for each value that goes into table, we can combine these into one op for all values for one table

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -1,165 +1,179 @@
...
+                                    .for_each(|rect| {
+                                        rect.max.y += 1;
+                                    });
+                            }
+                        }
+                    }
 
                     if !format_update.is_default() {
                         sheet_format_updates.set_format_cell(pos, format_update);
                     }
```
AyushAgrawal-A2, line 160: `format_update` should not apply to sheet, if value is part of data_table

AyushAgrawal-A2 (changes requested):

quadratic-core/src/controller/operations/cell_value.rs
AyushAgrawal-A2: done

quadratic-core/src/controller/operations/cell_value.rs
AyushAgrawal-A2: done

quadratic-core/src/controller/operations/cell_value.rs
AyushAgrawal-A2: will finish this later, when handling multiple table overlap with input values vec

AyushAgrawal-A2 (approved):

davidkircos (commented):
Nice, this works really well. Really improves AI <> Tables interactions 🙌

davidkircos (approved):

PR #2433, opened 2025-03-06
Loading fonts in pixi timeouts on slow connections

davidfig:
Fixes an issue when loading the app over a slow connection. Before, after 3s, the font would timeout, which would stop the app from loading.


jimniels (approved):

davidkircos (commented):
What happens if it fails to load within 10s?

PR #2442, opened 2025-03-06
fix: avatar spacing

jimniels:
The `className` prop wasn't being passed to the `<Avatar>` component, so context-specific UI overrides weren't being applied correctly. For example, grouped avatars weren't spaced/overlaid correctly.

Here's an example from prod:

CleanShot 2025-03-06 at 16 54 54@2x

And on local:

CleanShot 2025-03-06 at 16 52 58@2x

**The above are examples of what it SHOULD NOT look like.** Here's what it looks like in the current preview branch:
CleanShot 2025-03-07 at 16 01 45@2x

### DIFF
--- a/quadratic-client/src/shared/components/Avatar.tsx
+++ b/quadratic-client/src/shared/components/Avatar.tsx
@@ -41,6 +41,7 @@ export const Avatar = forwardRef<HTMLImageElement, AvatarProps>(
             crossOrigin="anonymous"
             onError={() => setError(true)}
             style={{ ...stylePreset, ...style }}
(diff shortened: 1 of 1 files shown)


davidkircos:
This doesn't look constant in the spacing? What am I missing?

jimniels:
@davidkircos the screenshots on the PR are the _buggy_ version. You'll have to view the preview branch to see that it works as expected.

I invited you. Right now, it looks like this (which is expected) with just two people:

CleanShot 2025-03-07 at 15 59 31@2x

PR #2447, opened 2025-03-07
Fixes column placement when toggling name

### LINKED ISSUE #2434: bug: table column headers selection is off by one when table name is hidden
- Toggle table header to be off
- Clicking the column headers doesn't select the column. Instead the 2nd row in the table is now the column header.

https://github.com/user-attachments/assets/f306d657-5a6d-4a48-8939-1a5fb415f20d

davidfig:
Fixes #2434


luke-quadratic:
Can no longer recreate the issue this PR will close 👍

PR #2460, opened 2025-03-08
Fix code peek

### LINKED ISSUE #2419: Remove code peek
Code peek is currently not working on any code cells when enabled. 

We should remove this feature and revisit code peek in the context of Formula bars.

davidfig:
Fixed #2419 

I decided to re-enable code peek instead of removing it. We can remove it when we have formula bar.

- [x] re-enable code peek
- [x] don't show code peek for charts (as it gets in the way) 
- [x] add table name to the output

 
CleanShot 2025-03-08 at 08 36 00@2x

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2690,8 +2690,8 @@ version = "0.1.0"
 dependencies = [
  "chrono",
  "proc-macro2 0.4.30",
- "quote 0.6.13",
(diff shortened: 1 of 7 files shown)


luke-quadratic:
Works as expected (tested code peek on Python and Formulas)

PR #2462, opened 2025-03-09
Ensure table is not highlighted after copy as PNG on table column

### LINKED ISSUE #2404: bug: copy as PNG for table column leaves table in highlighted state
1. Select a column in a table
2. Right click, Copy as PNG
3. Note that the table becomes blue as if its highlighted

https://github.com/user-attachments/assets/a98c7024-849f-49da-a956-82637530c6ee

davidfig:
Fixes #2404


ddimaria (approved):

luke-quadratic:
Fix works 👍

Issue #2471, opened 2025-03-11
[BUG] Display table name at correct times


luke-quadratic:
## Describe the bug
Not much of a bug but rather an edit to our mistake of choosing when to show table name in code outputs. 

## Expected behavior
Table name should show in the following Python situations: 
- DataFrame 

Table name should not show by default in every other situation

To be discussed
- Series

PR #2480, opened 2025-03-12
Single cell code output has no table name

ddimaria:
## Relevant issue(s)
https://github.com/quadratichq/quadratic/issues/2471

## Description
- [x] code cell w/1-cell output automatically has show_name and show_columns = false
- [x] context menu for code cell w/1-cell output does not have options to toggle show_name or show_columns

### Some of the use cases we’re trying to solve for:
* Initial code run with a single value output (don’t show name)
* Initial code run with a non-single value output (show name)
* Code re-run of a single value output after initial code run with a single value output (don’t show name)
* Code re-run of a non-single value output after initial code run with a single value output (show name)
* Code re-run of a single value output after initial code run with a non-single value output (don’t show name)
* Code re-run of a non-single value output after initial code run with a non-single value output (show name)

### Initial single value
no-name-initial-single-value

### Initial non-single value
no-name-initial-non-single-value

### Initial non-single value, rerun to a single value
no-name-initial-non-single-value-rerun-single-value


luke-quadratic (approved):
Works as expected (tested single values, lists, dataframes, and charts) in Python

quadratic-core/src/controller/execution/run_code/mod.rs
```diff
@@ -103,6 +107,14 @@ impl GridController {
                 new_data_table.header_is_first_row |= old_data_table.header_is_first_row;
             }
 
+            // if the new data table is a single value and is a code cell, then we don't want to show the name
+            if new_data_table.is_single_value()
+                && is_code_cell
+                && !new_data_table.is_html_or_image()
```
AyushAgrawal-A2, line 113: `is_single_value` checks for `html_or_image`

`!new_data_table.is_html_or_image()` is redundant

quadratic-core/src/controller/execution/run_code/mod.rs
```diff
@@ -499,6 +513,11 @@ impl GridController {
         transaction.cells_accessed.clear();
         data_table.show_columns = js_code_result.has_headers;
 
+        // if the new data table is a single value and is a code cell, then we don't want to show the name
+        if is_single_value && language.is_code_language() && !data_table.is_html_or_image() {
```
AyushAgrawal-A2, line 517: we can use `data_table.is_single_value()` fn here as well and remove the redundant `!data_table.is_html_or_image()` check

AyushAgrawal-A2 (approved):
approved with 2 nits

PR #2486, opened 2025-03-12
remove legacy database tables

davidkircos:
## Description
The old QFile table takes a ton of space in our database, this cleans up unused tables and will help our database migration go faster.

We have backups of this data.


ddimaria (approved):
I checked my local database and these tables are empty, which validates we're not writing to them anymore.  The only remaining task is to backup this data in case it's needed in the future (which you may have already done).

davidkircos:
> The only remaining task is to backup this data in case it's needed in the future (which you may have already done).

Yup we have backed up the data

PR #2488, opened 2025-03-12
better capture sentry errors

### LINKED ISSUE #2487: File Error Screen should log special event to Sentry with console logs and context
Image

davidkircos:
## Relevant issue(s)
Closes #2487

### DIFF
--- a/quadratic-client/src/dashboard/components/Empty.tsx
+++ b/quadratic-client/src/dashboard/components/Empty.tsx
@@ -5,6 +5,8 @@ import { ROUTES } from '@/shared/constants/routes';
 import { Button } from '@/shared/shadcn/ui/button';
 import { cn } from '@/shared/shadcn/utils';
 import type { StopwatchIcon } from '@radix-ui/react-icons';
+import * as Sentry from '@sentry/react';
+import mixpanel from 'mixpanel-browser';
 import type { ReactNode } from 'react';
 import { useSubmit } from 'react-router-dom';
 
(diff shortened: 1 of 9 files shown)


ddimaria (approved):
Looks good.  Are we able to test in QA, or do we need this to be in prod?

quadratic-client/src/dashboard/components/Empty.tsx
```diff
@@ -28,6 +30,19 @@ export function Empty({
...
+      description,
+    });
+    Sentry.captureException(new Error('error-page'), {
+      extra: {
+        title,
+        description,
+      },
+    });
+  }
+
```
AyushAgrawal-A2, line 45: this will run on each re render of this page, which should be just one render, but still should be inside an `useEffect`

AyushAgrawal-A2 (changes requested):

quadratic-client/src/dashboard/components/Empty.tsx
ddimaria: Good point, Gemini agrees:
> If Sentry's capture methods (e.g., captureException, captureMessage) are used without proper handling, such as wrapping them in a useEffect hook with appropriate dependencies, they can cause infinite re-render loops. This is because these methods are side effects and should not be called on every render

davidkircos: nice catch

AyushAgrawal-A2 (approved):

Issue #2483, opened 2025-03-12
[BUG] Data missing after changing a DataFrame code cell to single output


luke-quadratic:
## Describe the bug
If you create a DataFrame and then attempt to change it to a single value output, the single value output will not appear in the sheet

![Image](https://github.com/user-attachments/assets/7bab7389-18da-4b8a-9b96-8a88c2eb14ef)
![Image](https://github.com/user-attachments/assets/8e9c0dfd-e626-40d5-aaae-e41176af0b1c)

## Expected behavior
Single value should appear. 

## Reproduction
1. Create a Python cell with a DataFrame
2. Run the cell 
3. Replace the code with a single output, e.g. '1' 
4. Press run and nothing will appear

PR #2491, opened 2025-03-12
Data missing when changing dataframe code cell to a single value 

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2483

Also fixes a bug in prod where changing from data frame to list caused headers in the lis.

## Description
In Python, you can now create code that outputs a data frame, then switch to code that outputs a single value (and back again) with data still displaying properly

### DataFrame output, then single value output, then DataFrame output
dataframe-to-single-value


luke-quadratic:
Note two different bugs that exist here still: 
- Going from DataFrame to multi-row output (list) causes first value to be a header
- Going from DataFrame to single value to list causes the first value of list to not appear (in QA this doesn't exist, but the first value does become a header for some reason)

CleanShot 2025-03-12 at 14 49 32

quadratic-client/src/app/gridGL/cells/tables/Table.ts
```diff
@@ -102,12 +102,17 @@ export class Table extends Container {
       console.log('Expected CellsMarkers to be defined in sheet');
       return;
     }
-    if (this.codeCell.state === 'RunError' || this.codeCell.state === 'SpillError') {
+    if (
+      this.codeCell.state === 'RunError' ||
+      this.codeCell.state === 'SpillError' ||
+      this.codeCell.language !== 'Import'
```
AyushAgrawal-A2, line 108: This will make this `if` condition true for all code cells, checking `RunError` or `SpillError` is redundant

AyushAgrawal-A2 (changes requested):

quadratic-client/src/app/gridGL/cells/tables/Table.ts
davidfig: Yep. We can remove the error check since it's handled further down.

AyushAgrawal-A2: we should also add here check for width and height is 1, otherwise we are adding CellsMarkers for all codecells

ddimaria: We had a single cell check in here, not sure what happened to it.  I'll add it back in.

ddimaria: Updated conditional: 
```typescript
if (this.codeCell.language !== 'Import' && this.isSingleValue()) {
```

ddimaria: It looks like a commit yesterday removed the icons, so will remove the ` && this.isSingleValue()` portion above.

AyushAgrawal-A2 (approved):

luke-quadratic:
Issue, up to you guys if blocking or not 

Dataframe --> single value --> list 
CleanShot 2025-03-14 at 10 05 04@2x

davidkircos (changes requested):
It never adds the headers back
it will only remove them

https://github.com/user-attachments/assets/c2d1b381-2313-4146-af9e-00d6d135f903

davidkircos (approved):

PR #2495, opened 2025-03-13
Guess at first row is heading

ddimaria:
## Description
This PR is a simple approach to guessing the header of a CSV file.  

## Justification
We currently assume that the first row is never the header.  This PR simply gets the types of the first 3 rows and only sets the header if the types of the first row is different from the second and the second is the same as the third.  If there less than 3 rows, it assumes that the first row is not the header.

## Demo
guess-at-first-row


AyushAgrawal-A2 (approved):

davidkircos (approved):
This is great, works most of the time.

Can always improve down the line if we want to

davidkircos (changes requested):
fix rust lint, then gtg

davidkircos (approved):

PR #2496, opened 2025-03-13
improve the starting prompts

davidkircos:
Currently it's not clear that these are examples if you only read the headers (not the subheaders). 
CleanShot 2025-03-13 at 13 19 04@2x

Changed to prompts that work both if the sheet is empty and if the sheet already has users data. If they click on create a chart it will either add example data if the sheet is blank or it will chart the users data on the sheet.
CleanShot 2025-03-13 at 13 18 26@2x


luke-quadratic (approved):
awesome

PR #2430, opened 2025-03-06
infra: improve qa deployment

AyushAgrawal-A2:
- Client builds on gh actions, deploys to s3 and is served via Cloudflare
 - Client and pulumi infra deploys together, deploys only if all builds are successful or no change to infra

PR #2500, opened 2025-03-13
infra: move production client to s3 + cloudflare

AyushAgrawal-A2:
## Relevant issue(s)
no existing issue for this

## Description
follow up of https://github.com/quadratichq/quadratic/pull/2430 for production client deploy on s3 + cloudflare


.github/workflows/production-deploy-client.yml
```diff
@@ -0,0 +1,148 @@
...
+      - main
+    types: [opened, synchronize, reopened]
+
+jobs:
+  build_client:
+    name: Build Client
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    concurrency:
+      group: production-build-client
+      cancel-in-progress: true
```
davidkircos, line 15: i don't think we need to cancel in progress

.github/workflows/production-deploy-client.yml
```diff
@@ -0,0 +1,148 @@
...
+    timeout-minutes: 30
```
davidkircos, line 16: make this 60m so it doesn't fail if it takes a little too long

AyushAgrawal-A2: done

.github/workflows/production-publish-images.yml
```diff
@@ -40,17 +41,10 @@ jobs:
...
+        id: get-ecr-url
         run: |
-          echo "REPO_NAME=quadratic-${{ matrix.service }}" >> $GITHUB_OUTPUT
-
-      - name: Create Public ECR Repository if not exists
-        id: create-ecr
-        env:
-          REPO_NAME: ${{ steps.repo-name.outputs.REPO_NAME }}
-        run: |
-          aws ecr-public create-repository --repository-name $REPO_NAME || true
```
davidkircos, line 53: why remove this?

davidkircos: If we add a service in the future it'll create the ecr automatically, doesn't do anything otherwise

AyushAgrawal-A2: repos are already created, now using the same method of getting ECR_URL as other workflows

AyushAgrawal-A2: >If we add a service in the future it'll create the ecr automatically, doesn't do anything otherwise

saw this later, will add this back

davidkircos (approved):

PR #2502, opened 2025-03-14
Nit: Fix Hover Error + Click Cell, Error Disappears

davidkircos:
When hovering an error, then clicking the cell with the error the error disappears. This fixes that.

New Behavior 
https://github.com/user-attachments/assets/932a9fb2-204c-446e-815c-9b32534e8cfd

Old Behavior
https://github.com/user-attachments/assets/8c565529-5ff8-4460-9d5c-80661e8df474


luke-quadratic (approved):
works as expected 👍

AyushAgrawal-A2 (approved):

PR #2522, opened 2025-03-16
feat: add file support to ai chat

AyushAgrawal-A2:
## Relevant issue(s)
none

## Description
add client to backend to ai providers pipeline for sharing files with AI in chat, no user facing changes in this PR


quadratic-shared/ai/helpers/message.helper.ts
```diff
@@ -53,3 +58,23 @@ export const getSystemPromptMessages = (
 
   return { systemMessages, promptMessages };
 };
+
+export const isToolResultMessage = (message: ChatMessage): message is ToolResultMessage => {
```
ddimaria, line 62: Consider adding these to `quadratic-client/src/app/helpers/files.ts`

ddimaria (approved):

quadratic-shared/ai/helpers/message.helper.ts
AyushAgrawal-A2: this is need in client and api both, hence in quadratic-shared

quadratic-api/src/routes/v0/ai.chat.POST.test.ts
```diff
@@ -15,7 +15,7 @@ const payload: AIRequestBody = {
   messages: [
     {
       role: 'user',
-      content: 'Hello',
+      content: [{ type: 'text', text: 'Hello' }],
```
davidkircos, line 18: add a test with a image too?

davidkircos (approved):

quadratic-api/src/routes/v0/ai.chat.POST.test.ts
AyushAgrawal-A2: image or any file is base64 encoded text, same as normal test
ai sdk in test is mocked, so will will not be testing anything
everything is strongly typed so in case of any mismatch build will fail

PR #2525, opened 2025-03-17
Table columns and rows should be inserted, deleted, and moved at same time as sheet columns and rows

### LINKED ISSUE #2332: Column/row delete from column/row number should also delete the table content

davidfig:
## Relevant issue(s)
Closes #2332 

## Description
- [x] delete data table column when deleting sheet column
- [x] delete data table row when deleting sheet row
- [x] insert data table column when inserting sheet column
- [x] insert data table row when inserting sheet row

## Moved to a separate PR
- [ ] highlight table in selection when it would result in a deletion
- [ ] move data table column when moving sheet column
- [ ] move data table row when moving sheet row

Note: it should not delete/insert/move code tables (unless it's the anchor column or row)


luke-quadratic (approved):
This works great. 

Would like @davidkircos to review on user functionality as well.

quadratic-core/src/grid/data_table/mod.rs
```diff
@@ -407,18 +401,29 @@ impl DataTable {
         }
     }
 
-    /// Returns the height of the data table.
+    /// Returns the height of the data table. The force_table_bounds parameter
+    /// will return the actual height of the table, including any UI elements.
+    /// If false, it includes the array height, which may contain the column
+    /// header row if header_is_first_row is true.
```
AyushAgrawal-A2, line 407: looks opposite, when true return array height

quadratic-core/src/grid/data_table/mod.rs
```diff
@@ -761,6 +704,24 @@ impl DataTable {
...
+            if self.show_columns && !self.is_html_or_image() {
+                if self.show_name {
+                    rows.push(pos.y + 1);
+                } else {
+                    rows.push(pos.y);
+                }
+            }
+        }
+        rows
+    }
```
AyushAgrawal-A2, line 724: should we handle single cell output data tables here, which does not have ui even when show_* flags are true ?

quadratic-core/src/grid/sheet/col_row/rows_delete_table.rs
```diff
@@ -0,0 +1,540 @@
...
+        let mut rows = rows.clone();
+        rows.sort_unstable();
+        rows.dedup();
+        rows.reverse();
+
+        if self.ensure_no_table_ui(&rows) {
+            if transaction.is_user_undo_redo() && cfg!(target_family = "wasm") {
+                crate::wasm_bindings::js::jsClientMessage(
+                    "delete_rows_error".to_string(),
+                    "warning".to_string(),
```
AyushAgrawal-A2, line 199: `JsSnackbarSeverity::Warning.to_string()`

quadratic-core/src/controller/operations/operation.rs
```diff
@@ -59,6 +65,20 @@ pub enum Operation {
...
     },
+    /// Moves a cell value from one place on a sheet to another. This op
+    /// purposefully does not check if the cell value is a data table. This
+    /// should be used to move only the CellValue without impacting any
+    /// other data in the sheet.
+    MoveCellValue {
+        sheet_id: SheetId,
+        from: Pos,
+        to: Pos,
+    },
```
AyushAgrawal-A2, line 76: Not used anywhere, not sure we need this

AyushAgrawal-A2 (approved):

luke-quadratic:
If a column/row deletion intersects the root cell of a chart, the chart is not being deleted.

luke-quadratic (changes requested):
comment above

luke-quadratic (approved):

Issue #1950, opened 2024-10-04
Connections sidebar should create a query in the sheet instead of a new sheet


davidfig:
If you create a connections query from the sheet, it creates the query in a new file instead of inserting it at the cursor position.

jimniels:
Notes from discussion in standup:

- Change the "connections" button in the sidebar so that, when clicked, it shows the cell type menu filtered down to connections
- Change the "connections" dialog in the app so that you cannot view the schema for the connection. Clicking on a connection will allow you to edit its details, but that's it.
  
Possibly consider changing so a click on the connection takes you to edit details (on both app and dashboard) and then dashboard gets an extra special button that takes you to today's schema viewer where you can create a new file querying that connection, e.g.

![CleanShot 2024-11-19 at 11 16 18@2x](https://github.com/user-attachments/assets/08f2628c-84d3-409b-a260-ba048e8fb7e7)

PR #2526, opened 2025-03-17
feat: tweaks to connections flow

### LINKED ISSUE #2269: Improved Connections flow
Objective: make the connections menus more intuitive for users to go from created connection to using within sheet.

[placeholder, @jimniels plans on self-scoping and proposing how this should look when time comes to work on this]

jimniels:
## Relevant issue(s)

Resolves #2269 and #1950

## Description

From the dashboard, the connections menu continues to work as it works today.

From the app, when you click the connections menu, it will show the list of connections. 

CleanShot 2025-03-17 at 15 09 30@2x

The "meta" info for connections you see on the dashboard is accessible via the help icon.

CleanShot 2025-03-17 at 15 10 39@2x

Clicking on a connection will open a code cell for that connection at the current cursor (equivalent to opening the cell type menu and choosing the connection you clicked on).

### Block localhost

Additionally, the schema was updated for Postgres, MySql, and MsSql to disallow localhost-related hostnames. Users will see an error when they try to create a connection that way.

image

## Testing

If you click to insert a connection when your cursor is on a cell that is already a code cell or a data table, it will pop a message noting that you can't insert a connection over an existing code cell or inside a data table.

So, for testing, expect that the feature works when your cursor is on:

- A cell that's empty
- A cell that has static data in it

But it pops a message when your cursor is on:

- A code cell
- A table

Additionally, you should not be able to create a new connection that points to localhost, 127.x.x.x, 0.0.0.0, or other local IP addresses.


jimniels:
You can get into a funky state where if you insert a connection on a data table, it'll open the code editor.

https://github.com/user-attachments/assets/e4aa6c73-3336-4fe3-aa31-94833a7c94d9

This is unexpected and weird. Might want to think through this a bit more...

jimniels:
@AyushAgrawal-A2 is going to add support so when you click to insert a connection from the connections menu, but your cursor is on a code cell or inside a data table, it will pop a message saying "Cannot create a connection inside a code cell / data table". Similar to what happens today when you press `/` inside a data table.

CleanShot 2025-03-18 at 11 04 23@2x

davidkircos (approved):
This works really well! This is a great change for connections

PR #2540, opened 2025-03-21
Viewport Improvements

davidkircos:
This does a few things to the Viewport:
- While dragging prevent snap back until dragging is released (when pressing space to drag and beyond the bounds.)
- Don't allow momentum scrolling past the bounds, this was a bit glitchy looking. You can still zoom beyond the bounds, and drag past the bounds.
- Reduce the time after zooming to snap back.

This also makes some changes to the GridHeadings:
- always shows A and 1 heading labels
- on the numbered axis adds a step to count by 2s showing 1, 3, 5, etc
- on the vertical axis, make the counting by numbers that fit into the alphabet 
- fixes a bug where if you zoom out past 13% all the letter labels start to disappear

### DIFF
--- a/quadratic-client/src/app/gridGL/UI/gridHeadings/GridHeadings.ts
+++ b/quadratic-client/src/app/gridGL/UI/gridHeadings/GridHeadings.ts
@@ -70,10 +70,10 @@ export class GridHeadings extends Container {
   }
 
   private findIntervalX(i: number): number {
-    if (i > 100) return 50;
+    if (i > 100) return 52;
     if (i > 20) return 26;
-    if (i > 5) return 10;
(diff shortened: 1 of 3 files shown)


luke-quadratic:
This is awesome - very snappy! 

One thing that might make sense in this PR (exists in prod as well so not marking requested changes) is the behavior after scrolling down by the origin is the super slow going back up - hard to explain see GIF. 

CleanShot 2025-03-21 at 08 01 24

luke-quadratic (approved):
Very snappy! And doesn't seem to introduce any buggy behavior. 

Only note is see comment above if that's relevant to solve in this PR.

PR #2541, opened 2025-03-21
fix: use presigned urls for file-storage

AyushAgrawal-A2:
## Description
File storage now uses presigned urls same as s3, so same file permissions flow works now for shared files on self-hosted app


davidkircos (approved):
lgtm

PR #2546, opened 2025-03-21
perf: improve getCells

AyushAgrawal-A2:
## Description
Improves getCells speed in py and js


luke-quadratic (approved):
Huge speed up and not finding any bugs in the implementation

davidkircos (approved):

quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
```diff
@@ -835,7 +835,10 @@ class Core {
       chart_pixel_output: null,
       has_headers: false,
     };
-    this.gridController.calculationComplete(JSON.stringify(codeResult));
+    const jsCodeResult = JSON.stringify(codeResult);
```
ddimaria, line 838: This pattern is repeated in multiple places, it might be a good idea to consolidate the code to make it DRY.

quadratic-kernels/python-wasm/quadratic_py/utils.py
```diff
@@ -178,12 +177,16 @@ def to_python_type_df(value: str, value_type: str) -> int | float | str | bool:
             return str(value)
         elif value_type == "logical":
             return ast.literal_eval(normalize_bool(value))
+        elif value_type == "time":
```
ddimaria, line 180: It looks like `elif value_type == "time":` appears twice

ddimaria (approved):
Great PR.  I have a refactor request and a possible extra conditional branch

quadratic-kernels/python-wasm/quadratic_py/utils.py
AyushAgrawal-A2: done

quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
AyushAgrawal-A2: done

PR #2551, opened 2025-03-23
chore: use u8 type in code response types

AyushAgrawal-A2:
## Description
reduces getCells data size, perf should is same as the base branch


quadratic-client/src/app/web-workers/javascriptWebWorker/worker/javascript/runner/javascriptLibrary.ts
```diff
@@ -207,13 +207,30 @@ export const relCells = (deltaX0: number, deltaY0: number, deltaX1: number, delt
 
 export const rc = relCell;
 
+// type_u8 as per cellvalue.rs
+export enum CellValueType {
```
ddimaria, line 211: We may not have to redefine these.  There should be many options available to just have these defined in Rust.

quadratic-core/src/controller/execution/run_code/mod.rs
```diff
@@ -456,7 +456,7 @@ impl GridController {
                 }
             } else if let Some(output_value) = js_code_result.output_value {
                 let (cell_value, ops) =
-                    CellValue::from_js(&output_value[0], &output_value[1], start.into(), sheet)
+                    CellValue::from_js(output_value.to_owned(), start.into(), sheet)
```
ddimaria, line 459: I haven't traced out the code, but wanted to make sure this copy was intentional.

ddimaria: `from_js()` used to accept read only copies, but now requires ownership.

quadratic-kernels/python-wasm/quadratic_py/utils.py
```diff
@@ -5,12 +5,31 @@
...
 from decimal import Decimal, DecimalException
+from enum import Enum
 from typing import Tuple
 
 import numpy as np
 import pandas as pd
 import pytz
 
+# type_u8 as per cellvalue.rs
+class CellValueType(Enum):
```
ddimaria, line 16: I can this being brittle.

ddimaria (approved):
Some optional, non-blocking comments/requests.

Issue #2383, opened 2025-03-03
Connection tables getting renamed on run


luke-quadratic:
If you name a connection table and then re-run, the name will get overwritten, e.g. "MyConnection" after rerunning will become Postgres2

PR #2561, opened 2025-03-24
Preserve data table name across re-runs

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2383

## Description
Re-running a sql connection after the table is renamed preserves the name.

## Demo
rename-data-table


quadratic-core/src/controller/execution/run_code/mod.rs
```diff
@@ -614,7 +602,7 @@ mod test {
         };
         let mut new_data_table = DataTable::new(
             DataTableKind::CodeRun(new_code_run),
-            "Table_2",
+            "Table_1",
```
ddimaria, line 605: @davidfig I kept the names the same here b/c we want to preserve the data table name on re-runs.

davidkircos (approved):
Functionality review looks good to me! Nice.

davidfig (approved):

PR #2573, opened 2025-03-25
chore: use grid controller in formula ctx

AyushAgrawal-A2:
## Description
Switches to using `GridController` in `ctx` context of formulas instead of `Grid`


HactarCE (approved):
all looks good, nice job! I didn't realize quite how much work that was gonna be 😅

ddimaria (approved):
Easy one to approve 🙌

PR #2576, opened 2025-03-26
chore: use cached a1_context everywhere

AyushAgrawal-A2:
## Description
replaces the `a1_context()` fn, which create new context object, to cached a1_context


davidfig:
Deployed failed.

ddimaria:
@AyushAgrawal-A2 97 files changed here.  Can you point out the ones where more than a parameter rename is happening?

AyushAgrawal-A2:
> @AyushAgrawal-A2 97 files changed here. Can you point out the ones where more than a parameter rename is happening?

Changes in this PR include - 
 - in many places the gird's or sheet's `a1_context` function was being used, which creates a new a1_context from scratch, this pr replaces all those calls to using GridController's cached a1_context

 - functions on the sheet needs cached `a1_context` to be passed as a function argument

 - related test update due to additional arg in function definition

- rename `context` name to `a1_context` for consistency, (this is in a separate commit, 2nd commit)

quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
```diff
@@ -202,11 +217,16 @@ impl GridController {
             copy_formats,
         } = op
         {
-            if let Some(sheet) = self.try_sheet_mut(sheet_id) {
-                sheet.insert_row(transaction, row, copy_formats, true);
+            if let Some(sheet) = self
+                .grid
+                .sheets
+                .iter_mut()
+                .find(|sheet| sheet.id == sheet_id)
```
HactarCE, line 224: Why not `self.try_sheet_mut(sheet_id)` or `self.grid.try_sheet_mut(sheet_id)`? then it's more concise here and we can optimize it later with a hashmap.

There's some other places in this PR where the same pattern shows up.

quadratic-core/src/controller/send_render.rs
```diff
@@ -223,11 +223,12 @@ impl GridController {
...
-            sheet.recalculate_bounds()
-        } else {
-            false
-        };
+        let recalculated = self
+            .grid
+            .sheets
+            .iter_mut()
+            .find(|sheet| sheet.id == sheet_id)
+            .is_some_and(|sheet| sheet.recalculate_bounds(&self.a1_context));
```
HactarCE, line 231: good use of `is_some_and()`

HactarCE (approved):
wow! unfortunate that we have to pass the context around _everywhere_ but I understand why.

just one comment about `sheets.iter().find()`; otherwise LGTM!

quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
AyushAgrawal-A2: rust complains that cannot borrow a mutable and immutable ref at the same time,
I need to borrow 
 - `sheet` as mutable from `GridController -> Grid`
 - `a1_context` as immutable from `GridController`

AyushAgrawal-A2: replaced all with `self.grid.try_sheet_mut(sheet_id)`

PR #2578, opened 2025-03-26
Perf metrics

ddimaria:
- [x] Creates a test harness for benchmarking individual functions or lets criterion benchmark
- [x] Some performance enhancement to the speed of excel file imports (about 25% atm)


quadratic-core/src/controller/operations/import.rs
```diff
@@ -239,7 +239,7 @@ impl GridController {
     /// Imports an Excel file into the grid.
     pub fn import_excel_operations(
         &mut self,
-        file: Vec<u8>,
+        file: &[u8],
```
AyushAgrawal-A2, line 242: can we benchmark what is the time difference if we take a owned object vs a ref here
taking ownership is good for memory, we should atleast the drop the file after generating operations

AyushAgrawal-A2 (approved):
just one doubt which I will followup in slack

quadratic-core/src/controller/operations/import.rs
AyushAgrawal-A2: `file: &[u8]` slice is faster

ddimaria: We benchmarked and `&[u8]` was 1s faster in the 10k excel import benchmark.

PR #2509, opened 2025-03-14
feat: application loading state

davidkircos:
## Description **(by @jimniels)**

To test, here's the experience you should see:

1. Initial paint from browser: gray logo
2. Animated gif (colored logo) fades in after ~.5s (if the app loads fast enough, you won't even see this fade in)
3. Animated gif runs on top of the gray logo until
4. App renders

I've added mixpanel tracking so we get events every time somebody loads the app anew and we can track how long it takes people to get to the initial render of the app (with information about which route is loading)

---

**Footnotes:**

This is the smoothest way for us have a loading state without weird/janky loading state re-renders. It's a stepping stone to a more enhanced loading state we can work on later that actually gives live updates of the different things that are happening, e.g.

- file downloading...
- multiplayer loading...
- offline loading...

(Leaving link to [thi diagram](https://www.figma.com/board/PRlwj300J5sWahHEmCtU2W/App-Loading?node-id=0-1&t=Yso7tYiN3dyj9IEk-1) for reference on how the mechanics are working under the hood)

PR #2579, opened 2025-03-26
perf: upgrade to framework mode in react-router v7

jimniels:
Following [this guide](https://reactrouter.com/upgrading/router-provider)

Unlocks the ability to properly solve #2509 

## Description
Changes to leverage react-router's newest features, including the `react-router` vite plugin, which provides the ability to render+hydrate so no more static `index.html` file. Rather the build will create the root html file.

## (Optional) Any extra testing considerations?
tk


jimniels:
Not convinced that this is worth our time/effort. We may eventually want framework mode, but I don't see a compelling user/business need to make these changes ATM and introduce possible regressions when we're not gaining anything by doing it.

In our particular case, [the docs say](https://reactrouter.com/start/modes) to use "data mode" if:

> started a data router in v6.4 and are happy with it

Which is us. So going to close this PR and keep it as a reference for later if/when we ever decide to upgrade to framework mode.

Issue #2592, opened 2025-03-27
Copying multiple columns from one table to another sheet copies entire table


luke-quadratic:
## Describe the bug
Copying multiple columns from one table to another sheet copies entire table
![Image](https://github.com/user-attachments/assets/38a0408b-cf38-453f-9511-8daa31bea087)

## Expected behavior
Copying multiple columns would create new table with just those columns, not with entire table

## Reproduction
See video above

Issue #2591, opened 2025-03-27
Copy-pasting single column out of table with default columns exhibits strange behavior


luke-quadratic:
## Describe the bug
Note behavior in attached video when copy-pasting first column. 

Cursor bounces up and column header doesn't get copied. 

## Expected behavior
_DISCUSS THIS BEFORE IMPLEMENTING, NOT 100% SURE THIS IS CORRECT:_ Column header gets copied since it's part of selection and cursor doesn't bounce up

Issue #2369, opened 2025-03-01
bug: Pasting multiple columns of data from a table shifts down a row


cole-at-quadratic:
When pasting a single column to another tab, the data goes where I want it. When pasting two columns, it shifts down a row.

luke-quadratic:
@cole-at-quadratic can't recreate - please post a vid recreating / clear steps etc

PR #2600, opened 2025-03-28
Copy columns unexpected paste

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2369
Fixes https://github.com/quadratichq/quadratic/issues/2591
Fixes https://github.com/quadratichq/quadratic/issues/2592

## Features
* Copying an entire column copies the column heading and the data without the table UI
* Copying multiple column copies the column heading and the data of each column without the table UI
* Copying the entire table copies the table UI

## Demo
sane-data-table-copy-paste


luke-quadratic:
There's one unintentional state I can find where all data is selected but the whole table isn't selected. One of these two must be incorrect? I assume it's the first one where all data is selected but entire table doesn't get selected? 

To recreate, drag over all data in the table.
CleanShot 2025-04-03 at 07 50 42

To recreate: If you click the first column, then shift click to the last one, the entire table gets highlighted. 
CleanShot 2025-04-03 at 07 48 19

luke-quadratic (approved):
Determined that above is correct behavior, good on functionality

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -326,6 +327,8 @@ impl GridController {
                     return Err(Error::msg(message));
                 }
 
+                println!("source_pos: {:?}", source_pos);
+
```
HactarCE, line 331: stray `println!()`

HactarCE (approved):
LGTM! just a stray `println!()` to clean up

davidfig (approved):

Issue #2459, opened 2025-03-08
bug: importing the same filename twice in a row (deleting in between) doesn't work


N-Demir:
I imported a csv with Insert -> Data -> CSV and then deleted that data from quadratic so that I was on an empty sheet again. I then changed the data in the csv locally and tried to import it again but nothing happened. Importing a different named csv, deleting it, then importing the original one worked. Kinda weird!

luke-quadratic:
This is an awesome find - marking high prio, thanks

gabrielrapoport5:
I've heard this from a few users where they get an error when importing multiple files

ddimaria:
I'm able to duplicate this using the user's instructions.  Dragging after deleting works fine though.

ddimaria:
This is happening with any supported file when the next import has the same name.  It's due to the onChange event not firing in these cases.

PR #2604, opened 2025-03-28
bugfix: clear GridFileInput value after import

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2459

## Description
This is happening with any supported file when the next import has the same name. It's due to the onChange event not firing in these cases.

## Demo
import-csv-same-name


davidfig (approved):

luke-quadratic (approved):
This works as intended, noting that in exploring this I discovered a nasty bug with insert --> data when the data is .xlsx, I filed that here because I assume it doesn't make sense to do here - https://github.com/quadratichq/quadratic/issues/2610

Issue #2266, opened 2025-02-08
bug: `101-250` is parsed as a date


davidkircos:
#2248 is related

(There are many of these cases, hopefully a lot can be tackled all at once)

HactarCE:
Fixed in #2605

PR #2605, opened 2025-03-28
Improve date parsing

### LINKED ISSUE #2248: Date parsing for "Mon Year" doesn't work (Jan 2024)
To recreate type "Jan 2024" into a sheet and it'll get parsed as 0024 instead of 01/24.

HactarCE:
This should fix #2248 and #2266.


luke-quadratic:
One introduced common case I found was that 4/10 produces 04/01/0010 when it should produces 4/10/current_year

(continuing to search for more cases)

luke-quadratic (approved):
Works great, created separate tracking issue for any future follow-up date parsing cases that get caught (I'm sure there's many we still need to discover that are more edge casey - the ones covered in this PR are very important)

PR #2607, opened 2025-03-28
fix: duplicate sheet name on excel import shows app crashed screen

AyushAgrawal-A2:
## Description
Prevents showing app crashed page when dropping in a excel file with duplicate name.

The changes in this pr are also need for csv and parquet files, will do that in a separate pr via qa


quadratic-core/src/wasm_bindings/controller/import.rs
```diff
@@ -103,11 +104,26 @@ impl GridController {
...
-
-        Ok(())
+    ) -> Result<JsValue, JsValue> {
+        match self.import_excel(file, file_name, cursor) {
+            Ok(_) => Ok(serde_wasm_bindgen::to_value(&JsResponse {
+                result: true,
+                error: None,
+            })?),
+            Err(e) => {
+                dbgjs!(format!(
```
ddimaria, line 114: Duplicate format code here, maybe make a variable:
```rust
let error = format!("Error importing Excel file: {:?}, error: {:?}", file_name, e);
 dbgjs!(error);
 
 Ok(serde_wasm_bindgen::to_value(&JsResponse {
    result: false,
    error,
})?)
```

AyushAgrawal-A2: done

ddimaria (approved):

PR #2617, opened 2025-03-30
Billing Soft Limit

davidkircos:
Todo
- [x] add env variable for ai usage limit in quadratic-selfhost, aws for preview, qa and prod heroku

## Relevant issue(s)

Known issues
- [x] Edit button on messages is not clickable except at the edges
- [x] Message is not cleared after sending

Encourages AI users to upgrade to Quadratic Pro after a certin number of prompts. Users who exceed this limit will have to wait for their prompts to be submitted.

https://github.com/user-attachments/assets/6802ac3b-e8fe-4e9b-a9c4-86f47b717468


luke-quadratic (approved):
works, awesome one

PR #2620, opened 2025-04-01
#DATA for reference is popping up in Python 

luke-quadratic:
## Description

There are a few places in the docs where #DATA aren't explicitly explaining they're for Formulas and JavaScript. As a result a lot of Python requests are popping up that erroneously use this reference.


ddimaria (approved):

ddimaria:
@luke-quadratic CI is failing

PR #2621, opened 2025-04-01
Matplotlib attempts are emerging from AI

luke-quadratic:
## Description
Matplotlib attempts are emerging in a very small (but meaningful) amount of charting attempts. Little extra armor against this emerging attempt by AI. It's not surprising as most common library but it should never happen. 

Anecdotal: in a 6 hour session of 100+ charting attempts, this expressed itself about 5 times.

### DIFF
--- a/quadratic-api/src/ai/docs/PythonDocs.ts
+++ b/quadratic-api/src/ai/docs/PythonDocs.ts
@@ -335,7 +335,7 @@ df = q.cells('A:B'), first_row_header=True)
 
 # Charts/visualizations
 
-Plotly is the only charting library supported in Quadratic. Do not try to use other libraries like Seaborn or Matplotlib. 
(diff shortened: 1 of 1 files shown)


quadratic-api/src/ai/docs/PythonDocs.ts
```diff
@@ -335,7 +335,7 @@ df = q.cells('A:B'), first_row_header=True)
 
 # Charts/visualizations
 
-Plotly is the only charting library supported in Quadratic. Do not try to use other libraries like Seaborn or Matplotlib. 
+Plotly is the only charting library supported in Quadratic. Do not try to use other libraries like Seaborn or Matplotlib. Matplotlib DOES NOT WORK in Quadratic.
```
ddimaria, line 338: What about adding `Seaborn` to the `DOES NOT WOK in` message?

luke-quadratic: Seaborn has never emerged as it's not very popular. Just adding reinforcement on matplotlib which is the most popular so emerges on occasion.

davidkircos (approved):
Great!

PR #2625, opened 2025-04-01
Fix stuck in rename

### LINKED ISSUE #2432: Stuck in tab rename if you rename and don't submit
On blur of tab rename, save and submit. Right now it can stay in the editing state while you go and do other things on the sheet.

davidfig:
## Relevant issue(s)
Fixes #2432 

## Description
Renaming sheet should not get stuck. If clicked away with invalid entry, then it will revert. Otherwise, it should return to the edit.


ddimaria (approved):

luke-quadratic (approved):
Works

davidkircos (approved):
nice fix!

PR #2627, opened 2025-04-01
Ignore table edges for tables that are of size 1

### LINKED ISSUE #2494: [BUG] CMD + Arrow doesn't work on formulas
## Describe the bug
CMD + Arrow doesn't work on formulas

## Expected behavior
It should perform as it does on regular data. 

https://github.com/user-attachments/assets/730efb52-85f1-4ff2-b518-eb73a833b06b

davidfig:
## Relevant issue(s)
Fixes #2494 

## Description
For ctrl+arrow, we treat single-value tables the same as sheet values, so we skip over multiple rows or columns.


ddimaria (approved):

luke-quadratic (approved):
Works

davidkircos (approved):
Nice fix, this works great.

PR #2631, opened 2025-04-02
feat: AI PDF Import

### LINKED ISSUE #2142: Import PDFs to the sheet with AI
## Basic description 
Various file types are becoming increasingly able to get parsed and understood with AI. These specific types like PDF, images, etc should get automatically placed in chat with focused context. 

PDFs are first validation of this feature, let's conquer that first. This feature would get triggered if a PDF is dropped directly into the sheet or directly into the chat. In both cases they're placed into chat as focused context. 

The following data types would remain default inserted to the sheet
- CSV
- Parquet
- Excel
- 
As part of this PR we are retiring model choice. Model choice should only be enabled for dev with ?debug. Users should get whatever we view are the best model choices for every scenario by default (currently Claude 3.5 for standard scenarios, claude 3.7 thinking when thinking is toggled on, and either claude or flash when PDFs etc are in the context). 

If single or multiple files are uploaded they should continue to get shoved into context. 

Image

AyushAgrawal-A2:
Todo
- [x] update `GCP_REGION` in heroku for qa and prod 

## Relevant issue(s)
closes #2142


davidkircos (approved):
awesome!

PR #2636, opened 2025-04-02
improved right click on table and table column

### LINKED ISSUE #2453: Column # incorrect in context menu
Image

davidfig:
## Relevant issue(s)
Fixes #2453

## Description
Improved behavior of right click on table and table columns. It now changes the selection so it's clear what will trigger.


quadratic-core/src/a1/a1_selection/query.rs
```diff
@@ -564,6 +564,29 @@ impl A1Selection {
...
+        a1_context: &A1Context,
+    ) -> bool {
+        let Some(table) = a1_context.try_table(table_name) else {
+            return false;
+        };
+        let mut found_column = false;
+        self.ranges.iter().all(|range| match range {
+            CellRefRange::Table { range } => {
+                if range.col_range.has_col(column, table) {
+                    found_column = true;
```
ddimaria, line 583: Break out of loop here?

quadratic-core/src/a1/a1_selection/query.rs
```diff
@@ -1383,4 +1406,54 @@ mod tests {
         let selection = A1Selection::test_a1("A1:C3");
         assert_eq!(selection.selected_rows_finite(&context), vec![1, 2, 3]);
     }
+
+    #[test]
```
ddimaria, line 1410: Great tests 💯

ddimaria (approved):
Code looks good, just have a small question/request.

luke-quadratic:
Reversed issue now 
CleanShot 2025-04-02 at 12 22 52@2x

luke-quadratic (changes requested):
see comment above

quadratic-core/src/a1/a1_selection/query.rs
davidfig: We can't break here since we need to check the rest of the ranges to ensure they're all within the same table

AyushAgrawal-A2 (approved):

PR #2649, opened 2025-04-03
Fix command palette column/row inserts 

### LINKED ISSUE #2638: Insert/delete row & column from command palette doesn't work
## Describe the bug
Insert/delete row & column from command palette doesn't work 

## Expected behavior
Should mirror menu behaviors

luke-quadratic:
## Relevant issue(s)
Resolves #2638 

Based on user report from yday (easier to just resolve than scope/assign/add to the end of the list)


quadratic-client/src/app/ui/menus/CommandPalette/commands/ColumnRow.tsx
```diff
@@ -19,7 +20,16 @@ const columnRowCommandGroup: CommandGroup = {
         );
       },
     },
+    Action.InsertColumnLeft,
+    Action.InsertColumnRight,
+    Action.DeleteColumn,
```
AyushAgrawal-A2, line 25: we should not mix command objects with actions here

quadratic-client/src/app/ui/menus/CommandPalette/CommandPalette.tsx
```diff
@@ -114,20 +115,24 @@ export const CommandPalette = () => {
...
+                        // @ts-expect-error - Some actions expect args but can run without them
+                        actionSpec.run();
+                      } else {
+                        // @ts-expect-error - Pass empty object as args if args are required
+                        actionSpec.run({});
+                      }
+                    }}
                     icon={'Icon' in actionSpec && actionSpec.Icon ? <actionSpec.Icon /> : null}
-                    shortcut={keyboardShortcutEnumToDisplay(commandOrAction)}
+                    shortcut={keyboardShortcutEnumToDisplay(commandOrAction as Action)}
```
AyushAgrawal-A2, line 135: we should not type cast here but check type and call functions accordingly
type casting should not be necessary and will be very brittle

AyushAgrawal-A2 (changes requested):

quadratic-client/src/app/ui/menus/CommandPalette/commands/ColumnRow.tsx
AyushAgrawal-A2: done

davidfig (approved):

AyushAgrawal-A2 (approved):

PR #2661, opened 2025-04-04
fix: calculationComplete failed to parse response

AyushAgrawal-A2:
## Description
fixes: 

image


quadratic-kernels/python-wasm/quadratic_py/process_output.py
```diff
@@ -94,7 +94,7 @@ def process_output_value(output_value):
                 col_length_2d = len(array_output[row])
                 for col in range(0, length_2d):
                     if col > col_length_2d - 1:
-                        typed_array_output[row][col] = ("", "blank")
+                        typed_array_output[row][col] = ("", 0)
```
ddimaria, line 97: Strange, I thought you made this change already 🤔

ddimaria (approved):

quadratic-kernels/python-wasm/quadratic_py/process_output.py
AyushAgrawal-A2: missed this one

PR #2678, opened 2025-04-07
Tables are selected when fully contained within a selection box

davidfig:
## Description
Tables show as highlighted when contained within a selection box (helps to indicate which tables will be deleted).


luke-quadratic:
Noting some incorrect behavior, hard to explain, see video where Table1 doesn't get selected on first pass . 

CleanShot 2025-04-07 at 12 53 49

luke-quadratic (changes requested):
See comment above - either tables should get selected or they shouldn't be deleted unless actually selected.

davidfig:
We should talk about the delete behavior. I think it's wrong. Only code cells should be deleted when their anchor cells are deleted.

luke-quadratic:
Full summary of conversation: 

**Data table** 
- Entire table gets selected/deleted if any part of table header is selected or entire table content is selected 

CleanShot 2025-04-08 at 11 20 34@2x

**Code table** 
- Entire code table will get selected if any part of header is selected, will not be highlighted/selected otherwise and should not be deleted unless header is selected + highlighted; both in this image count as entire table 
CleanShot 2025-04-08 at 11 33 07@2x

- Case where everything but header is selected does not select table, only available option here is copy 
CleanShot 2025-04-08 at 11 27 36@2x

davidfig:
This is ready for a re-review. It now selects the same as it deletes. Please open a separate issue for the comment above.

Issue #1282, opened 2024-04-17
Horizontal and Vertical Scrollbars for Sheets


davidfig:
Similar to Figma's implementation

PR #2684, opened 2025-04-08
Scrollbars

davidfig:
## Relevant issue(s)
#1282 

## Description
- [x] Draw horizontal and vertical scrollbars 
- [x] Hide scrollbar if content fits on screen and at offset 0
- [x] Allow dragging of scrollbar
- [x] Add setting to show/hide scrollbar


luke-quadratic (changes requested):
Slack 🧵 for feedback

PR #2685, opened 2025-04-08
feat: make theme switcher accessible in dashboard sidebar

jimniels:
## Description

Clones the theme picker from the app (on the bottom left) into the sidebar of the dashboard (on the bottom left) so it's universally accessible via a single click.

- The `/account` page was removed, and now clicking the user avatar shows a popup with the info that was previously on the `/account` page
- Theme picker is a flyout, same as in the app

https://github.com/user-attachments/assets/bfa7773c-3284-4dfa-b408-1802170cac0d


luke-quadratic (approved):
Works as expected, super clean! 

Tested by trying many themes and refreshing to confirm they stayed applied.

PR #2689, opened 2025-04-09
fix: prevent heroku timeout for pdf import

AyushAgrawal-A2:
## Description
prevents heroku's 30s request timeout by sending a default message every 25s

this is a temporary hack until we move off of heroku, required for pdf imports which can take a few minutes before receiving any response from gemini, (gemini 2.5 pro does not stream tool calls)


davidkircos (approved):
lgtm

ddimaria (approved):

PR #2693, opened 2025-04-09
fix: duplicating sheet duplicates all code cells with same name

AyushAgrawal-A2:
## Relevant issue(s)
#2281 2692

### DIFF
--- a/quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
+++ b/quadratic-core/src/controller/execution/execute_operation/execute_sheets.rs
@@ -67,9 +67,11 @@ impl GridController {
                     data_table.name = unique_name.to_owned().into();
 
                     // update table context for replacing table names in code cells
-                    if let Some(mut table_map_entry) = context.table_map.remove(&old_name) {
-                        table_map_entry.sheet_id = sheet_id;
-                        context.table_map.insert(table_map_entry);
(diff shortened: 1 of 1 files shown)


quadratic-core/src/controller/execution/execute_operation/execute_data_table.rs
```diff
@@ -151,9 +160,10 @@ impl GridController {
             // select the entire data table
             Self::select_full_data_table(transaction, sheet_id, data_table_pos, &data_table);
 
-            let old_value = sheet.set_cell_value(data_table_pos, cell_value);
-            let (old_index, old_data_table) =
-                sheet.data_tables.insert_sorted(data_table_pos, data_table);
+            let old_value = sheet.set_cell_value(data_table_pos, cell_value.to_owned());
```
ddimaria, line 163: You ended up making copies here rather than at the beginning of the fn, but I guess this avoids copying the sheet_id.

ddimaria (approved):

quadratic-core/src/controller/execution/execute_operation/execute_data_table.rs
AyushAgrawal-A2: I am enforcing unique table name in add data table, the forward op needs to contain `data_table` with the new name, hence cloning after calling `unique_data_table_name` and not at the start

luke-quadratic (approved):
Works

PR #2695, opened 2025-04-09
add an ab test on ai exceeded copy

davidkircos:
Run an A/B test using Mixpanel on the copy for upgrading to Pro. 

Uses `team.uuid` to determine control vs variant group, so a user will always see the same one.

Control
CleanShot 2025-04-09 at 15 59 47@2x

Variant
CleanShot 2025-04-09 at 16 00 33@2x

### DIFF
--- a/quadratic-client/src/shared/components/ABTest.tsx
+++ b/quadratic-client/src/shared/components/ABTest.tsx
@@ -1,6 +1,7 @@
-import { useFileRouteLoaderData } from '@/shared/hooks/useFileRouteLoaderData';
+import { editorInteractionStateTeamUuidAtom } from '@/app/atoms/editorInteractionStateAtom';
 import mixpanel from 'mixpanel-browser';
 import { memo } from 'react';
+import { useRecoilValue } from 'recoil';
 
 type ABTestProps = {
   name: string;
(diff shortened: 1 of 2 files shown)


AyushAgrawal-A2 (approved):

quadratic-client/src/shared/components/ABTest.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const ABTest = memo(({ name, control, variant, probability = 0.1 }: ABTestProps) => {
+  const teamUuid = useRecoilValue(editorInteractionStateTeamUuidAtom);
+
+  // Convert the team's UUID to a number between 0 and 1
+  // This is a simple hash function that will return a number between 0 and 1
+  // This will allow us to split the users into two groups, control and variant
+  // consistently putting the same user in the same group by using the same UUID
+  const hash = parseInt(teamUuid.replace(/-/g, '').slice(0, 8), 16) / 0xffffffff;
```
AyushAgrawal-A2, line 20: not sure why `replace(/-/g, '')` is needed here, doesn't hurt either

davidkircos: I think so it doesn't parse as a negative int.

AyushAgrawal-A2: sample team id - `9c845bca-2fe6-42eb-bc3f-1f0dc051970d`
slice will take the first 8 characters - `9c845bca`
dashes are always ignored because of `.slice(0, 8)`

davidkircos: fair

PR #2697, opened 2025-04-10
chore: add chart image in tool result response

AyushAgrawal-A2:
## Description
- When a py / js chart is generated by ai (analyst or assistant), the chart is fed back as an image to ai as tool result. On multiple iterations, only the last image is sent in chat, older ones are dropped
- Add plotly chart to thumbnail
- Fixes auto resize of plotly chart
- Fixes re-renders of Goto dropdown menu, when not open


luke-quadratic:
I haven't been able to create a case yet where being able to see the chart triggered the AI to attempt a fix. 

Or is this purely so that on subsequent messages the chart is in context and isn't intended as a tool to respond if AI creates a poor chart?

quadratic-client/src/app/gridGL/HTMLGrid/htmlCells/HtmlCell.ts
```diff
@@ -345,4 +343,64 @@ export class HtmlCell {
       this.activate();
     }
   }
+
+  private autoResize = () => {
```
davidfig, line 347: ❤️

davidfig (approved):

luke-quadratic (approved):
Awesome, tested: 
Managed to create an image with bad result, AI persevered until a fix and was aided by being able to see it.

PR #2698, opened 2025-04-10
Scrollbars (div)

davidfig:
This is a possible replacement for [#2684](https://github.com/quadratichq/quadratic/pull/2684) using divs instead of pixi to render the scrollbars. This allows u to place it over iframes like python charts.

## Relevant issue(s)
#1282 

## Description
- [x] use div instead of pixi to render scrollbars
- [x] Draw horizontal and vertical scrollbars 
- [x] Hide scrollbar if content fits on screen and at offset 0
- [x] Allow dragging of scrollbar
- [x] Add setting to show/hide scrollbar


AyushAgrawal-A2 (approved):

luke-quadratic (approved):
awesome - they feel perfect

Issue #1941, opened 2024-10-02
Code editor panels lack protections on smaller windows, resizing, etc.


luke-quadratic:
Some states that feel incorrect below as images -
![CleanShot 2024-10-02 at 10 27 22@2x](https://github.com/user-attachments/assets/29dd7291-0ed7-482e-8708-2a1dcca5c511)
![CleanShot 2024-10-02 at 10 28 16@2x](https://github.com/user-attachments/assets/eaa58561-5f2d-4a74-ab8e-f1250bda8edd)
![CleanShot 2024-10-02 at 10 28 46@2x](https://github.com/user-attachments/assets/33904814-36cd-4815-8894-edf74a3fb004)

Issue #1573, opened 2024-07-19
bug: panels don't resize with browser


jimniels:
When you resize the browser window, the panels don't resize their heights to fill the new height. 

This can get you into a weird state, especially if you shrink the browser window and your panes can disappear.

https://github.com/user-attachments/assets/ec597662-24b9-429a-9605-c324b920d77d

vayunbiyani:
Hi @jimniels I've created a PR for the fix [here](https://github.com/quadratichq/quadratic/pull/1822). Hope it helps

Issue #449, opened 2023-04-28
feature: customizable paneling within the app


jimniels:
Like how Zed just lets you drag panels around wherever you want, would be neat if Quadratic let you put things wherever you need them given your current working context.

![Apr-28-2023 10-20-36](https://user-images.githubusercontent.com/1316441/235201936-61a2e6db-860f-4d76-9a95-84e26f0e29bf.gif)

For example: in Zed I can easily create a view with 4 panels by just dragging them around:

<img width="1548" alt="CleanShot 2023-04-28 at 10 16 21@2x" src="https://user-images.githubusercontent.com/1316441/235201996-f3b4e9d4-92da-4f3d-b296-75455aaa8388.png">

luke-quadratic:
Did a lot of this via recent paneling work #1334 , reopen if more paneling work becomes a prio in the future

PR #2708, opened 2025-04-11
fix: code editor resizes beyond its bounds

### LINKED ISSUE #2196: Code editor can be expanded further than its bounds
This expansion ends up hiding buttons etc., see gif

CleanShot 2025-01-15 at 13 35 46

jimniels:
## Relevant issue(s)
_Very narrowly_ fixes #2196 

## Description

Some notes here after researching a solve for this issue:

There is a much larger problem than #2196, which is that we don't have proper "paneling" in the app. The code editor resizing beyond its bounds is a function of the different panels in the app not knowing about each other (and their sizes) when one of them is resized.

Currently there are four different panels that can all be open at the same time:

1. The AI Analyst
2. The grid (x/y axis, cells, sheetbar)
3. Code editor
4. Validations

CleanShot 2025-04-11 at 13 46 02@2x

There are all kinds of bugs that come out of these four different panels competing for space when resizing mechanisms are triggered in the app. And, if you start playing with the four different panels, you're going to find _lots_ of bugs. All of those we'll have to solve at a later time and date.

In the future we'll have lots of things we'll have to think through. For example, we know the grid will always be visible, but that's all we can assume. How will the different panels respond when:

- A new panel is opened or closed?
- An open panel is resized?
- The window is resized?
- Presentation mode is toggled on or off?

Other future considerations:

- What happens on narrower devices (e.g. mobile, tablet, etc)?
- How does the code handle us adding a new panel or removing an existing one?

For later reference, I'm going to link relevant issues that all stem from the underlying problem that we don't have a comprehensive solution to paneling in Quadratic:

- https://github.com/quadratichq/quadratic/issues/2196
- https://github.com/quadratichq/quadratic/issues/1117
- https://github.com/quadratichq/quadratic/issues/1941
- https://github.com/quadratichq/quadratic/issues/1573
- https://github.com/quadratichq/quadratic/issues/449

## Testing considerations

Test _only_ the functionality in #2196, which is: when the AI Analyst is open and you drag the code editor too far to the left, it won't go beyond its bounds. That's it. Don't test anything beyond that. You'll find all kinds of existing bugs that this PR doesn't aim to solve this.

### DIFF
--- a/quadratic-client/src/app/ui/menus/CodeEditor/panels/CodeEditorPanelsResize.tsx
+++ b/quadratic-client/src/app/ui/menus/CodeEditor/panels/CodeEditorPanelsResize.tsx
@@ -1,10 +1,13 @@
+import { showAIAnalystAtom } from '@/app/atoms/aiAnalystAtom';
 import { ResizeControl } from '@/app/ui/components/ResizeControl';
+import { useAIAnalystPanelWidth } from '@/app/ui/menus/AIAnalyst/hooks/useAIAnalystPanelWidth';
 import {
   MIN_WIDTH_PANEL,
   MIN_WIDTH_VISIBLE_GRID,
   useCodeEditorPanelData,
 } from '@/app/ui/menus/CodeEditor/panels/useCodeEditorPanelData';
 import { memo, useEffect, useState } from 'react';
+import { useRecoilValue } from 'recoil';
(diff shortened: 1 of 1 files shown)


ddimaria (approved):
Code looks good, one nit.

PR #2709, opened 2025-04-11
fix: finalize transaction on cancel connection

AyushAgrawal-A2:
## Description
fixes persistent red dot on the kernel menu, and rendering issue, after connection code execution is cancelled


luke-quadratic (approved):
Fix works

ddimaria (approved):

PR #2710, opened 2025-04-11
fix: replace browser confirm dialog with one from the app

jimniels:
## Description

Resolves https://github.com/quadratichq/quadratic/issues/2716

Also: replaces browser-native confirm dialogs with app-specific ones. This makes these dialogs feel like part of the app overall and makes the app feel more holistic.

Before (synchronous browser confirm dialog):

CleanShot 2025-04-11 at 16 05 52@2x

After:

CleanShot 2025-04-21 at 12 18 56@2x

## Testing

Test each dialog that was changed. When you try to do the destructive action, you'll get a dialog where you can either cancel the action or do the action. These dialogs include:

- [ ] Delete file
    - 
- [ ] Delete connection
    - 
- [ ] Delete _another user_ from being invited to a file
    - 
- [ ] Delete _yourself_ from being invited to a file
    - 
- [ ] Delete _another user_ from being a member of a team
    - 
- [ ] Delete _yourself_ from being invited to a file
    -

PR #2721, opened 2025-04-14
feat: require valid connection to create/update

jimniels:
## Relevant issue(s)

Supersedes #2654

## Description

When creating or editing a connection, we now send a test to the connection behind the scenes to validate that the info being entered is correct before saving/updating it.

- If the test fails, you'll see an error and the connection won't be created/updated. This message will give you the option to save the connection anyway.
- If the test succeeds, the connection will be created/updated.

**Editing a connection**

**Creating a new connection**

## To Test

- [x] You **CAN NOT** create or update a connection if a test ping to the connection fails
- [x] You **CAN** override the test failing and saving the connection anyway.
- [x] You **CAN** create or update a connection if a test ping to the connection succeeds


quadratic-client/src/shared/components/connections/ConnectionForm.tsx
```diff
@@ -108,8 +109,31 @@ function ConnectionFormWrapper({ type, props }: { type: ConnectionType; props: C
   const { ConnectionForm, useConnectionForm } = connectionsByType[type];
   const { form } = useConnectionForm(props.connection);
 
+  // This is a middleware that tests the connection before saving
+  const handleSubmitMiddleware = async (formValues: ConnectionFormValues) => {
+    const { name, type, ...typeDetails } = formValues;
+
```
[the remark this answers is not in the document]
jimniels: Nope.

quadratic-client/src/shared/components/connections/ConnectionFormActions.tsx
```diff
@@ -37,95 +32,74 @@ export function ConnectionFormActions({
...
-              </div>
+            {connectionUuid && (
+              <Button
+                type="button"
+                variant="outline-destructive"
+                className="flex-shrink-0"
+                onClick={() => {
+                  const doDelete = window.confirm(
+                    'Deleting this connection will disable it in all existing sheets and it will no longer be usable elsewhere. This cannot be undone.'
+                  );
```
jimniels, line 55: @jimniels update this once #2710 merges

quadratic-client/src/shared/components/connections/ConnectionsSidebar.tsx
```diff
@@ -76,9 +76,8 @@ function SidebarCopyContent({ text }: { text: string }) {
           className="w-full resize-none bg-accent px-3 py-2 font-mono"
           readOnly
           rows={text.split('\n').length}
-        >
-          {text}
-        </textarea>
+          value={text}
+        ></textarea>
```
jimniels, line 80: fixes console warning

CleanShot 2025-05-07 at 13 44 38@2x
ixes

luke-quadratic (approved):
Works great. Expecting this to have a meaningful impact on user connection success, great stuff.

quadratic-client/src/shared/components/connections/ConnectionForm.tsx
```diff
@@ -119,8 +122,38 @@ function ConnectionFormWrapper({
   const { ConnectionForm } = connectionsByType[type];
   const { form } = connectionsByType[type].useConnectionForm(props.connection);
 
+  // This is a middleware that tests the connection before saving
+  const handleSubmitMiddleware: SubmitHandler<ConnectionFormValues> = async (formValues, event) => {
+    // @ts-expect-error
```
ddimaria, line 127: What error is it expecting?

quadratic-client/src/shared/components/connections/ConnectionForm.tsx
```diff
@@ -119,8 +122,38 @@ function ConnectionFormWrapper({
...
+      if (connected === false) {
+        form.setError('root', { message: message ?? 'Unknown error' });
+        return;
+      }
+
+      // If it worked, update the connection
+      props.handleSubmitForm(formValues);
+    } catch (e) {
+      console.error(e);
+      form.setError('root', { message: 'Network error: failed to make connection.' });
```
ddimaria, line 150: I guess this could be a non-network error, but not sure it matters.

quadratic-client/src/shared/components/connections/ConnectionForm.tsx
jimniels: I just changed it to explicitly set the type to `any` because I couldn't figure out the right typescript incantation to get this to understand the type coming from `SubmitHandler<ConnectionFormValues>` which was `event?: React.BaseSyntheticEvent`

ddimaria (approved):
Looks good, just had a question and a nit, but they are non-blocking and can be ignored.

PR #2726, opened 2025-04-15
Rework date parsing

### LINKED ISSUE #2651: Date parsing tracking issue
## Describe the bug
Tracking issue for any time there's a date parsing issue, put an example of what fails into the list: 

1. 6/10/18
2. 4/2025

HactarCE:
This should fix #2651.

We've got completely custom date parsing this time. It handles ordinal numbers (like "Jan 1st") and even has a compile-time setting to select between American and international dates, which we can make configurable at runtime in the future. For now we assume American formats. 🦅 🇺🇸

- 2-digit years less than 50 are assumed to be 20xx while anything 50 or greater is assumed to be 19xx.
- The new parser is based on "ambiguous components" (1- and 2-digit numbers) and "unambiguous components" (ordinal days, named months, and 4-digit years). For example, `10/12/24` is October 12, 2024 (assuming 🇺🇸) but `2010/12/24` is October 24, 2010
- `/`, `-`, `.`, and whitespace are supported as separators. Formats _should_ allow only things that make sense -- e.g., `12/2024` is parsed as December 1, 2024, but `12 2024` and `12-2024` are invalid.


davidfig (approved):

HactarCE:
ah, I updated it with feedback from Luke but forgot to update the test. should be good now!

PR #2730, opened 2025-04-16
fix: enable mouse edges for autocomplete

### LINKED ISSUE #2406: bug: drag to autocomplete doesn't scroll the viewport when you go outside it
Would expect that when I drag to autocomplete, the viewport would scroll as I go outside its bounds so I can drag to the bottom of a column

https://github.com/user-attachments/assets/27fb8aee-c2a4-4bad-bc09-a7d6b83242b0

I would expect dragging left/right/up/down would all work this way.

AyushAgrawal-A2:
## Relevant issue(s)
closes #2406

### DIFF
--- a/quadratic-client/src/app/gridGL/interaction/pointer/PointerAutoComplete.ts
+++ b/quadratic-client/src/app/gridGL/interaction/pointer/PointerAutoComplete.ts
@@ -10,6 +10,7 @@ import { quadraticCore } from '@/app/web-workers/quadraticCore/quadraticCore';
 import type { Point } from 'pixi.js';
 import { Rectangle } from 'pixi.js';
 import { isMobile } from 'react-device-detect';
+import { htmlCellsHandler } from '../../HTMLGrid/htmlCells/htmlCellsHandler';
 
 export type StateVertical = 'expandDown' | 'expandUp' | 'shrink' | undefined;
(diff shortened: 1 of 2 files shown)


luke-quadratic:
This is an improvement but also doesn't work quite how I'd expect. 

Note how there's only a small area where it'll continue scrolling. If I venture into the bottom bar it stops scrolling. 

CleanShot 2025-04-16 at 07 41 07

Note how this would optimally work 

CleanShot 2025-04-16 at 07 42 32

luke-quadratic (changes requested):
This is an improvement, so if it's really difficult to make this work as shown above then we should merge this as is.

AyushAgrawal-A2:
>This is an improvement, so if it's really difficult to make this work as shown above then we should merge this as is.

@luke-quadratic 
mouse edges works over the canvas only, I don't intend to implement scroll when outside canvas in this PR

luke-quadratic:
> > This is an improvement, so if it's really difficult to make this work as shown above then we should merge this as is.
> 
> @luke-quadratic mouse edges works over the canvas only, I don't intend to implement scroll when outside canvas in this PR

Sounds good, changed this to code review needed

luke-quadratic (approved):

jimniels (approved):
Given the intended narrow scope of this fix, this looks 👍  to me

PR #2733, opened 2025-04-16
fix: loading screen

jimniels:
## Description
Fixes [an issue](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1744810749412499) where, if you signed up for the 1st time and came into the app, the loading screen was still showing (fixed upon refresh).

## To Test
- Sign up as a new user
- Loading screen should disappear when you come into the app

Also, any subsequent reloads of the app as a logged-in user should also remove the loading screen once the app loads.

### DIFF
--- a/quadratic-client/src/routes/_root.tsx
+++ b/quadratic-client/src/routes/_root.tsx
@@ -8,8 +8,6 @@ import { ThemeAccentColorEffects } from '@/shared/hooks/useThemeAccentColor';
 import { ThemeAppearanceModeEffects } from '@/shared/hooks/useThemeAppearanceMode';
 import { initializeAnalytics } from '@/shared/utils/analytics';
 import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
-import mixpanel from 'mixpanel-browser';
-import { useEffect } from 'react';
 import type { LoaderFunctionArgs } from 'react-router';
 import { Outlet, useRouteError, useRouteLoaderData } from 'react-router';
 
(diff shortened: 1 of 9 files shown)


quadratic-client/src/routes/_root.tsx
```diff
@@ -83,6 +73,14 @@ export const ErrorBoundary = () => {
...
     />
   );
 };
+
+function useRemoveInitialLoadingUI() {
+  useEffect(() => {
+    document.documentElement.removeAttribute('data-is-loading');
+  }, []);
+  return null;
+}
```
jimniels, line 86: We needed to move this outside of HydrateFallback because that component won't render for a unique one-time flow like signup

Instead, we ensure the initial loading UI will get removed by moving this effect into two places we know will always render: the root component, or it's error state.

quadratic-client/src/shared/components/Empty.tsx
```diff
@@ -20,20 +32,16 @@ export function Empty({
...
+}: EmptyProps) {
+  const [loggedInUser, setLoggedInUser] = useState<User | undefined>(undefined);
   const submit = useSubmit();
 
+  useEffect(() => {
+    if (showLoggedInUser && !loggedInUser) {
+      authClient.user().then((user) => setLoggedInUser(user));
+    }
+  }, [showLoggedInUser, loggedInUser]);
+
```
jimniels, line 44: This component _can_ render in the root where it doesn't have access to the root loader, so we need to get this information (if we have it) async from the auth client.

luke-quadratic (approved):
Works as expected, tested: 
- sign-up
- log-in
- new file creation

AyushAgrawal-A2 (approved):

PR #2737, opened 2025-04-16
fix: fetching users from auth0

### LINKED ISSUE #2734: bug: unexpected error page after 1st sign up
## Describe the bug
Occasionally, some new user signups will see an error immediately after signing up. After a few refreshes, the error screen goes away and they're let into the app.

## Expected behavior
When you login for the first time, you don't see an error screen.

## Reproduction
This bug is difficult to reproduce, but [it has been reported by QA Wolf](https://quadratichq.slack.com/archives/C06Q34HGV8A/p1744751480401119) (they couldn't reproduce) and the side effect of it happening has been [logged in Sentry a number of times](https://quadratic.sentry.io/issues/6094768985/?project=4505110202351616&query=Auth0&referrer=issue-stream&stream_index=0).

[Original slack thread](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1744767286163999)

jimniels:
## Relevant issue(s)
Fixes #2734 

## Description

When the app is being used with auth0, `getUsers` from the auth0 sdk uses their search API endpoint which only guarantees **eventual consistency**.

Because we can be dealing with operations that follow each other in rapid succession (e.g. new user signed up, then logged in) we need to make sure we ask for any missing users from an endpoint that guarantees **immediate consistency**.

Rather than create a separate API like `getUser` in our auth APIs, I took a hybrid approach with our existing `getUsers` auth0 function. It still uses the same call it uses today, but in cases where a user is returned as `undefined`, it then falls back to an auth0 endpoint with immediate consistency. If that then _still_ fails to return a user, then we trigger the same error we were triggering previously and it'll get recorded in sentry.

So in theory this isn't changing how things fundamentally work in prod. For auth0, one extra API call is made when 0 users are being returned. If 0 users are still returned, it functions the same as it does today. But if 1 is found, it should resolve the issues we were seeing previously.

## Testing considerations
This is difficult to simulate because we never found a way to reliably reproduce it.

That said, this code change affects endpoints where we deal with getting users from auth0, so we should test each of these pieces of functionality that they continue to work as expected.

- First time user signup 
- Check education enrollment status (visit `/education/enrollment`)
- Invite someone (by email) to a file
- Invite someone (by email) to a team

Over time, we should see [this exception in Sentry](https://quadratic.sentry.io/issues/6094768985/?project=4505110202351616&query=Auth0&referrer=issue-stream&stream_index=0) go down


quadratic-api/src/auth/auth0.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    const result = await getUsersFromAuth0([
+      { id: 1, auth0Id: 'user1' },
+      { id: 2, auth0Id: 'user2' },
+    ]);
+    expect(Object.keys(result).length).toEqual(2);
+  });
+  it('throws when looking up a user that does not exist', async () => {
+    expect(getUsersFromAuth0([{ id: 3, auth0Id: 'user3' }])).rejects.toThrow();
+  });
+});
```
jimniels, line 29: Added a test that specifies what we expect from auth0, which is that if we provide a user we get one back. Otherwise it throws.

quadratic-api/src/middleware/getTeam.test.ts
```diff
@@ -34,29 +35,18 @@ beforeAll(async () => {
...
-              user_id: 'userNoTeam',
-              email: 'noteam@example.com',
-            },
-          ];
-          return auth0Users.filter(({ user_id }) => q.includes(user_id));
-        }),
-      };
-    }),
-  };
-});
```
jimniels, line 59: We have one centralized function for all these, so I swapped them all out to use it

AyushAgrawal-A2 (approved):

luke-quadratic:
Tested the items that weren't stated as testing considerations and nothing else.

PR #2738, opened 2025-04-16
chore: add version number to all health endpoints

AyushAgrawal-A2:
## Description
- [x] `/health` endpoint of all services will return the current version
- [x] `/health/full` endpoint of all services will return the current version and the health of all dependent services
- [x] CI action to verify current version on deployment


HactarCE (approved):
Looks good!

ddimaria (approved):
Looks good.  You may want to look into [composable actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) to reuse that logic at some point.

PR #2742, opened 2025-04-17
initial commit

jimniels:
## Description

The concerns of the `<Empty>` component began to conflate and so I've split it into two different components:

- `<EmptyPage>` which is used for the pages in the app where nothing renders except a message (e.g. on a crash, a 404 page, etc)
  - This component is the one that reports errors happening, as it takes over the entire page.
- `<EmptyState>` which is used in a variety of places in the app, including `<EmptyPage>`, and represents a state where not content exists. For example, if the app crashes, that's an empty state in an empty page. But also, if a list of files is empty, that's also an empty state but it's wrapped inside the UI of the dashboard.
  - This is a pure component with no side effects like tracking errors, etc. It is solely for display purposes.

Additionally, full screen empty pages now render centered in the screen. I think this is better because that's where your eyes go when the page loads (because that's where the loading UI is). It's weird when, for example, a file fails to load and your eyes go from middle to top in an otherwise blank screen. Now they stay centered on the area of focus.

Before:

https://github.com/user-attachments/assets/da8d6f23-5ce0-405f-af63-811554897202

After:

https://github.com/user-attachments/assets/58c36508-6bab-4793-bcdc-7c5e53368ee2

## To test
Test easily-reproducible views that replaced `<Empty>`  with either `<EmptyPage>` or `<EmptyState>`. This includes:

`<EmptyPage>` (should be an EmptyState centered in the middle of the page (with login info if you're logged in):

- [ ] File 404
- [ ] General 404 page - [link](TODO)
- [ ] Team 404
- [ ] File history - [link]() (try to access this file as a logged in user which is on a team you don't belong to)

`<EmptyState>`

- [ ] Shared files when you have none - [link](TODO)
- [ ] Connections list with 0 connections
- [ ] Files - no files
- [ ] Files - search filter with 0 results

There are more views like this, but they're harder to reproduce easily. If the above work, we can assume the others work too.

### DIFF
--- a/quadratic-client/src/app/ui/QuadraticUI.tsx
+++ b/quadratic-client/src/app/ui/QuadraticUI.tsx
@@ -31,7 +31,7 @@ import { QuadraticSidebar } from '@/app/ui/QuadraticSidebar';
 import { UpdateAlertVersion } from '@/app/ui/UpdateAlertVersion';
 import { useRootRouteLoaderData } from '@/routes/_root';
 import { DialogRenameItem } from '@/shared/components/DialogRenameItem';
-import { Empty } from '@/shared/components/Empty';
+import { EmptyPage } from '@/shared/components/Empty';
 import { ShareFileDialog } from '@/shared/components/ShareDialog';
 import { UserMessage } from '@/shared/components/UserMessage';
 import { COMMUNITY_A1_FILE_UPDATE_URL } from '@/shared/constants/urls';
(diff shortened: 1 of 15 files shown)


luke-quadratic (changes requested):
Support link on general 404 page is not correct - should go to https://www.quadratichq.com/contact

It should probably just be a "Get help" button like on the other 404 pages 

CleanShot 2025-04-22 at 20 18 41@2x

Issue #2739, opened 2025-04-16
Auto-expand on sorted table causes disappearing data


luke-quadratic:
## Describe the bug

Expanding a table that is sorted causes a major issue, see attached vid for recreation (sort table and then try to auto expand on bottom)

![Image](https://github.com/user-attachments/assets/e732fb90-98d4-4234-a9e8-716e5e724ee9)

PR #2746, opened 2025-04-17
bugfix: Auto-expand on sorted table causes disappearing data

ddimaria:
## Relevant issue(s)
Fixes: https://github.com/quadratichq/quadratic/issues/2739

## Description
Expanding a table that is sorted causes a major issue, see attached vid for recreation (sort table and then try to auto expand on bottom)

## (Optional) QA Wolf tests 
@luke-quadratic your call here

## Demo
auto-grow-on-sorted-data-table-bug

### DIFF
--- a/quadratic-core/src/grid/data_table/row.rs
+++ b/quadratic-core/src/grid/data_table/row.rs
@@ -36,7 +36,6 @@ impl DataTable {
     /// Insert a new row at the given index.
     pub fn insert_row(&mut self, row_index: usize, values: Option<Vec<CellValue>>) -> Result<()> {
         let row_index = row_index as i64 - self.y_adjustment(true);
-
         let array = self.mut_value_as_array()?;
         array.insert_row(usize::try_from(row_index)?, values)?;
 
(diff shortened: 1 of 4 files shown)


quadratic-core/src/grid/data_table/row.rs
```diff
@@ -51,7 +50,7 @@ impl DataTable {
             let len = display_buffer.len();
             let index = usize::try_from(row_index)?;
 
-            if index >= len {
+            if index > len {
```
ddimaria, line 53: This was the fix

quadratic-core/src/grid/data_table/row.rs
```diff
@@ -135,16 +134,24 @@ pub mod test {
...
 
-        // expect index out of bounds error
         // first sort the table to create the display buffer
         data_table.sort_column(0, SortDirection::Ascending).unwrap();
+
+        data_table.insert_row(6, None).unwrap();
+        pretty_print_data_table(&data_table, Some("Data Table with second New Row"), None);
+
+        // this should be a 4x7 array
+        let expected_size = ArraySize::new(4, 7).unwrap();
```
ddimaria, line 151: Test added to avoid regressions.

AyushAgrawal-A2 (approved):

jimniels (approved):
From a UX perspective, this works as expected and I was never able to crash anything.

PR #2756, opened 2025-04-18
Allow move column and row

davidfig:
Allows all users to drag columns and rows. 

Known limitations:
* when dragging from a table, it will paste as plain text (same as copy/paste)
* trying to drag a column with a table into another table will fail and remove the content. undoing gets it back to the correct state


AyushAgrawal-A2 (approved):
there is no code to review

luke-quadratic:
Unexpected behavior when dragging the root column of a table. 

CleanShot 2025-04-22 at 17 30 44

luke-quadratic (changes requested):
Feels like above behavior should be blocking.

davidfig:
There's not a great way to fix this for now. We're going to change the internal representation of tables in the near future. When we do, this will be fixable.

luke-quadratic (approved):
This is mergeable, if we're okay with the behavior David F describes above when you move the root cell of a table 

cc @davidkircos https://github.com/quadratichq/quadratic/pull/2756#issuecomment-2822703043

PR #2761, opened 2025-04-19
Ensure checkboxes properly rerender hash

### LINKED ISSUE #2073: Validations checkboxes are improperly positioned after a row resize
https://github.com/user-attachments/assets/54625d60-81f5-40e5-9b28-cddb19af7524

### LINKED ISSUE #2660: Wrapped rows with checkboxes break sheet
## Describe the bug
Copy pasting wrapped rows with checkboxes breaks sheet. Can be recreated in same sheet or in new sheet.

## Reproduction
1. Create some data with wrapped rows in same row as checkboxes 
2. Delete any of that content 
3. Things will be all over the place, if you click into any of these items you're stuck and only recoverable with pressing escape or refresh, can't click out 

Image

davidfig:
## Relevant issue(s)
Fixes #2660
closes #2073

## Description
Fixes a bug where checkboxes did not properly update locations when resizing offsets, copy/pasting, etc..


AyushAgrawal-A2 (approved):

luke-quadratic (changes requested):
This fixes #2073 but does not fully fix #2660; #2660 no longer breaks the sheet but it still has issues.

I can still recreate #2660 with the steps shown in the issue.

You can ignore and mark this PR to only close the other one if you'd like, or fix all in this, up to you. Ready to merge if you want to keep the copy-paste bug open 

CleanShot 2025-04-22 at 19 59 18

luke-quadratic (approved):
Bug from above fixed.

luke-quadratic:
Noting if merged in this state there is one small bug that we should file: 
The checkboxes that are clicked into after pasting don't show the true/false value, but after toggling them they start to show the true/false value (see gif)

CleanShot 2025-04-23 at 09 47 56

PR #2762, opened 2025-04-19
Version refresh

### LINKED ISSUE #2664: Force users to refresh on version update
## Describe the bug
David F: "When you get a chance, please add a bug to the next release that forces users to refresh on upgrade. it doesn't work currently."

davidfig:
## Relevant issue(s)
Fixes #2664

## Description
Ensures that the user receives a refresh message if the multiplayer and client versions do not match. It will wait until the client's /VERSION file matches the server before displaying the message.


quadratic-client/src/app/web-workers/multiplayerWebWorker/multiplayerTypes.ts
```diff
@@ -41,15 +43,10 @@ export interface MultiplayerUser extends MultiplayerUserServer {
...
-  recommendedVersion: number;
-  requiredVersion: number;
-}
-
 export interface ReceiveRoom {
   type: 'UsersInRoom';
   users: MultiplayerUser[];
-  min_version: Version;
+  version: string;
 }
```
AyushAgrawal-A2, line 50: this will break open clients when we ship this change

`private receiveUsersInRoom(room: ReceiveRoom)` function will expect `room.min_version.requiredVersion`
`room` does not have `min_version` property anymore and the existing client will try to access `requiredVersion` of `room.min_version` which will be undefined

AyushAgrawal-A2 (changes requested):
need to discuss as some changes will crash open clients

quadratic-client/src/app/web-workers/multiplayerWebWorker/multiplayerTypes.ts
davidfig: added it back for now

PR #2763, opened 2025-04-19
Fix cells accessed for Python table references

### LINKED ISSUE #2468: [BUG] Code table references missing column headers
## Describe the bug
When you reference a table from Python it should outline the table, because this is what is referenced

## Expected behavior
Would expect the lines to be drawn correctly around exactly what is being referenced

Image

### LINKED ISSUE #2850: not all references show up when using formulas
Image

The second reference is not drawn

davidfig:
## Relevant issue(s)
Fixes #2468
closes #2850

## Description
For Python, we outline the entire table when tables are referenced in q.cells.


AyushAgrawal-A2 (changes requested):
CellsAccessed for formula having TableRef is incorrect

Prod:
image

PR:
image

davidfig:
Fixed JS/Python so it properly shows tables without UI

davidfig:
Please check again.

quadratic-client/src/app/gridGL/UI/cellHighlights/CellHighlights.ts
```diff
@@ -73,27 +72,25 @@ export class CellHighlights extends Container {
...
+              isSelected: this.selectedCellIndex === index,
               range: refRangeBounds,
             });
           }
         });
       });
     }
 
-    this.dirty = true;
+    this.dirty = false;
```
ddimaria, line 93: Does conditional use of `drawDashedRectangle()` need to conditionally trigger dirty?

quadratic-core/src/a1/table_ref/convert.rs
```diff
@@ -1,8 +1,44 @@
...
+        if table.is_html_image {
+            return Some(RefRangeBounds::new_relative(
+                table.bounds.min.x,
+                table.bounds.min.y,
+                table.bounds.min.x,
+                table.bounds.min.y,
+            ));
+        }
+        let (mut y_start, y_end) = table.to_sheet_rows();
+        if self.headers && !self.data {
```
ddimaria, line 29: The conditional after this sets y_start to the same thing, they should be combined into a single branch.

quadratic-core/src/a1/table_ref/convert.rs
```diff
@@ -61,6 +97,16 @@ impl TableRef {
         }
         let y_end = if !self.data { y_start } else { y_end };
 
+        self.finish_convert(table, y_start, y_end, use_unbounded)
+    }
+
+    fn finish_convert(
```
ddimaria, line 103: Can you add documentation to this?

quadratic-core/src/a1/table_ref/convert.rs
```diff
@@ -320,4 +368,145 @@ mod tests {
         let bounds = table_ref.convert_to_ref_range_bounds(false, &context, false, true);
         assert_eq!(bounds, Some(RefRangeBounds::test_a1("A2")));
     }
+
+    #[test]
+    fn test_convert_cells_accessed_to_ref_range_bounds() {
```
ddimaria, line 373: Great tests 💯

ddimaria (approved):
Code looks good, I have some nits and questions.

ddimaria:
@davidfig we use `is_python` in a lot of places here.  Even if the problem we're solving specific to Python, maybe we can generalize this to the aspect of Python that needs special treatment?

davidfig:
is_python is a special case, since we treat column headers than we do in formulas or JS. Hopefully, one day, we can remove this special handling.

PR #2775, opened 2025-04-22
Improve delete tables

davidfig:
## Relevant issue(s)
Combines #2767 and #2678 to fix a bunch of delete issues

- [x] selection accurately highlights tables that will be deleted when pressing delete
- [x] selecting any part of the table name row will result in table selection for data tables and code tables
- [x] with no name ui, code tables can be deleted by (1) context menu, (2) selecting the entire table, or (3) selecting the anchor cell (top-left cell)
- [x] with no name ui, data tables can only be deleted by (1) context menu, or (2) selecting the entire table. Cannot delete a data table by selecting its anchor cell
- [x] can delete entries within a data table if the selection box overlaps the values (this should not work in code tables)


quadratic-core/src/a1/a1_selection/query.rs
```diff
@@ -442,20 +442,44 @@ impl A1Selection {
...
                     }
-                    None
                 }
-                _ => None,
-            })
-            .collect()
+            }),
+        });
+        names.sort();
+        names.dedup();
```
ddimaria, line 481: This would not be necessary if you used a HashSet.

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -241,67 +241,140 @@ impl GridController {
...
+                                && table_column_selection.is_some()
+                                && !data_table.readonly;
+
+                            // we also delete a data table if it is not fully
+                            // selected but any cell in the name ui is selected
+                            if !is_full_table_selected
+                                && data_table.show_ui
+                                && data_table.show_name
+                                // the selection intersects the name ui row
+                                && rect.intersects(Rect::new(
```
ddimaria, line 272: We may have an intersect function that works off of 2 rects (`rect` and `data_table_full_rect`) that would make this more readable.

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -241,67 +241,140 @@ impl GridController {
...
+                                    data_table_full_rect.min.x,
+                                    data_table_full_rect.min.y,
+                                    data_table_full_rect.max.x,
+                                    data_table_full_rect.min.y,
+                                ))
+                                // but the selection does not contain the
+                                // top-left cell (as it will automatically
+                                // delete it in that case)
+                                && !rect.contains(data_table_full_rect.min)
```
ddimaria, line 281: The check above checks for an intersection, which may make this condition not need to be checked.

quadratic-core/src/grid/sheet/data_table.rs
```diff
@@ -478,6 +481,12 @@ impl Sheet {
             .is_some_and(|data_table| data_table.show_name || data_table.show_columns)
     }
 
+    /// Returns true if the cell at pos is a data table cell
+    pub fn is_data_table_cell(&self, pos: Pos) -> bool {
```
ddimaria, line 485: This would be false for code output.  We should come up with a name for "not code output" and apply it in this function name (or at least add info the the fn description).

ddimaria (approved):
Looks good, some small nits and one medium-sized nit.

PR #2792, opened 2025-04-23
fix: new users go to a new file

### LINKED ISSUE #2789: bug: app no longer opens directly to a blank file for new users
## Reproduction
1. Sign up as a new user
2. You are taken to the dashboard instead of in the app

jimniels:
## Relevant issue(s)
Fixes #2789 

Logic: if you're a new user, you come to the root of the app (`/`), and you're not on mobile, we'll create a new file and send you to that.

### DIFF
--- a/quadratic-client/src/routes/login-result.tsx
+++ b/quadratic-client/src/routes/login-result.tsx
@@ -1,5 +1,6 @@
 import { authClient } from '@/auth/auth';
 import { apiClient } from '@/shared/api/apiClient';
+import { isMobile } from 'react-device-detect';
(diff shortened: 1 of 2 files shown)


davidkircos (changes requested):
This used to create a team file by default, then this Create a File was checked
image
Now it is a private file, and this isn't marked as checked
I think the user would be confused as to where their file went

jimniels:
Good catch @davidkircos 

I've updated this so now when a new user signs up and they're sent to a new file, it's a team file.

ddimaria (approved):

PR #2795, opened 2025-04-23
chore: switch to claude on bedrock

AyushAgrawal-A2:
## Description
switch to using claude on bedrock, gcp is throwing `RESOURCE_EXHAUSTED` error


luke-quadratic (changes requested):
I'm getting thinking mode on by default in this branch, was that on purpose?

luke-quadratic (approved):
Thinking mode as default was on purpose, cheaper but better performance.

PR #2803, opened 2025-04-24
[WIP] Fine-tuned 4.1 

luke-quadratic:


luke-quadratic:
This branch is only used locally 

See #2839 for preview

PR #2804, opened 2025-04-24
Resize multiple columns and rows

### LINKED ISSUE #2205: "Resize column to fit data" does not work on multiple columns
Steps to reproduce:
1. Type `a` in A1, `b`, and B1, and `c` in C1
2. Select `A:C`
3. Open the command palette
4. Select "resize column to fit data"

Expected behavior: all columns shrink

Actual behavior: only the leftmost column shrinks

Ideally this should also work if resizing a column by double-clicking. See Excel's behavior:

https://github.com/user-attachments/assets/0ba9f3df-92fc-4075-b003-2df29a80abda

The full logic is: if the resized column is within the selection, then all selected columns are affected.

(And of course the same should work for rows.)

davidfig:
## Relevant issue(s)
Closes #2205

- [x] resize multiple columns and rows
- [x] double-click on resize handle to auto fit size for arbitrarily selected columns and rows
- [x] resizing when select-all will (1) change all rows/columns; and (2) change the default row/column size
- [x] inserting a col or row sets the col/row size to the current selection's size

Demo:

https://github.com/user-attachments/assets/1711b8b9-6d28-4fc1-b432-2c1e7a5ff839


quadratic-client/src/app/web-workers/quadraticCore/coreClientMessages.ts
```diff
@@ -1275,6 +1276,48 @@ export interface CoreClientCoreError {
...
+  cursor: string;
+}
+
+export interface ClientCoreResizeAllRows {
+  type: 'clientCoreResizeAllRows';
+  sheetId: string;
+  size: number;
+  cursor: string;
+}
+
```
AyushAgrawal-A2, line 1320: duplicate

luke-quadratic (approved):
This feels awesome. 

I get nervous when nothing is found in a big PR review so asking that @davidkircos please take a solid look at this before merging.

quadratic-client/src/app/web-workers/quadraticCore/quadraticCore.ts
```diff
@@ -212,6 +213,9 @@ class QuadraticCore {
       return;
     } else if (e.data.type === 'coreClientCoreError') {
       events.emit('coreError', e.data.from, e.data.error);
+      if (debug) {
```
ddimaria, line 216: Why wouldn't we always want this in the console?

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
```diff
@@ -190,6 +191,159 @@ impl GridController {
             }
         }
     }
+
+    pub fn execute_resize_columns(&mut self, transaction: &mut PendingTransaction, op: Operation) {
+        if let Operation::ResizeColumns {
+            sheet_id,
+            column_widths: column_heights,
```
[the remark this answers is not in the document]
ddimaria: Or is `column_widths` off here?

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
```diff
@@ -190,6 +191,159 @@ impl GridController {
...
+                })
+                .collect();
+
+            old_column_heights.sort_by_key(|JsColumnWidth { column, .. }| *column);
+
+            if old_column_heights == column_heights {
+                return;
+            }
+
+            transaction
```
ddimaria, line 223: I think this is just 
```rust
transaction.forward_operations.push(op);
```

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
```diff
@@ -190,6 +191,159 @@ impl GridController {
...
+                });
+
+            transaction
+                .reverse_operations
+                .push(Operation::ResizeColumns {
+                    sheet_id,
+                    column_widths: old_column_heights,
+                });
+
+            if (cfg!(target_family = "wasm") || cfg!(test)) && !transaction.is_server() {
```
ddimaria, line 237: Consider moving this to the the wasm dir.

ddimaria (approved):
Looks great!  I have some questions/comments, but ✅ unless their answers block.

quadratic-client/src/app/web-workers/quadraticCore/quadraticCore.ts
davidfig: We're already reporting it via sentry. No need to report it twice.

PR #2810, opened 2025-04-24
Handle nulls of all types for postgres and mysql

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2241

## Description
Many types did not support nulls properly.

## Testing considerations
Test null values of various data types in Postgres and MySQL

## Database with nulls
Screenshot 2025-04-24 at 5 47 02 PM

## Grid output of nulls
image


luke-quadratic:
I'm seeing successful behavior here when there's only one row with a null, but as soon as there are multiple the queries fail. 

Possible this is something I'm doing wrong, please confirm on your end if you can recreate. 

CleanShot 2025-04-25 at 13 38 07

luke-quadratic (changes requested):
see comment above

ddimaria:
@luke-quadratic can you revalidate this?

luke-quadratic (approved):
Prior bugs fixed. 

I tested: 
- Works with table with nulls across multiple columns and rows. 
- Tested a prior working big table, worked as expected

quadratic-rust-shared/src/arrow/arrow_type.rs
```diff
@@ -48,57 +51,112 @@ pub enum ArrowType {
     Json(Value),
     Jsonb(Value),
     Void,
+    Null,
     Unsupported,
 }
 
+impl From<&ArrowType> for DataType {
```
HactarCE, line 58: Could this be a method `ArrowType::ty()` or `ArrowType::data_type()`? `From<&T>` feels a bit strange to me.

I would expect `ArrowType` to be called `ArrowValue`

quadratic-rust-shared/src/sql/mod.rs
```diff
@@ -167,3 +168,21 @@ pub trait Connection {
...
+/// else return a null value
+#[macro_export]
+macro_rules! to_arrow_type {
+    ( $arrow_type:path, $kind:ty, $row:ident, $index:ident ) => {{
+        match convert_sqlx_type!($kind, $row, $index) {
+            Some(value) => $arrow_type(value),
+            None => ArrowType::Null,
+        }
+    }};
+}
```
HactarCE, line 188: Good idea to use macros for this

HactarCE (approved):
Good cleanup! just one nit, feel free to ignore if you disagree

quadratic-rust-shared/src/arrow/arrow_type.rs
ddimaria: Good feedback!  I updated this to be `ArrowType::data_type()`

PR #2816, opened 2025-04-25
AI Paywall

davidkircos:
Snow a paywall instead of wait ui

CleanShot 2025-04-25 at 12 10 41@2x


AyushAgrawal-A2 (approved):

luke-quadratic (approved):
Only note is that the learn more and pay buttons go to the same place, consider combining

PR #2817, opened 2025-04-25
chore: change ai default, fix thinking state on update

AyushAgrawal-A2:
## Description
- change ai default to thinking off
- fix think toggle state on update


luke-quadratic (approved):
Works as expected

PR #2825, opened 2025-04-26
chore: indicate paid users in feedback

jimniels:
## Description
tk

### DIFF
--- a/quadratic-api/src/routes/v0/feedback.POST.ts
+++ b/quadratic-api/src/routes/v0/feedback.POST.ts
@@ -27,6 +27,23 @@ async function handler(req: RequestWithUser, res: express.Response) {
     },
   });
 
+  // See if the user has paid team(s)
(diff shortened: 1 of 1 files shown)


davidkircos (approved):
nice!

luke-quadratic:
Don't believe this can be user tested since Feedback button won't trigger from non-prod branch since I assume Slack webhook is configured for prod? 

This is low stakes so I assume can just test in prod.

jimniels:
Yes, that's right. To get this to work, it requires an env variable `SLACK_FEEDBACK_URL`.

I tested this locally by pointing the `SLACK_FEEDBACK_URL` env var at my own private slack channel and it worked. We would have to update the preview branch to with the env variable to get it to work. I _think_ QA has this env variable set (or at least it did at one point) so maybe once it's in staging you'll see it? Otherwise, yeah it'll have to go to prod — which is probably fine. It's pretty low stakes

PR #2834, opened 2025-04-28
Hotfix index out of bounds issue

ddimaria:
Fixes datadog alerts:

```text
Time | Host
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
index out of bounds: the len is 14 but the index is 14
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
thread 'tokio-runtime-worker' panicked at /quadratic/quadratic-core/src/controller/execution/run_code/mod.rs:131:18:
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
index out of bounds: the len is 63 but the index is 63
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
thread 'tokio-runtime-worker' panicked at /quadratic/quadratic-core/src/controller/execution/run_code/mod.rs:131:18:
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
index out of bounds: the len is 1 but the index is 1
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
thread 'tokio-runtime-worker' panicked at /quadratic/quadratic-core/src/controller/execution/run_code/mod.rs:131:18:
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
index out of bounds: the len is 3 but the index is 3
-----------------------------
20:46:22 UTC | i-0d68e8d685a933ff7
thread 'tokio-runtime-worker' panicked at /quadratic/quadratic-core/src/controller/execution/run_code/mod.rs:131:18:
-----------------------------
20:45:02 UTC | i-0d68e8d685a933ff7
index out of bounds: the len is 63 but the index is 63
-----------------------------
20:45:02 UTC | i-0d68e8d685a933ff7
thread 'tokio-runtime-worker' panicked at /quadratic/quadratic-core/src/controller/execution/run_code/mod.rs:131:18
```

PR #2835, opened 2025-04-28
fix: console.warn messages from react-router

jimniels:
## Description
Sentry is blowing up now with `console.warn` messages from react-router that aren't useful.

CleanShot 2025-04-28 at 09 53 01@2x

This fixes the router so it's explicitly doing what its already implicitly doing (and therefore it won't log these errors.

## (Optional) Testing considerations

- Nothing to test here, other than ship it and see sentry errors about missing leaf routes disappear


jimniels:
Closing as fix is being included in #2834

jimniels:
Closing. Fixed in #2834

PR #2836, opened 2025-04-28
ai uses the current sheet when request was made

### LINKED ISSUE #2441: Lock AI prompts to the sheet it was asked in
Currently if you switch sheet mid prompt the AI will place the output in the newly selected sheet. It should insert in whatever sheet the prompt was submitted in. Once generated you should get moved to wherever the data/code is placed.

davidfig:
## Relevant issue(s)
Fixes #2441

### DIFF
--- a/quadratic-client/src/app/ai/tools/aiToolsActions.ts
+++ b/quadratic-client/src/app/ai/tools/aiToolsActions.ts
@@ -108,8 +108,8 @@ export const aiToolsActions: AIToolActionsRecord = {
     return `Executed set chat name tool successfully with name: ${args.chat_name}`;
   },
   [AITool.AddDataTable]: async (args) => {
-    const { sheetId, top_left_position, table_name, table_data } = args;
-
+    const { sheet_name, top_left_position, table_name, table_data } = args;
+    const sheetId = sheets.getSheetByName(sheet_name)?.id ?? sheets.current;
     try {
       const selection = stringToSelection(top_left_position, sheetId, sheets.a1Context);
       if (!selection.isSingleSelection()) {
@@ -139,7 +139,8 @@ export const aiToolsActions: AIToolActionsRecord = {
     }
   },
   [AITool.SetCellValues]: async (args) => {
-    const { sheetId, top_left_position, cell_values } = args;
+    const { sheet_name, top_left_position, cell_values } = args;
+    const sheetId = sheets.getSheetByName(sheet_name)?.id ?? sheets.current;
     try {
       const selection = stringToSelection(top_left_position, sheetId, sheets.a1Context);
       if (!selection.isSingleSelection()) {
@@ -161,7 +162,8 @@ export const aiToolsActions: AIToolActionsRecord = {
     }
   },
   [AITool.SetCodeCellValue]: async (args, messageMetaData) => {
-    let { sheetId, code_cell_language, code_string, code_cell_position } = args;
+    let { sheet_name, code_cell_language, code_string, code_cell_position } = args;
+    const sheetId = sheets.getSheetByName(sheet_name)?.id ?? sheets.current;
     try {
       const selection = stringToSelection(code_cell_position, sheetId, sheets.a1Context);
       if (!selection.isSingleSelection()) {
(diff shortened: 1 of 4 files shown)


luke-quadratic (changes requested):
Looks like the set code cell value tool is not using ensure rect visible correctly. It's panning down on the new sheet and not switching back. 

Noting that it's working correctly for add data table tool.

The PR fixes the core issue though, awesome! 

Up to you whether or not the fix I'm requesting here belongs in this PR. Please open a new issue or ping me to open a new issue if you do not apply the fix here. 

https://github.com/user-attachments/assets/d81f9f6a-7123-4b32-a893-9c0297c3486d

davidfig:
This is fixed. BTW: this only moves the viewport and not the cursor to the updated region. Open to change this.

luke-quadratic:
> This is fixed. BTW: this only moves the viewport and not the cursor to the updated region. Open to change this.

I think the current behavior is correct: 
- Cursor gets refocused on new creation if it's moving sheets. 
- Cursor doesn't get moved if it's in the same sheet. 

This feels correct.

luke-quadratic (approved):
This is great.

quadratic-client/src/app/ai/tools/aiToolsActions.ts
```diff
@@ -139,18 +139,19 @@ export const aiToolsActions: AIToolActionsRecord = {
     }
   },
   [AITool.SetCellValues]: async (args) => {
-    const { top_left_position, cell_values } = args;
+    const { sheet_name, top_left_position, cell_values } = args;
+    const sheetId = sheets.getSheetByName(sheet_name)?.id ?? sheets.current;
```
ddimaria, line 143: This is used so many places, it might be worth abstracting: `getSheetId()`

quadratic-client/src/app/gridGL/interaction/viewportHelper.ts
```diff
@@ -65,7 +65,10 @@ export function isColumnVisible(column: number): boolean {
 }
 
 // Makes a rect visible in the viewport
-export function rectVisible(min: JsCoordinate, max: JsCoordinate): boolean {
+export function rectVisible(sheetId: string, min: JsCoordinate, max: JsCoordinate): boolean {
+  if (sheetId !== sheets.current) {
+    sheets.current = sheetId;
```
ddimaria, line 70: This global mutation might cause issues in the future (e.g. race conditions) if it happens elsewhere.

quadratic-shared/ai/specs/aiToolsSpec.ts
```diff
@@ -194,12 +203,12 @@ Always prefer using this tool to add structured data to the current open sheet.
...
         },
       },
-      required: ['top_left_position', 'table_name', 'table_data'],
+      required: ['sheet_name', 'top_left_position', 'table_name', 'table_data'],
       additionalProperties: false,
     },
     responseSchema: AIToolsArgsSchema[AITool.AddDataTable],
     prompt: `
-Adds a data table to the current open sheet, requires the top_left_position (in a1 notation), the name of the data table and the data to add. The data should be a 2d array of strings, where each sub array represents a row of values.\n
+Adds a data table to the current sheet defined in the context, requires the sheet name, top_left_position (in a1 notation), the name of the data table and the data to add. The data should be a 2d array of strings, where each sub array represents a row of values.\n
```
ddimaria, line 211: @luke-quadratic @AyushAgrawal-A2 you might want to verify this.

ddimaria: There are more in the PR as well.

ddimaria (approved):
Looks good, but would like @luke-quadratic and/or @AyushAgrawal-A2 to check out the copy.

luke-quadratic:
> Looks good, but would like @luke-quadratic and/or @AyushAgrawal-A2 to check out the copy.

👍

luke-quadratic (approved):
approved on functionality and quickly scanned tool changes

AyushAgrawal-A2 (approved):

Issue #2448, opened 2025-03-07
bug: paste formatting not working predictably


jimniels:
Paste formatting doesn't work in tables. Here's a video from prod copying a column and pasting in the one next to it.

https://github.com/user-attachments/assets/06ab0d4e-df77-4cce-a841-e9533c2be57e

In seems like "paste formatting" in tables isn't the only issue. Past formatting outside of tables is a little funky too.

For example: copy a column with formatting, and paste it. Formatting is preserved the first few times, but then gets lost.

https://github.com/user-attachments/assets/cc28556f-1a9d-40ed-8639-6a07eb053a8a

Also: when you paste formatting outside of a table, it overwrites the data in the first cell when applying the formatting.

https://github.com/user-attachments/assets/c8a5daf9-ac57-4465-9d8d-f73d80f956c3

jimniels:
Marking as high priority as it seems like a pretty straightforward fix to tables.

davidkircos:
Could be done with #2448 similar issue

ddimaria:
I cannot duplicate this.

jimniels:
@ddimaria one of the issues in the original description seems to have been resolved. However, there was a second issue around paste formatting that was still a problem. In reproducing it, I found a few other issues. I've updated the bug description to reflect this.

PR #2847, opened 2025-04-29
Combine sheet and data table formats when copying

ddimaria:
## Relevant issue(s)
Fixes: https://github.com/quadratichq/quadratic/issues/2448

## Description
Copying any format within a data table wasn't working.  We now can copy formats in a data table, as well as have a selection that overlaps any number of tables and non-table formats and properly copies formats in the selection.

## Testing considerations
* Copy a format from the sheet
* Copy a format from a data table
* Copy formats that overlap the sheet and data table


quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -714,140 +743,114 @@ impl GridController {
...
-
-                // collect all cell values, values and sheet format updates for a a single operation
-                let mut cell_values =
-                    CellValues::new(cell_value_width as u32, cell_value_height as u32);
-                let mut formats = clipboard.formats.to_owned().unwrap_or_default();
-                let mut borders = clipboard.borders.to_owned().unwrap_or_default();
-                let source_columns = clipboard.cells.columns;
+        let mut ops = vec![];
+        let mut compute_code_ops = vec![];
+        let mut clipboard = Clipboard::decode(&html)?;
```
ddimaria, line 748: This decode line is the only real new code in this file, the rest is just moving the decode to a separate function to be re-used in tests.  All other diffs in this file aren't actual diffs.

quadratic-core/src/grid/data_table/formats.rs
```diff
@@ -52,17 +52,30 @@ impl DataTable {
         sheet_format_updates: &mut SheetFormatUpdates,
     ) -> Result<()> {
         for x in formats_rect.x_range() {
-            let format_display_x = u32::try_from(x - data_table_pos.x)?;
+            let relative_x = x - data_table_pos.x;
+
+            // ignore non-overlapping areas
```
ddimaria, line 57: @AyushAgrawal-A2 I added some checks in the function so that the passed in rect can be to the left or above the data table w/o generating errors (just the overlapping rect).

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -108,6 +108,33 @@ pub struct Clipboard {
     pub operation: ClipboardOperation,
 }
 
+impl Clipboard {
+    /// Decode the clipboard html and return a Clipboard struct.
+    pub fn decode(html: &str) -> Result<Self> {
```
ddimaria, line 113: The `drop()` invocations were no longer needed.

luke-quadratic (approved):
Looks great, tested what was recommended: 
- Copied from inside to outside data table 
- Copied from outside data table to inside data table 
- Put data table on top of existing formatting and added formatting 

Plus: 
- Cut in same scenarios 
- Cut/copy entire table 
- Semi-overlaps 
- Full overlaps

davidfig (approved):

PR #2854, opened 2025-04-30
Test more rows in context 

luke-quadratic:
Testing the AI performance hit from showing 75 rows per table by default 

I'm seeing negligible performance hits at 75 rows across datasets where the 75 are numerical, heavy text, etc.

### DIFF
--- a/quadratic-core/src/grid/sheet/ai_context.rs
+++ b/quadratic-core/src/grid/sheet/ai_context.rs
@@ -215,7 +215,7 @@ impl Sheet {
                 1,
             );
             let first_row_visible_values = self
-                .js_cell_value_pos_in_rect(first_row_rect, Some(1))
+                .js_cell_value_pos_in_rect(first_row_rect, Some(75))
                 .into_iter()
                 .flatten()
                 .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


quadratic-core/src/grid/sheet/ai_context.rs
```diff
@@ -51,7 +51,7 @@ impl Sheet {
                 rect_origin: tabular_data_rect.min.a1_string(),
                 rect_width: tabular_data_rect.width(),
                 rect_height: tabular_data_rect.height(),
-                starting_rect_values: self.js_cell_value_pos_in_rect(tabular_data_rect, Some(3)),
+                starting_rect_values: self.js_cell_value_pos_in_rect(tabular_data_rect, Some(75)),
```
ddimaria, line 54: It might be useful to assign the magic number to a descriptive variable and use throughout.

ddimaria (approved):
Looks good, one nit.

quadratic-core/src/grid/sheet/ai_context.rs
```diff
@@ -212,10 +212,10 @@ impl Sheet {
                 bounds.min.x,
                 bounds.min.y + table.y_adjustment(false),
                 bounds.width() as i64,
-                1,
+                75,
```
AyushAgrawal-A2, line 215: this is incorrect, use table bounds

AyushAgrawal-A2 (changes requested):

luke-quadratic:
I don't like this approach if we don't need it. 

I think #2862 is a great start and should be the focus over this. Closing.

PR #2856, opened 2025-04-30
feat: suggest users start a new chat

### LINKED ISSUE #2809: Suggest users clear their chat
[@jimniels ping for discussion whenever you're ready to look at this, have some various ideas and inspo to draw from]

## Description 

When chats get long users aren't necessarily aware they need to clear the chat to improve model performance. We need to encourage users to take an action to start a new chat as they get too long.

jimniels:
## Relevant issue(s)
Resolves #2809 

### Start a new chat message

If a user crosses a threshold, show a message educating them that they'll get better results with new chats.

The threshold is set at `15` in the code, but also allows overriding via the env var `VITE_AI_ANALYST_START_NEW_CHAT_MSG_THRESHOLD`

### Chat history
Remind users they have a chat history when they start a new chat (and they have at least 1 already).

Once this history button has been clicked `15` times or more, this message disappears.

### Events

To better track information around usage, we'll add mixpanel events for:

- When the user clicks the 'new chat' button, how many of (their) messages are currently in the chat?
- When the user submits a prompt, what's the total number of (their) messages in the current chat?


jimniels:
Ok here's where we're at after further discussion with @luke-quadratic 

- It feels a lot easier to start with A) don't allow this to be dismissible, and B) tend towards a higher threshold. Persisting extra data is more work and we may change our minds in 6 months.
- I will investigate softening the warning a in light of the decision, like maybe changing the language or the visual style  so it's not so prominent (since you can't dismiss it).
- I will add events to track A) how often this pops up, B) how often people click start a new chat when it's there, and C) how often people continue to submit a new prompt even when it's there.

Starting at 15 chats and making it not dismissible is the most minimum viable version of this. It requires not additional changes to data structures, etc. It's purely UI. We ship that, monitor, measure, and re-evaluate if we want to make this more sophisticated in the future, like taking token count into effect or making the message dismissible.

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
```diff
@@ -12,73 +12,95 @@ import { AddIcon, CloseIcon, HistoryIcon } from '@/shared/components/Icons';
...
-        </TooltipPopover>
+        <div className="flex items-center gap-2">
+          <TooltipPopover label="New chat">
+            <Button
+              variant={showStartFreshMsg ? 'outline' : 'ghost'}
+              size="icon-sm"
+              className="text-muted-foreground hover:text-foreground"
+              disabled={loading || currentUserMessages === 0}
+              onClick={() => {
+                console.log('mixpanel startNewChat with: ', currentUserMessages);
```
ddimaria, line 52: `console.log` here

jimniels: thank you — removed!

ddimaria (approved):
The code looks good, just one nit regarding a console.log statement.

quadratic-client/src/app/atoms/aiAnalystAtom.ts
```diff
@@ -315,6 +315,11 @@ export const aiAnalystCurrentChatMessagesCountAtom = selector<number>({
   get: ({ get }) => get(aiAnalystCurrentChatAtom).messages.length,
 });
 
+export const aiAnalystCurrentChatUserMessagesCountAtom = selector<number>({
+  key: 'aiAnalystCurrentChatUserMessagesCountAtom',
+  get: ({ get }) => get(aiAnalystCurrentChatAtom).messages.filter((message) => message.role === 'user').length,
+});
+
```
AyushAgrawal-A2, line 322: this will include tool call, is that intentional ?

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
```diff
@@ -12,73 +12,94 @@ import { AddIcon, CloseIcon, HistoryIcon } from '@/shared/components/Icons';
...
 export const AIAnalystHeader = memo(({ textareaRef }: AIAnalystHeaderProps) => {
   const [showChatHistory, setShowChatHistory] = useRecoilState(aiAnalystShowChatHistoryAtom);
   const chatsCount = useRecoilValue(aiAnalystChatsCountAtom);
   const setCurrentChat = useSetRecoilState(aiAnalystCurrentChatAtom);
-  const messagesCount = useRecoilValue(aiAnalystCurrentChatMessagesCountAtom);
+  const currentUserMessages = useRecoilValue(aiAnalystCurrentChatUserMessagesCountAtom);
   const setShowAIAnalyst = useSetRecoilState(showAIAnalystAtom);
   const loading = useRecoilValue(aiAnalystLoadingAtom);
 
+  const showStartFreshMsg = currentUserMessages > THRESHOLD_START_FRESH_MSG && !showChatHistory;
```
AyushAgrawal-A2, line 33: useMemo

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
```diff
@@ -12,73 +12,94 @@ import { AddIcon, CloseIcon, HistoryIcon } from '@/shared/components/Icons';
...
+          Long chat? New topic? Fresh chats = better results.
+          <span className="absolute -top-2 right-[86px] h-0 w-0 border-b-8 border-l-8 border-r-8 border-b-foreground border-l-transparent border-r-transparent" />
+        </p>
+      )}
+      {showHistoryMsg && (
+        <p className="relative mx-2 mb-1.5 rounded bg-secondary px-2 py-1.5 text-center text-xs text-muted-foreground">
+          Previous chats are saved in history.
+          <span className="absolute -top-2 right-[50px] h-0 w-0 border-b-8 border-l-8 border-r-8 border-b-secondary border-l-transparent border-r-transparent" />
+        </p>
+      )}
```
AyushAgrawal-A2, line 102: should be dismissable, hate this

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystUserMessageForm.tsx
```diff
@@ -29,21 +30,22 @@ export const AIAnalystUserMessageForm = memo(
...
     const [context, setContext] = useState<Context>(initialContext ?? defaultAIAnalystContext);
+    const userMessagesCount = useRecoilValue(aiAnalystCurrentChatUserMessagesCountAtom);
     const waitingOnMessageIndex = useRecoilValue(aiAnalystWaitingOnMessageIndexAtom);
     const delaySeconds = useRecoilValue(aiAnalystDelaySecondsAtom);
     const { submitPrompt } = useSubmitAIAnalystPrompt();
 
     const handleSubmit = useCallback(
       ({ content, onSubmit }: SubmitPromptArgs) => {
-        mixpanel.track('[AIAnalyst].submitPrompt');
+        mixpanel.track('[AIAnalyst].submitPrompt', { userMessageCountUponSubmit: userMessagesCount });
```
AyushAgrawal-A2, line 40: this will also include tool calls, is this intentional ?

AyushAgrawal-A2 (dismissed):

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
AyushAgrawal-A2: there is a helper function to calculate this

quadratic-client/src/app/atoms/aiAnalystAtom.ts
jimniels: No. Or at least, I don't think so? 

The intent is for this to be a measurement of how many messages in the chat are from the user themselves? e.g. in this screenshot, you would see 4 messages. It's possible I am not getting that info from the right spot...are `message.role` types of `user` also tool calls?

CleanShot 2025-05-05 at 16 25 32@2x

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
jimniels: See [this comment](https://github.com/quadratichq/quadratic/pull/2856#issuecomment-2848197024) for rationale. tldr; this is the most minimum viable version of this feature that is pure UI (requires no extra changes to data structures, etc). Idea is we ship this, monitor for a bit, and re-evaluate how (if at all) we want to change this.

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystUserMessageForm.tsx
jimniels: Per [this comment](https://github.com/quadratichq/quadratic/pull/2856#discussion_r2074293283), tool call is not intentional. Intent is to get "how many user prompts are on the screen" however we best do that

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
AyushAgrawal-A2: I would prefer to not do something than do it wrong, this is so wrong.
we should remind / educate / nudging the user to do something and not force / annoy / punch him in the face to do it

quadratic-client/src/app/atoms/aiAnalystAtom.ts
AyushAgrawal-A2: change to 
```
(message) => message.role === 'user' && message.contextType === 'userPrompt'
```

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystUserMessageForm.tsx
jimniels: Update this to pull the right data from the atom, [per this recommendation](https://github.com/quadratichq/quadratic/pull/2856#discussion_r2074645844)

quadratic-client/src/app/ui/menus/AIAnalyst/AIAnalystHeader.tsx
AyushAgrawal-A2: I still disagree, don't want to waste time discussing this further

luke-quadratic (approved):
This works as expected - I'm wishing it were popped much earlier to educate users since 15 is a rarely breached amount, but this can always be changed later

Issue #2466, opened 2025-03-10
Tool call for AI to view more data


luke-quadratic:
[Let's discuss to finalize details before starting this work] 

## Basic description/background
Today we always pass 3 rows of data by default. Sometimes to answer prompts sufficiently requires accessing more than the first three rows. To solve this we can add a tool call for the AI to grab x number of extra rows as necessary. This may require successive tool calls when the data is too large to shove everything into context. 

## Wireframing, in-depth details, docs, etc.
**Questions from Ayush:**
"Max data queryable" - enforce an arbitrary amount that we can easily configure in code for testing, start with 1k cells max (we can use anthropic token counter if necessary to ensure tokens aren't too high, if tokens too high shrink a lot, otherwise continue) 
"Max tool call iterations" - start with an arbitrary limit like 3 that we can easily configure in code to test and refine 
"Maintain queried context or drop immediately?" - make easily configurable so we toggle a setting to test with/without
"Analyst, assistant, both" - ideally by the time we've started work on this assistant has been retired by #2263 
"UI for context tool call" - same as existing tool calls, put whatever icon for now and we will replace in PR later

Tool call should be able to function for tabular data and code outputs 

## Open questions 

Is it worth doing something more complex with token counting? E.g. grab x # of rows, count tokens, if tokens way lower than available context window then increase rows being sent per tool call? If tokens higher than context window then reduce and try again?

PR #2862, opened 2025-05-02
Get cells for ai

davidfig:
closes https://github.com/quadratichq/quadratic/issues/2466

Provides an AI tool call to get cells via a selection and sheet name.

- [x] get cell tool call
- [x] use get cell tool call when manipulating or adding existing data on the sheet that is not in a table
- [x] when using set_cell_values, column headers should be placed above the first row of data
- [x] when adding a data tables, check if there is sufficient room using get_cell_values


luke-quadratic:
Intended behavior to grab viewport before doing things? 
CleanShot 2025-05-05 at 09 06 34@2x

luke-quadratic:
AI is feeling much more confident with doing calculations itself now that it knows what the data is. 

We need to doubly enforce that it should use formulas and code to solve problems, not do math and insert the raw values itself. 

CleanShot 2025-05-06 at 16 14 23@2x

PR #2863, opened 2025-05-02
Fix insert next to code

### LINKED ISSUE #2845: Inserting column next to SQL table deletes content
## Describe the bug
Inserting column next to SQL table deletes content

This seems to only happen if you do insert column right from the column directly left of a SQL table 

## Expected behavior
Column would get cleanly inserted and not delete SQL content 

## Reproduction

https://github.com/user-attachments/assets/250a4a01-5c1f-42e8-acd3-5e18c7209883

davidfig:
## Relevant issue(s)
Fixes #2845 

## Description
Fixes a bug where inserting a column to the left of a code table causes unexpected behavior.

### DIFF
--- a/quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
+++ b/quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
@@ -306,9 +306,8 @@ mod tests {
             CellsAccessed, CodeCellLanguage, CodeCellValue, CodeRun, DataTable, DataTableKind,
             sheet::validations::{validation::Validation, validation_rules::ValidationRule},
         },
-        test_util::{
-            assert_data_table_size, first_sheet_id, test_create_html_chart, test_create_js_chart,
-        },
+        test_create_gc,
+        test_util::*,
         wasm_bindings::js::{clear_js_calls, expect_js_call_count, expect_js_offsets},
     };
 
@@ -1081,5 +1080,14 @@ mod tests {
 
         gc.undo(None);
         assert_data_table_size(&gc, sheet_id, pos![B2], 3, 3, false);
+    }
+
+    #[test]
+    fn test_insert_col_next_to_code() {
+        let mut gc = test_create_gc();
+        let sheet_id = first_sheet_id(&gc);
+
+        test_create_formula_array(&mut gc, pos![C2], 2, 2);
+        print_first_sheet(&gc);
     }
 }
--- a/quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
+++ b/quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
@@ -1087,7 +1087,12 @@ mod tests {
         let mut gc = test_create_gc();
         let sheet_id = first_sheet_id(&gc);
 
-        test_create_formula_array(&mut gc, pos![C2], 2, 2);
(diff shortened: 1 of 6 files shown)


luke-quadratic (approved):
fix works, nice one

quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
```diff
@@ -1082,4 +1081,31 @@ mod tests {
...
+        let mut gc = test_create_gc();
+        let sheet_id = first_sheet_id(&gc);
+
+        let table = test_create_data_table(&mut gc, sheet_id, pos![C2], 2, 2);
+
+        print_first_sheet(&gc);
+        gc.insert_column(sheet_id, 3, false, None);
+
+        // todo: this is wrong, the data table should be at c2, not d2
+        assert_data_table(&gc, pos![sheet_id!c2], &table);
```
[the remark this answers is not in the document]
ddimaria: @davidfig what do you think ☝️?

quadratic-core/src/test_util/assert_data_table.rs
```diff
@@ -56,6 +57,28 @@ pub fn assert_chart_size(
     assert_data_table_size(gc, sheet_id, pos, width, height, include_ui);
 }
 
+/// Run an assertion that a data table is equal to the given data table.
+#[cfg(test)]
+#[track_caller]
+pub fn assert_data_table(gc: &GridController, sheet_pos: SheetPos, dt: &DataTable) {
```
ddimaria, line 63: Maybe rename to be more descriptive as you're asserting that 2 data tables are equal, one is passed in and the other has the source cell at the `sheet_pos`.
```rust
assert_data_tables_eq(...) {}
```

quadratic-core/src/test_util/assert_data_table.rs
```diff
@@ -290,4 +313,32 @@ mod tests {
         test_create_data_table(&mut gc, SheetId::TEST, pos![D1], 1, 1);
         assert_table_count(&gc, sheet_id, 2);
     }
+
+    #[test]
+    fn test_assert_data_table() {
```
ddimaria, line 318: A test that tests a test, very meta 💡

quadratic-core/src/test_util/create_code_table.rs
```diff
@@ -86,11 +86,13 @@ pub fn test_create_code_table_with_values(
         index: None,
     };
     gc.start_user_transaction(vec![op], None, TransactionName::Unknown);
+
+    gc.data_table(pos.to_sheet_pos(sheet_id)).unwrap().clone()
```
ddimaria, line 90: I could see where this could get confusing in usage since it returns a copy, though not sure there's a great way around this.

ddimaria (approved):
Some non-blocking nits

quadratic-core/src/test_util/create_code_table.rs
davidfig: The function block indicates it's a clone.

quadratic-core/src/controller/execution/execute_operation/execute_col_rows.rs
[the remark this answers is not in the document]
davidfig: fixed it w/better comments

PR #2867, opened 2025-05-04
Ai set formats

davidfig:
- [ ] bold, italic, underline, strikethrough
- [ ] alignments
...


luke-quadratic:
Overall so far this seems to work really great. 

One observation I've had is that the AI is over-eager to look at data. It has the ranges, so why in the case like what's attached here would it need to see the data before just applying the formatting? 

CleanShot 2025-05-06 at 10 13 52@2x

luke-quadratic:
AI is feeling much more confident with doing calculations itself now that it knows what the data is.

We need to doubly enforce that it should use formulas and code to solve problems, not do math and insert the raw values itself.

CleanShot 2025-05-06 at 16 14 23@2x

luke-quadratic:
When the sheet is empty the AI is over-eager to look at what's in the sheet. 

Why wouldn't it know at this point that there's nothing in the sheet? 

CleanShot 2025-05-07 at 09 55 38@2x

luke-quadratic:
There are some compatibility issues with non-Anthropic models 

CleanShot 2025-05-07 at 14 05 22@2x

PR #2868, opened 2025-05-05
chore: remove lock files

AyushAgrawal-A2:
## Description
removes lock files from repo (constantly causing merge conflicts)


.github/workflows/ci.yml
```diff
@@ -45,7 +45,6 @@ jobs:
       - uses: useblacksmith/setup-node@v5
         with:
           node-version: 23
-          cache: "npm"
```
ddimaria, line 48: What's the reasoning behind remove these?

ddimaria:
I started in [🧵 in Slack](https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1746483147295819) to discuss this, but I think we should do this.

.github/workflows/ci.yml
AyushAgrawal-A2: needs lock files for npm cache to work
all ci failed because of use of cache here and lock files not being available https://github.com/quadratichq/quadratic/actions/runs/14828701235

ddimaria (approved):

PR #2879, opened 2025-05-05
feat: allow stop AI interaction via keyboard shortcut

jimniels:
## Description
When the user submits a prompt, allow them to use a keyboard shortcut to stop the chat (uses the same keyboard combo that cursor users).

CleanShot 2025-05-05 at 16 20 27@2x

It's worth noting that this is tied to the chat prompt having focus. I'm not sure if we want to make this a global thing? When you submit a prompt, your focus remains in the prompt, so if you move it this actually won't work. Maybe that's ok and is a worthy tradeoff to making it global.

### DIFF
--- a/quadratic-client/src/app/ui/components/AIUserMessageForm.tsx
+++ b/quadratic-client/src/app/ui/components/AIUserMessageForm.tsx
@@ -5,7 +5,7 @@ import { KeyboardSymbols } from '@/app/helpers/keyboardSymbols';
 import { AIContext } from '@/app/ui/components/AIContext';
 import { AIUsageExceeded } from '@/app/ui/components/AIUsageExceeded';
 import ConditionalWrapper from '@/app/ui/components/ConditionalWrapper';
-import { ArrowUpwardIcon, BackspaceIcon, EditIcon } from '@/shared/components/Icons';
+import { ArrowUpwardIcon, EditIcon } from '@/shared/components/Icons';
 import { Button } from '@/shared/shadcn/ui/button';
 import { Textarea } from '@/shared/shadcn/ui/textarea';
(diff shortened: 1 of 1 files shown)


jimniels:
@AyushAgrawal-A2 curious to hear your thoughts on this (and the implementation)

Intent is to make this action operable via the keyboard and not just a GUI button (found myself wanting this). Cursor has something similar.

CleanShot 2025-05-05 at 16 31 48@2x

quadratic-client/src/app/ui/components/AIUserMessageForm.tsx
```diff
@@ -261,6 +261,8 @@ export const AIUserMessageForm = memo(
                   setEditing(false);
                   bottomTextareaRef.current?.focus();
                 }
+              } else if (loading && event.shiftKey && event.metaKey && event.key === 'Backspace') {
```
AyushAgrawal-A2, line 264: does not work on ubuntu, will not work on windows as well

AyushAgrawal-A2:
> @AyushAgrawal-A2 curious to hear your thoughts on this (and the implementation)
>
> Intent is to make this action operable via the keyboard and not just a GUI button (found myself wanting this). Cursor has something similar.

this is a nice to have shortcut, but I personally never use this as this is a 3 key shortcut and needs both hands on keyboard
after submitting a prompt my hand is always on mouse waiting for AI's reply and browsing something else

jimniels:
Hm...I think that's a good point. I'm not convinced elevating it to the level of global shortcut makes a ton of sense, but don't like the implementation when it's local to the prompt.

I'm gonna close this. Don't feel convinced about an implementation one way or the other (for now).

PR #2884, opened 2025-05-06
WIP: centralize & validate env vars

jimniels:
## Description
- Centralize all env vars in a vite plugin (so we can validate the vars at _build time_). In the plugin:
  - Define a schema for the expected env var values with zod
  - Parse the values and 
    - If it fails, throw an error
    - If it succeeds, generate a TS file with the values (we'll ignore it with git)
- Import all env vars from the generated `@/env.ts` file (instead of haphazardly using `import.meta.env.VITE_*` refs everywhere in the code)

Doing it this way ensures that the app will never build if it's missing an env var we expect to be defined but isn't for a given environment.

To do:

- [ ] Consider if we should do anything with the `env` that gets passed to workers which reference values via `env.VITE_` keys
- [ ] Have everyone check proper values/defaults in these


jimniels:
Gonna close this after a conversation with @AyushAgrawal-A2 around how deployments work for this.

I think it could still be a good idea to centralize all our env vars on the client into a single file with a schema outlining what we expect as values, then we export them wherever we use them (e.g. `import x from '@/env'` rather than doing `import.meta.*` everywhere we need them).

But we can do that separate from this PR in another chunk of work

PR #2887, opened 2025-05-07
New AI tools with different model(s) 

luke-quadratic:
We're adding a lot of context with these tools. This PR tests models that should handle all this added context a lot better.


luke-quadratic:
nvm can just use debug mode, long day

PR #2892, opened 2025-05-09
Fixes bug with multiple col/rows resizing

davidfig:
Fixes a bug where we would resize all columns that had any cell selected, instead of only columns with the column fully selected. And same for rows.

Also fixes the same bug with auto resize col/rows.


quadratic-core/src/a1/a1_selection/query.rs
```diff
@@ -227,6 +227,29 @@ impl A1Selection {
...
+                    for col in range.start.col()..=range.end.col() {
+                        columns.insert(col);
+                    }
+                }
+            }
+            CellRefRange::Table { .. } => (),
+        });
+
+        let mut columns = columns.into_iter().collect::<Vec<_>>();
+        columns.sort_unstable();
```
ddimaria, line 249: Why this instead of `sort()` or `IndexMap` (vs hashset)?

quadratic-core/src/a1/a1_selection/query.rs
```diff
@@ -290,8 +313,30 @@ impl A1Selection {
             .all(|range| !range.selected_rows_finite(a1_context).is_empty())
     }
 
+    /// Returns a list of fully selected rows.
+    pub fn selected_rows(&self) -> Vec<i64> {
```
ddimaria, line 317: This function duplicates `selected_columns()`.  You can just have a single function with a more generic name here.

ddimaria: I know `col()` and `row()` are different, but that could be a param switch.

quadratic-core/src/a1/js_selection/query.rs
```diff
@@ -286,4 +286,22 @@ impl JsSelection {
     pub fn get_selected_table_columns(&self) -> u32 {
         self.selection.selected_table_columns(&self.context) as u32
     }
+
+    #[wasm_bindgen(js_name = "getSelectedColumns")]
+    pub fn get_selected_columns(&self) -> Vec<u32> {
```
ddimaria, line 291: You could just cast as a `u32` in `selected_columns()`, or use generics in the function to allow for different numeric types.

jimniels (approved):
From a usage perspective, this looks good to me 👍

quadratic-core/src/a1/a1_selection/query.rs
davidfig: Same difference.

quadratic-core/src/a1/js_selection/query.rs
davidfig: Not sure which is more idiomatic. We do know it's coming in as i32 and using it as u32 makes more sense.

PR #2909, opened 2025-05-12
Minor tweaks to attempt to fix the freezing viewport

### LINKED ISSUE #2843: Viewport gets locked in place
## Describe the bug
Viewport sometimes gets locked in place for trackpads. Once locked you can't move the viewport, sometimes it gets recovered, sometimes it doesn't. 

## Reproduction 

It seems like this can happen for lots of actions that unfocus the viewport, but the only reliable way to reproduce found so far is via screenshots. 

Repro: while focused on the viewport start a screenshot, note the viewport is now locked in place. 

ping @luke-quadratic if you struggle to recreate, this is hard to show in a video 

It is possible solving this issue will also solve the viewport issue where it gets locked on occasion after PDF operations from AI Chat.

davidfig:
## Relevant issue(s)
Fixes #2843 

## Description
This was as super subtle bug. When pressing cmd+4, the keydown handler on the window registered the cmd down; but it never received the keyup event, since the window no longer had focus (the screenshot tool had focus). I fixed it by checking for the zoom key during the wheel event.

Also cleans up some code when creating renderers (I thought this caused issues--but that wasn't the cause of the bug)


luke-quadratic (changes requested):
This doesn't fix the freezing viewport when taking screenshot, lmk if you need help on repro

luke-quadratic (approved):
Yay works thanks

quadratic-client/src/app/gridGL/pixiApp/Update.ts
```diff
@@ -46,8 +46,17 @@ export class Update {
     this.scrollBarsHandler = undefined;
   }
 
+  private lastFocusElement?: HTMLElement;
+  private showFocus = () => {
+    const focus = document.activeElement;
+    if (focus !== this.lastFocusElement) {
+      this.lastFocusElement = focus as HTMLElement;
+      console.log('current focus:', focus);
```
ddimaria, line 54: Remove this?

ddimaria (approved):
Looks good, just a console.log there that can be removed.

quadratic-client/src/app/gridGL/pixiApp/Update.ts
davidfig: This is a debug setting. If you set debugShowFocus=true, then that fn is called, and we print the current focus to console.

Issue #397, opened 2023-04-03
feat: drag a `.grid` file to import it


jimniels:
Similar to how dragging a CSV file will drop it's data on to the grid, dragging a `.grid` file will save the current file on the grid and import the one that the user just dragged (if it's a valid file, otherwise show some kind of error noting what happened).

davidkircos:
We may want to add a dialogue that confirms this is what they user wants to
do. Because it does deviate from how a CSV is just get imported into the
current sheet

PR #2920, opened 2025-05-13
F4 toggles cell ref types (eg, A1 to $A$1) when editing a formula

davidfig:


luke-quadratic:
This is awesome. 

One spot where I didn't sufficiently scope this - in the case where you've added a ranged reference it would optimally apply to both refs. 

https://github.com/user-attachments/assets/1dc429c9-98de-485a-9b5f-050f7c49f14a

luke-quadratic:
We decided not to implement the proposed feature above in this pass, we'll do in the future since it's a big lift - https://github.com/quadratichq/quadratic/issues/2923

quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
```diff
@@ -98,6 +98,17 @@ class InlineEditorMonaco {
     model.applyEdits([{ range, text: '' }]);
   }
 
+  replaceRange(text: string, range: monaco.Range) {
+    if (!this.editor) {
+      throw new Error('Expected editor to be defined in replaceRange');
+    }
+    const model = this.editor.getModel();
```
ddimaria, line 105: You could call `this.getModel()` to avoid the check below.

quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
```diff
@@ -351,6 +384,48 @@ class InlineEditorMonaco {
...
+      return undefined;
+    }
+
+    // Otherwise, use regex to find the reference at the cursor position
+    else {
+      const position = this.getPosition();
+      const cursorIndex = Math.max(0, position.column - 2);
+
+      // Use regex to find the complete reference at or around cursor position
+      const referenceRegex = /[$A-Za-z0-9']+/g;
```
ddimaria, line 409: This is nearly identical to line #397, except it traverses to the end of the line.  What that difference intentional?  If not, you could reuse this parsing/validation.

quadratic-core/src/a1/a1_context/mod.rs
```diff
@@ -9,7 +9,6 @@ use ts_rs::TS;
 mod sheet_map;
 mod table_map;
 mod table_map_entry;
-pub mod wasm_bindings;
```
ddimaria, line 12: 💯

quadratic-core/src/a1/cell_ref_end.rs
```diff
@@ -337,6 +337,30 @@ impl CellRefRangeEnd {
...
+            self.col.is_absolute = true;
+            self.row.is_absolute = false;
+        } else if col_abs && !row_abs {
+            // from $A1 to A$1
+            self.col.is_absolute = false;
+            self.row.is_absolute = true;
+        } else if !col_abs && row_abs {
+            // from A$1 to $A$1
+            self.col.is_absolute = false;
+            self.row.is_absolute = false;
```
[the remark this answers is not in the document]
ddimaria: @davidfig this seems to be a valid remark.

quadratic-core/src/a1/wasm_bindings.rs
```diff
@@ -2,7 +2,7 @@ use std::str::FromStr;
```
ddimaria, line 2: Why not move this to the wasm_bindings folder?

ddimaria (commented):
I left some non-blocking comments, but please review them.

ddimaria (approved):

quadratic-client/src/app/gridGL/HTMLGrid/inlineEditor/inlineEditorMonaco.ts
davidfig: not sure. but one is a search, and the other is a check

PR #2921, opened 2025-05-13
Fixes bug with marching ants erasing dashed references when editing formulas

davidfig:


quadratic-client/src/app/gridGL/UI/cellHighlights/CellHighlights.ts
```diff
@@ -130,17 +131,14 @@ export class CellHighlights extends Container {
...
       color: colorNumber,
       march: this.march,
       range: refRangeBounds,
       alpha: FILL_SELECTION_ALPHA,
     });
     this.march = (this.march + 1) % Math.floor(DASHED);
-    if (render) {
-      this.dirty = true;
-    }
   };
```
AyushAgrawal-A2, line 142: is this necessary, will cause pixi rerender even when marching highlighted cell is not visible

AyushAgrawal-A2 (changes requested):

luke-quadratic (approved):
Feels great from functionality standpoint.

quadratic-client/src/app/gridGL/UI/cellHighlights/CellHighlights.ts
AyushAgrawal-A2: my bad

AyushAgrawal-A2 (approved):

Issue #2915, opened 2025-05-12
Column/row references shouldn't increment with absolute references on autocomplete


luke-quadratic:
## Describe the bug
Column/row references currently increment with absolute references on autocomplete. They should not.

This exists for all code languages (Formulas, Python, JS)

https://github.com/user-attachments/assets/30e1e6d7-bbc7-428d-b833-55d877b2e74d

## Expected behavior
With a fixed reference like SUM($A:$B) when you autocomplete it would stay $A:$B, not increment to $A2:$B etc 

## Reproduction
As in video: 
- Fill in q.cells('$A:$B') or any absolute reference in any code language
- Autocomplete the cell, note the reference change

PR #2922, opened 2025-05-13
Properly handle absolute rows and absolute columns

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2915

## Description
Column/row references currently increment with absolute references on autocomplete and copy/paste. They should not.
This exists for all code languages (Formulas, Python, JS)

## Testing considerations
This happens with Absolute Full Columns ($A) and Absolute Full Rows ($1).

## QA Wolf tests 
@luke-quadratic has already requested this.

## Demo
autocomplete-absolute-col


AyushAgrawal-A2 (approved):

luke-quadratic (changes requested):
Outside the scope of this issue but seems appropriate to fix here (issue didn't cover full problem here); note how what are absolute references otherwise are not getting treated that way for column/row refs. 

Otherwise this is a great fix. Apologies for poor scoping - didn't think to test these cases in the original bug report. 

### Column should not increment 
In video, each entry should have remained A:B, instead they incremented 
https://github.com/user-attachments/assets/33643284-70de-4baa-a4dc-bc4bd046603e

### Row should not increment 

In video, each entry should have remained 1:3, instead they incremented 
https://github.com/user-attachments/assets/5c832c32-2734-4721-a4fa-a4f350c33853

luke-quadratic (approved):
Looks like my above were addressed 👍

Issue #2914, opened 2025-05-12
Safari unsupported screen


luke-quadratic:
## Describe the bug

We are having major issues with keeping up with Safari bugs since Chrome is our primary supported browser. 

## To resolve 

If a user visits Quadratic with Safari we should pop a screen that states the browser is supported. User can persevere and push through the warning if they'd like to use Quadratic in Safari regardless. 

Recommend they should use Chrome. 

I assume this would best work as a cookie so that Safari users would only see this once, regardless of whether they try safari at the beginning of their Quadratic journey or at their 100th session.

PR #2924, opened 2025-05-13
feat: show browser message to users not on Blink

jimniels:
## Relevant issue(s)
#2914 

## Description
For users who land on the app (not the dashboard, just the app), if they're using a browser that's not a Chromium-based browser, they'll see a message educating them that the app works best in a Chromium browser (with a link to our docs page on the subject).

They have the option to continue anyway (or reach out to us and tell us why we should support their use case).

CleanShot 2025-05-13 at 15 55 35@2x

##  Testing considerations

Open a file in various browsers and ensure you either do or don't see this message. 

Should see the message:

- Safari
- Firefox

Should not see the message:

- Chrome
- Arc
- Brave
- Edge

Here's a publicly-viewable link so you don't have to login:

### DIFF
--- a/quadratic-client/src/routes/file.tsx
+++ b/quadratic-client/src/routes/file.tsx
@@ -1,14 +1,22 @@
 import { EmptyPage } from '@/shared/components/EmptyPage';
-import { DOCUMENTATION_BROWSER_COMPATIBILITY_URL } from '@/shared/constants/urls';
+import { CONTACT_URL, DOCUMENTATION_BROWSER_COMPATIBILITY_URL } from '@/shared/constants/urls';
+import useLocalStorage from '@/shared/hooks/useLocalStorage';
+import { Button } from '@/shared/shadcn/ui/button';
 import { isWASMSupported } from '@/shared/utils/isWASMSupported';
 import { isWebGLSupported } from '@pixi/utils';
-import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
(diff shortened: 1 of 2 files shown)


quadratic-client/src/routes/file.tsx
```diff
@@ -1,14 +1,22 @@
...
 import mixpanel from 'mixpanel-browser';
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
+import { engineName, isDesktop } from 'react-device-detect';
 import { Outlet } from 'react-router';
 
 export function Component() {
+  const [overrideNonBlinkMsg, setOverrideNonBlinkMsg] = useLocalStorage('overrideNonBlinkMsg', false);
+  const [showNonBlinkMsg, setShowNonBlinkMsg] = useState(
+    isDesktop && engineName.toLowerCase() !== 'blink' && !overrideNonBlinkMsg
```
jimniels, line 17: we're using `engineName` because `isChromium` in [react-device-detect](https://github.com/duskload/react-device-detect) doesn't work quite as expected.

ddimaria: Validation seemed to work correctly on Chrome, Brave and Safari.

ddimaria (approved):

luke-quadratic (approved):
Testing worked as expected - recommending minor text change for complete clarity 

"You’re welcome to try others, but things might not work as expected." --> "You're welcome to try non-Chromium browsers, but things might not work as expected"

PR #2928, opened 2025-05-14
chore: upgrade to node 24

AyushAgrawal-A2:


infra/package.json
```diff
@@ -8,6 +8,6 @@
     "@pulumi/pulumi": "^3.0.0"
   },
   "devDependencies": {
-    "@types/node": "^22.13.10"
+    "@types/node": "^22.15.18"
```
ddimaria, line 11: It's crazy that the version of these types don't match up with the node version 🤯

ddimaria (approved):
Do we need to make changes to self-hosting and/or DEVELOPMENT.md?

AyushAgrawal-A2:
>Do we need to make changes to self-hosting and/or DEVELOPMENT.md?

no, DEVELOPMENT.md does not mention node version, just the says to use nvm
no change required for selfhosting, preview of this branch working confirms that

Issue #2624, opened 2025-04-01
Pasting to end of table doesn't expand table


luke-quadratic:
## Describe the bug
Pasting to end of table doesn't expand table 
Also doesn't expand in the columnar (horizontal) direction

See video for clear explainer

![Image](https://github.com/user-attachments/assets/b8f46379-7c6f-4a7c-9125-2e5ec17bd044)

## Expected behavior
Would expand same as if you type into that area

PR #2929, opened 2025-05-14
Pasting near a table grows the table to the left and down

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/2624

## Tasks
- [x] Extract out growing logic from set_cells in order to be reused
- [x] Apply table growing to pasting
- [x] Write tests 

## Description
Pasting to end of table (bottom or right) doesn't expand table.  This PR expands to the bottom or to the right when pasted directly to the bottom or right.  Apply the operations in 2 separate transactions so that the user can undo the table growing without undoing the paste.

Some exceptions:
* No expansion if the paste is touching table heading (would cause a spill otherwise)
* Expansions don't happen when within a data table
* Expansions don't happen grid data would be overwritten by the paste (for the overwritten portion only)
* Tables don't expand over other tables (they stop at the edge)

## Testing considerations
There were a lot of edge cases.  I tried to capture them all but please test all-the-things.

## QA Wolf tests 
QA Wolf should add tests for this new functionality.

## Demo
https://github.com/user-attachments/assets/d3800dc6-e5a0-497c-8f68-78108c9e9b95


luke-quadratic:
Redeploy

quadratic-core/src/grid/sheet/data_table.rs
```diff
@@ -331,18 +331,33 @@ impl Sheet {
...
+
+        let data_table_immediate_left = data_tables
+            .iter()
+            .find(|rect| rect.intersects(rect_moved_left));
+
+        let should_not_expand = data_table_immediate_left.map(|rect| rect_moved_left.min.y <= rect.min.y)
+            .unwrap_or(false);
+
+        !should_not_expand
+    }
```
AyushAgrawal-A2, line 347: one edge case because of this logic, not sure of if important or we want the remaining table to grow

quadratic-core/src/controller/operations/data_table.rs
```diff
@@ -292,6 +294,107 @@ impl GridController {
...
+                        rect.max.y += 1;
+                    });
+            }
+        }
+    }
+
+    /// Returns operations to grow a data table to the right or bottom if the
+    /// cell value is touching the right or bottom edge.
+    pub fn grow_data_table_operations(
+        &self,
```
AyushAgrawal-A2, line 364: `&self` can be removed

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -649,6 +661,14 @@ impl GridController {
...
+        // collect information for growing data tables
+        let mut data_table_columns: HashMap<SheetPos, Vec<u32>> = HashMap::new();
+        let mut data_table_rows: HashMap<SheetPos, Vec<u32>> = HashMap::new();
+        let existing_data_tables = self.a1_context_sheet_table_bounds(sheet_id);
+        let mut growing_data_tables = existing_data_tables.clone();
+        let bypass_growing =
+            !Sheet::should_expand_data_table(&existing_data_tables, clipboard_rect);
+
         // loop through the clipboard and replace cell references in formulas and other languages
         for (start_x, x) in (insert_at.x..=max_x)
```
AyushAgrawal-A2, line 673: update comment pls, we are checking for grow table also now

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -692,11 +707,34 @@ impl GridController {
...
                                     adjust,
                                 );
+                            } else {
+                                let new_x = x + cols_x as i64;
+                                let new_y = y + *cols_y as i64;
+                                let current_pos = Pos::new(new_x, new_y);
+                                let within_data_table = existing_data_tables
+                                    .iter()
+                                    .any(|rect| rect.contains(current_pos));
+
```
AyushAgrawal-A2, line 722: this part seems expensive, we are 3 loops deep here and then iterate over all data tables (Imports in this sheet), we can filter `existing_data_tables` to be just the ones intersecting with clipboard rect to reduce this loop size to only our region of interest

quadratic-core/src/controller/operations/cell_value.rs
```diff
@@ -30,14 +30,7 @@ impl GridController {
...
         let mut data_table_ops = vec![];
-        let existing_data_tables = self
-            .a1_context()
-            .tables()
-            .filter(|table| {
-                table.sheet_id == sheet_pos.sheet_id && table.language == CodeCellLanguage::Import
-            })
-            .map(|table| table.bounds)
-            .collect::<Vec<_>>();
+        let existing_data_tables = self.a1_context_sheet_table_bounds(sheet_pos.sheet_id);
```
AyushAgrawal-A2, line 33: same as prev, comment, we can filter `existing_data_tables` to be just the ones intersecting with `values` rect to reduce the loop size later, which is 2 loops deep

AyushAgrawal-A2 (changes requested):
few comments and 1 test failing

test failing as now table is growing on paste, the test assertion needs to be updated

luke-quadratic (approved):
approved on functionality - my discussion item around pasting empty/formatted was addressed in standup 👍

quadratic-core/src/controller/operations/clipboard.rs
ddimaria: Done

quadratic-core/src/controller/operations/cell_value.rs
ddimaria: Done

AyushAgrawal-A2 (approved):

Issue #2911, opened 2025-05-12
Onboarding Questions [wip]


davidkircos:
Are you using Quadratic for?
Personal
Work
Education

What kind of work do you do?
Data
Engineering
Sales
Marketing
Product
ML
Other

How technical are you?
Not technical
I can figure it out
My role is technical

What features are you excited about with the Quadratic spreadsheet?
AI
Python
Database connections
Performance
Data security

PR #2945, opened 2025-05-16
feat: onboarding questionnaire

jimniels:
## Relevant issue(s)
#2911 

## Description

Available at: [`/onboarding`](https://jim-onboarding.quadratic-preview.com/onboarding).

- Triggers for ~25% of new signups (on desktop who visit `/`) by redirecting them to `/onboarding`
- No way to "opt out" except by closing your browser tab (you won't see it again next time you come back to the app)
- Doesn't trigger for people coming:
        - To a new file with a prompt, e.g. `files/create?prompt=...`
        - To a new file cloned from a template
        - To a file that was shared with them
        - On a mobile/tablet device

See test cases below for details on expected inner workings.

### Data

Question data is defined purely on the front-end. Responses are saved to

- Our database (new column on `User` table, stored as JSON)
- Mixpanel

The data structure has a `__version` key to allow us to change questions/answers later.

This is the current structure of the v1 schema that will be saved to db/mixpanel:

```
{
  __version: 1,
  
  // Question 1
  use: 'work' | 'personal' | 'education';
  
  // Questions 2-x, depending on Question 1
  'work-role'?: string;
  'work-role-other'?: string;
  'personal-uses[]'?: string[];
  'personal-uses-other'?: string;
  'education-identity'?: string;
  'education-identity-other'?: string;
  'education-subjects[]'?: string[];
  
  // Penultimate Question
  'languages[]'?: string[];

  // Final Question
  'goals[]': string[];
}
```

## To Test

Main uses cases:

- [x] Come into the app the first time at `/` and see that, after signing up, you are redirected to the onboarding flow
    - [ ] Fill out the onboarding flow, and see that you're redirected to the creation of a new **public** team file afterwards
    - [ ] Skip the onboarding flow by closing your tab. Re-open the app at `/` and see that you're taken to the dashboard.
- [x] Come into the app the first time via a prompt at [`/files/create?prompt=...`](https://jim-onboarding.quadratic-preview.com/files/create?prompt=rick+roll+me) and see that you skip the onboarding flow
- [x] Try filling out the form in various ways at [`/onboarding`](https://jim-onboarding.quadratic-preview.com/onboarding), including:
    - Answering questions in 'Work' use
    - Answering questions in 'Personal' use
    - Answering questions in 'Education' use
    - Try filling out an "Other" option
    - Try not selecting a language
- [x] When you finish the onboarding survey, you will be taken to a new (public) team file with a prompt that automatically runs
     - If you chose "other" on some of your answers, it will be a generic prompt about being a new user and creating some data + a chart
     - If you chose the pre-filled answers, you'll get a prompt customized to the answer you chose.

Edge cases:

- [x] `/onboarding` is guarded by auth. If you visit and aren't authenticated, you get redirected to login
- [ ] `/onboarding` is visitable if you know the URL. If you re-fill out the form, we overwrite your existing answer. We don't check to see if you've filled it out already.

## (Optional) QA Wolf tests 

- Do we want to have QA wolf create tests for this? @luke-quadratic


quadratic-client/src/app/ai/toolCards/ToolCard.tsx
```diff
@@ -34,13 +34,13 @@ export const ToolCard = ({
           <div className="truncate font-bold">
             {label ? label : isLoading ? <Skeleton className="h-3 w-64 bg-accent" /> : null}
           </div>
-          <div className="truncate text-xs text-muted-foreground">
+          <div className="h-4 truncate text-xs text-muted-foreground">
```
jimniels, line 37: Setting a fixed height here so this doesn't shift as content comes in

CleanShot 2025-05-23 at 17 01 27

quadratic-client/src/app/ai/toolCards/AddDataTable.tsx
```diff
@@ -49,7 +49,7 @@ export const AddDataTable = ({ args, loading }: AddDataTableProps) => {
       icon={icon}
       label={
         <span>
-          {label} <span className="text-muted-foreground">| {table_name}</span>
+          {label} <span className="ml-1 font-normal text-muted-foreground">{table_name}</span>
```
jimniels, line 52: De-emphasizing the table name and using whitespace as a separator

CleanShot 2025-05-23 at 17 02 43@2x

quadratic-api/src/routes/v0/user.acknowledge.GET.ts
```diff
@@ -18,6 +18,6 @@ export default [validateAccessToken, userMiddleware, handler];
  * because multiple API calls are made in parallel on the client when the user hasn't
  * been created yet or associated with teams and/or files.
  */
-async function handler(req: RequestWithUser, res: Response<ApiTypes['/v0/users/acknowledge.GET.response']>) {
+async function handler(req: RequestWithUser, res: Response<ApiTypes['/v0/user/acknowledge.GET.response']>) {
```
jimniels, line 21: I renamed these from the plural `users` to the singular `user` because these endpoints are strictly about the individual user who is making the request (which we figure out from the auth token). So now we have `GET user/acknowledge` and `POST /user` in the API

quadratic-api/prisma/schema.prisma
```diff
@@ -9,15 +9,16 @@ datasource db {
...
-    ownedFiles      File[]            @relation(name: "OwnerRelation")
-    createdFiles    File[]            @relation(name: "CreatorRelation")
-    QFeedback       QFeedback[]
-    UserTeamRole    UserTeamRole[]
-    UserFileRole    UserFileRole[]
-    AnalyticsAIChat AnalyticsAIChat[]
+    id                  Int               @id @default(autoincrement())
+    auth0Id             String            @unique @map("auth0_id")
+    eduStatus           EduStatus?        @map("edu_status")
+    onboardingResponses Json?             @map("onboarding_responses")
```
jimniels, line 15: `onboardingResponses` is the new field and just takes a JSON which we define on the client

ddimaria: Sometimes github isn't great at these diffs, most are just added tabs 🤯

quadratic-api/src/routes/v0/user.POST.test.ts
```diff
@@ -0,0 +1,46 @@
+import request from 'supertest';
```
ddimaria, line 1: Nice tests 💯

quadratic-api/src/routes/v0/user.acknowledge.GET.ts
ddimaria: Will this be problematic when this service is deployed?  Older clients will be hitting the old endpoint.

quadratic-client/src/app/ai/toolCards/ToolCard.tsx
```diff
@@ -36,13 +36,13 @@ export const ToolCard = memo(
...
             </div>
-            <div className="truncate text-xs text-muted-foreground">
+            <div className="h-4 truncate text-xs text-muted-foreground">
               {hasError ? (
                 <span className="text-destructive">Something went wrong (try again)</span>
               ) : description ? (
                 description
               ) : isLoading ? (
-                <Skeleton className="h-2 w-24 bg-accent" />
+                <Skeleton className="h-3 w-24 bg-accent" />
```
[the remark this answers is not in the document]
ddimaria: @luke-quadratic you may want to play with browser widths to see if this fixed height is problematic.

quadratic-client/src/dashboard/onboarding/Questions.tsx
```diff
@@ -0,0 +1,563 @@
...
+  );
+}
+
+function QuestionFormFooter({ disabled }: { disabled?: boolean }) {
+  const navigate = useNavigate();
+  const fetcher = useFetcher({ key: FETCHER_KEY });
+  const { currentId, currentQuestionNumber, currentQuestionsTotal } = useOnboardingLoaderData();
+  const btnClassName = 'select-none';
+
+  let _disabled = disabled === undefined ? false : disabled;
```
ddimaria, line 421: What's the significance of this underscore?

quadratic-client/src/dashboard/onboarding/getPrompt.ts
```diff
@@ -0,0 +1,38 @@
+import { questionsById, type OnboardingResponseV1 } from '@/dashboard/onboarding/Questions';
+
+export const DEFAULT_PROMPT =
```
ddimaria, line 3: @luke-quadratic can you validate these prompts?

quadratic-api/src/routes/v0/user.acknowledge.GET.ts
jimniels: I actually don't think it will because the only place this endpoint gets hit is in `login-result`, meaning, the only time we will call this endpoint is when the user has gone through the auth-flow and gone from

our website -> auth0 -> our website

in which case, presumably, they'll always have the freshest version of the client which will redirect to this new endpoint.

quadratic-client/src/dashboard/onboarding/Questions.tsx
jimniels: it was actually used as an internal representation of the value which we would possibly mutate, but then I moved that out of this function and didn't rename. we should just use the prop `disabled` will fix.

ddimaria (approved):

quadratic-client/src/routes/login-result.tsx
```diff
@@ -3,6 +3,9 @@ import { apiClient } from '@/shared/api/apiClient';
 import { isMobile } from 'react-device-detect';
 import { redirect } from 'react-router';
 
+// Show onboarding for ~25% of new users
+const SHOW_ONBOARDING = Math.random() < 0.25;
```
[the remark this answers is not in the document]
jimniels: Hm...this doesn't really matter because it's a client loader route, but I'll make the change anyway in case of some future where we moved this (and it would be the case)

quadratic-client/src/dashboard/onboarding/getPrompt.ts
```diff
@@ -0,0 +1,38 @@
...
+ * for them to get started. If they chose 'other', we'll return a generic prompt
+ */
+export function getPrompt(data: OnboardingResponseV1) {
+  if (data.use === 'work' && data['work-role'] && data['work-role'] !== 'other') {
+    const value = data['work-role'];
+    const label = questionsById['work-role'].optionsByValue[value];
+    if (label)
+      return `Hi, I'm a new user to Quadratic and I plan on using it for work. I work in ${label}; show me what you can do by creating me a sample dataset for that field. Once finished with the dataset, create a chart that helps explain the data. Then tell me what else is possible in Quadratic.`;
+  } else if (data.use === 'personal' && data['personal-uses[]'] && data['personal-uses[]'].length > 0) {
+    const value = data['personal-uses[]'].filter((value) => value !== 'other')[0];
```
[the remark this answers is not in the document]
jimniels: This is fine because `[][0]` would be `undefined` and accessing `optionsByValue[undefined]` will result in `undefined` and we check for that value before interpolating it. So it's fine.

quadratic-shared/typesAndSchemas.ts
```diff
@@ -430,10 +430,18 @@ export const ApiSchemas = {
 
   /**
    * ===========================================================================
-   * Users
+   * User
    * ===========================================================================
    */
-  '/v0/users/acknowledge.GET.response': z.object({ message: z.string(), userCreated: z.boolean() }),
+  '/v0/user/acknowledge.GET.response': z.object({ message: z.string(), userCreated: z.boolean() }),
```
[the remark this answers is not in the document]
AyushAgrawal-A2: >given that we deploy in off times (and this will just result in a slightly broken experience, if you refresh you should be ok at that point but you'll miss getting the new onboarding experience) seems like this is an ok tradeoff?

all logins will fail during this duration, not just the `onboarding EXPERIENCE`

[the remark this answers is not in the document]
jimniels: ah, well then we should be fine then 👍

quadratic-shared/typesAndSchemas.ts
```diff
@@ -430,10 +430,18 @@ export const ApiSchemas = {
...
+  '/v0/user.POST.request': z.object({
+    onboardingResponses: z
+      .object({
+        __version: z.number(),
+      })
+      .catchall(z.any()),
+  }),
+  '/v0/user.POST.response': z.object({ message: z.string() }),
```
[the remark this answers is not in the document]
jimniels: Hm... `/v0/users/acknowledge.GET.response` is not referenced anywhere anymore.

The old version of the app served during the deploy window will  still point at `users/acknowledge` so we just need to make sure that _old_ route still works in the _new_ API — which it does, because the new API will be pointing at the new type definitions (without the `s`) so everything should work as expected.

Ignoring this suggestion.

quadratic-client/src/routes/onboarding.tsx
```diff
@@ -0,0 +1,162 @@
...
+      level: 'error',
+      extra: {
+        error: result.error,
+      },
+    });
+    sentryPromises.push(Sentry.flush(2000));
+  }
+
+  if (sentryPromises.length > 0) {
+    await Promise.all(sentryPromises);
```
AyushAgrawal-A2, line 157: pls add try catch here, this promise failing should not fail the redirect to app

AyushAgrawal-A2 (approved):

luke-quadratic (approved):
Works well from UI and prompt standpoint, checked the testing considerations that I explored

PR #2954, opened 2025-05-18
Fix copy chart

### LINKED ISSUE #2932: Pasting a chart on another chart creates major sheet issues
Note: I think this could be a 1/3 because of rarity (I only experienced it by fat-fingering), but marking 2/3 because the state this puts your sheet in isn't great. 

## Describe the bug
Creates charts in every cell space the chart currently occupies so they all spill. And the sheet becomes problematic afterwards. 

https://github.com/user-attachments/assets/f0d61aa4-a771-458b-bd50-eeffb63051c5

Strange behavior after - core stuck 

Image

## Expected behavior 

Assuming if you paste on top of the existing space it would just take over that space, would seem like nothing happened to the user other than presumably the code would be recomputed?

davidfig:
## Relevant issue(s)
Fixes #2932

## Description
- [x] copy the entire range of the chart bounds (it used to only copy the anchor cell)

## Testing 
Need to test cases of chart selection, since this changes how we convert from a table reference to a sheet reference when dealing with charts.


AyushAgrawal-A2 (changes requested):
pls add test

AyushAgrawal-A2 (approved):

luke-quadratic:
This is better behavior but why wouldn't pasting just paste over the existing; why does it create a second chart a column down? 

CleanShot 2025-05-19 at 07 41 58

luke-quadratic (approved):
👍

PR #2956, opened 2025-05-19
AI adds code cell name

### LINKED ISSUE #2937: AI automatically name code cells
Data tables when created are automatically receiving names. Code tables should work the same way - name based on the code generated. 

Only change the name of a code cell the first time the code is created; otherwise we run the risk of constantly breaking peoples' references when AI updates names constantly. 

Image

davidfig:
## Relevant issue(s)
Fixes #2937

## Description
This automatically adds a code cell name when the code cell is created by AI.

There are a few additional features that would be nice to have in the future:
1. name a code cell when a user manually creates the code cell
2. name a data table when a user manually converts the sheet to a data table
3. name a data table when a AI converts the sheet to a data table


luke-quadratic (approved):
Awesome!

AyushAgrawal-A2 (approved):

PR #2980, opened 2025-05-21
trivial frontend fixes

### LINKED ISSUE #2977: In Formula edit mode a broken link pops up in bottom bar - remove it
## Describe the bug

In Formula edit mode a broken link pops up in bottom bar - remove it. 

Image

### LINKED ISSUE #2979: bug: handle large connection names
## Describe the bug

If a user gives a connection a big name, it should be handle gracefully everywhere we show the connection name

Image

Image

Image

AyushAgrawal-A2:
## Relevant issue(s)
closes #2979
closes #2977

### DIFF
--- a/quadratic-client/src/app/ui/menus/BottomBar/BottomBar.tsx
+++ b/quadratic-client/src/app/ui/menus/BottomBar/BottomBar.tsx
@@ -26,9 +26,7 @@ export const BottomBar = () => {
 
         {inlineEditorState.visible && (
           <BottomBarItem className="hidden lg:block">
-            <a href="https://docs.quadratichq.com/TODO:" target="_blank" rel="noopener noreferrer">
-              {inlineEditorState.editMode ? 'Edit' : 'Enter'} (F2)
-            </a>
+            {inlineEditorState.editMode ? 'Edit' : 'Enter'} (F2)
           </BottomBarItem>
         )}
 
--- a/quadratic-client/src/app/ui/menus/CellTypeMenu/CellTypeMenu.tsx
+++ b/quadratic-client/src/app/ui/menus/CellTypeMenu/CellTypeMenu.tsx
@@ -171,7 +171,7 @@ function CommandItemWrapper({
       }}
     >
       <div className="mr-4 flex h-5 w-5 items-center">{icon}</div>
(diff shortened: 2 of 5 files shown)


quadratic-client/src/app/ui/components/Markdown.tsx
```diff
@@ -10,7 +10,18 @@ interface Props {
 export const Markdown = memo(({ children }: Props) => {
   return (
     <div className="markdown">
-      <ReactMarkdown remarkPlugins={[remarkGfm]}>{children}</ReactMarkdown>
+      <ReactMarkdown
+        remarkPlugins={[remarkGfm]}
+        components={{
+          a: ({ node, children, ...props }) => (
+            <a target="_blank" rel="noreferrer" {...props} className="test-link">
```
jimniels, line 17: Not sure if you meant to have this `test-link` class?

I would say remove the the `a {...}` styling in the Markdown.scss. Those are hard-coded and won't change with the theme. Instead, I'd suggest making it underlined and on hover it changes to the theme color, which you could do by putting tailwind classes on the `<a>` of `underline hover:text-primary`

CleanShot 2025-05-22 at 07 16 04

luke-quadratic (approved):
Good fixes thanks

quadratic-client/src/app/ui/components/Markdown.tsx
AyushAgrawal-A2: done

AyushAgrawal-A2:
@jimniels 

lmk if you find more, want to fix all

jimniels:
There's one more that I can think of:

When you have a connection it shows up in the dashboard under the "Import" menu. Looks like that probably needs like a max-width and truncation.

CleanShot 2025-05-22 at 15 14 56@2x

AyushAgrawal-A2:
> There's one more that I can think of:
> 
> When you have a connection it shows up in the dashboard under the "Import" menu. Looks like that probably needs like a max-width and truncation.
> 
> CleanShot 2025-05-22 at 15 14 56@2x

done

jimniels (approved):
nice, LGTM 👍  good job squashing out all those truncation problems

davidfig (approved):

PR #2988, opened 2025-05-24
Show when code was last run

### LINKED ISSUE #2950: Last run timestamp
(Filed on user feedback. Noting this is also a quite common feature across BI tools.) 

Ideally this would be available on all code cells. 

Need to figure out what its real estate would be but something like the wireframe below could make sense and works across all code cell types. 

## Wireframing, in-depth details, docs, etc.

Image

davidfig:
## Relevant issue(s)
Fixes #2950 

- [x] show last modified in table header
- [x] show last modified in return value for code editor
- [x] remove minimum size for html (it was limited to DEFAULT_CELL_WIDTH and DEFAULT_CELL_HEIGHT, which resulted in weird behavior when resizing the column/row smaller than those values)
- [x] fix bug when changing chart size, and then resizing column/rows within that chart

## Examples
On table headers:
CleanShot 2025-05-24 at 15 59 22@2x

In Code Editor:
CleanShot 2025-05-24 at 15 56 04@2x


AyushAgrawal-A2 (approved):

quadratic-client/src/shared/utils/timeAgo.ts
```diff
@@ -42,3 +50,61 @@ export function timeAgo(dateString: string | number, force = false) {
...
+
+  // If the difference is more than 24 hours, return the formatted date
+  if (duration > 86400000) {
+    return {
+      timeAgo: date.toLocaleDateString(undefined, {
+        year: 'numeric',
+        month: 'short',
+        day: 'numeric',
+      }),
+      nextInterval: 24 * 60 * 60 * 1000 - duration, // Time until next date change
```
AyushAgrawal-A2, line 75: revert to -1

AyushAgrawal-A2 (changes requested):

PR #2994, opened 2025-05-27
Better ensure visible handling

### LINKED ISSUE #2990: Autofill shouldn't pan back to root cell
## Describe the bug
As I autocomplete these formulas, note how I'm popped back to the root cell, this is quite annoying.

Image

## Expected behavior
Don't move viewport after finishing action. 

Image

davidfig:
## Relevant issue(s)
Fixes #2990

## Description
- [x] ensureVisible checks the entire selection to check if in bounds; doesn't move the viewport if any part of the selection is visible
- [ ] better cursor behavior when expanding

### DIFF
--- a/quadratic-client/src/app/grid/sheet/Sheet.ts
+++ b/quadratic-client/src/app/grid/sheet/Sheet.ts
@@ -159,7 +159,7 @@ export class Sheet {
   }
 
   // @returns screen rectangle from a selection rectangle
-  getScreenRectangleFromRect(rect: Rectangle): Rectangle {
+  getScreenRectangleFromRectangle(rect: Rectangle): Rectangle {
     return this.getScreenRectangle(rect.x, rect.y, rect.width, rect.height);
   }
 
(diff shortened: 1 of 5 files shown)


luke-quadratic (changes requested):
This case for column/row selections should not pan back 

https://github.com/user-attachments/assets/f9c633ea-47df-4d7c-af1b-013304505bd1

quadratic-client/src/app/grid/sheet/SheetCursor.ts
```diff
@@ -134,6 +134,25 @@ export class SheetCursor {
     return rectToRectangle(rect);
   };
 
+  /// Returns the largest rectangle that contains all the selection, including
+  /// unbounded ranges. Converts the BigInt::MAX to Number::MAX.
+  getLargestRectangleUnbounded = (): Rectangle => {
+    const rect = this.jsSelection.getLargestUnboundedRectangle();
+    console.log(rect.min.x, rect.min.y, rect.max.x, rect.max.y);
```
ddimaria, line 141: Remove these logs?

ddimaria (approved):

quadratic-client/src/app/grid/sheet/SheetCursor.ts
AyushAgrawal-A2: done

PR #2997, opened 2025-05-28
Fix bug with copyAsPNG

davidfig:


luke-quadratic (approved):
Nice lgtm, tested a few cases 

- Single table 
- Multiple tables 
- Multiple tables and non-table content 
image

PR #3001, opened 2025-05-29
Fix dt del row

### LINKED ISSUE #2971: Deleting rows from table with column selection is not undoable
## Describe the bug

Deleting rows from table with column selected is not undoable. 

Any combination of selection where the column header is included in the row deletion is broken.

## Expected behavior

Rows would be deleted but also be undoable. 

## Reproduction
1. Create a table of any size 
2. Delete rows from that table with column headers included
3. Attempt undo 

https://github.com/user-attachments/assets/f3b02ec0-a47e-428f-be7e-f6d86fa94f9b

davidfig:
## Relevant issue(s)
Fixes #2971

## Description
Fixes delete all rows in a table is not undoable

### DIFF
--- a/quadratic-core/src/controller/user_actions/data_table.rs
+++ b/quadratic-core/src/controller/user_actions/data_table.rs
@@ -621,4 +621,29 @@ mod tests {
         assert_eq!(gc.data_table(pos![sheet_id!A1]), Some(&dt));
         assert!(gc.data_table(pos![sheet_id!A2]).is_none());
     }
+
+    #[test]
+    fn test_data_table_delete_rows_undo() {
+        let mut gc = test_create_gc();
+        let sheet_id = first_sheet_id(&gc);
+
+        test_create_data_table(&mut gc, sheet_id, pos![A1], 2, 2);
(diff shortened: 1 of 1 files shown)


quadratic-core/src/controller/execution/execute_operation/execute_data_table.rs
```diff
@@ -1616,30 +1616,64 @@ impl GridController {
...
+            let sheet = self.try_sheet_result(sheet_id)?;
+            let data_table_pos = sheet.data_table_pos_that_contains(sheet_pos.into())?;
+            let data_table = sheet.data_table_result(&data_table_pos)?;
+
+            // check if the rows to delete are part of the data table's UI, bail if so
+            // we need rows relative to the data table, not the sheet, hence (0, 0)
+            let ui_rows = data_table.ui_rows((0, 0).into());
+            if !ui_rows.is_empty() && rows.iter().any(|row| ui_rows.contains(&(*row as i64))) {
+                let e = "delete_rows_error".to_string();
+                if transaction.is_user_undo_redo() && cfg!(target_family = "wasm") {
```
ddimaria, line 1629: Consider moving this to the wasm directory.

PR #3029, opened 2025-06-03
fix: display of text in prev user ai message

jimniels:
## Description

Today if you have some text that's really long and wraps, it messes up the overflow and scrolling of the sheet chat.

https://github.com/user-attachments/assets/51af991a-ad43-486d-852c-747d87a9f054

It looks like this was happening because we're conditionally rendering a `<textarea>` or a `<div>` depending on whether the box is currently being edited. But divs will wrap things differently than textarea, so my thought is to just always render a textarea and just conditionally make it editable/focusable based on the `editing` state we were using previously.

This should make it so text always wraps correctly inside the textarea, as the browser handles that kind of stuff by default for a textarea primitive.

CleanShot 2025-06-03 at 15 37 35@2x

### DIFF
--- a/quadratic-client/src/app/ui/components/AIUserMessageForm.tsx
+++ b/quadratic-client/src/app/ui/components/AIUserMessageForm.tsx
@@ -232,59 +232,51 @@ export const AIUserMessageForm = memo(
           textareaRef={textareaRef}
         />
 
-        {editing ? (
-          <Textarea
-            ref={textareaRef}
-            value={prompt}
-            className={cn(
-              'rounded-none border-none p-2 pb-0 shadow-none focus-visible:ring-0',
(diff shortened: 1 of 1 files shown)


quadratic-client/src/app/ui/components/AIUserMessageForm.tsx
```diff
@@ -232,59 +232,51 @@ export const AIUserMessageForm = memo(
           textareaRef={textareaRef}
         />
 
-        {editing ? (
-          <Textarea
```
jimniels, line 236: Didn't change any of the attributes on the `<Textarea>` here except for the addition of `readOnly` and different `className` values based on `editing`, since we were using that before to render a `<div>` or `<Textarea>` but now we just always render the same element and just make it "active" or "inactive"

AyushAgrawal-A2 (approved):

PR #3049, opened 2025-06-09
chore: add shimmer animation to web search loading state

AyushAgrawal-A2:


jimniels (approved):
nice, works as expected — shimmers until sources come in

CleanShot 2025-06-09 at 17 35 51

luke-quadratic (approved):
👍

PR #3051, opened 2025-06-10
Model router token count 

luke-quadratic:
## Description 

Testing token counting in router. Will scope an issue for this if it's determined to be a good metric for passing off requests to different models.


luke-quadratic:
Tokens are too variable, not inspired by this approach. 

Seems like number of code cells has the most signal so far.

PR #3072, opened 2025-06-12
perf: clipboard and move cells

### LINKED ISSUE #2894: bug: performance degradation when moving a large table
## Describe the bug
Moving a table with 26 cols and ~2,200 rows results in a really slow move. Feels like the app didn't register that I moved it, but after a few seconds it catches up.

https://github.com/user-attachments/assets/351bf038-8f06-4cde-832e-c0afd114b06f

## Expected behavior
It either happens much faster, or we provide some kind of mechanism to give the user feedback that we registered the operation and its taking place but needs a moment.

## Reproduction
Try using this file

AyushAgrawal-A2:
## Relevant issue(s)
closes #2894 

## Testing
Try Cut, Copy, Past, Select Move
- Flat data
- Tables
- With / without formatting
- With / without code cells
- from table / into table
- copy/paste from/into table having hidden columns, sorted rows, different show ui states


jimniels:
Played with this and it was WAY faster than current QA/Prod. 

However did encounter an issue where copy/pasting a table from existing version of the app pasted the data but not as a table.

https://github.com/user-attachments/assets/f6316804-a5c7-435f-b224-e643d8e1f148

AyushAgrawal-A2:
> Played with this and it was WAY faster than current QA/Prod.
> 
> However did encounter an issue where copy/pasting a table from existing version of the app pasted the data but not as a table.
> 
>  CleanShot.2025-07-02.at.14.26.49.mp4

@jimniels 
this is expected, old clipboard has different encoding, new clipboard is also zstd compressed so no compatible with each other

luke-quadratic:
Move action should just move the entire table. (noting QA/prod also has issues but the QA issue is different where the data is just getting copied)

https://github.com/user-attachments/assets/528de0d4-b99f-478d-905c-7ba47698d283

luke-quadratic:
There's a noticeable slowdown when copy/pasting content into a table when the table has sorted columns. 

Slower than QA in this regard.

luke-quadratic (approved):
Followed test plan + some and didn't run into anything obvious, aside from what was stated above. 

@jimniels a second pair of eyes wouldn't hurt

jimniels (approved):
Great improvement. SO much faster than before. It's noticeable that moving a sorted table is slightly slower than a non-sorted table. Non-sorted tables seem instantaneous, whereas sorted tables there's a very slight but noticeable slowdown. Both are still WAY better than what's in prod today. Nice work @AyushAgrawal-A2 !

jimniels (changes requested):
Actually I'm noticing a regression that I don't see in QA or prod.

Moving your cursor to a cell with content then typing doesn't replace the cell's contents (which it does in qa/prod today)

For example, in this video I move the cursor to an empty column and type: 1, enter, 2, enter, 3, enter

Then I go over to the column next to it that has content in it and do the same thing and it doesn't overwrite the cells.

https://github.com/user-attachments/assets/be73931d-4617-40bf-a18c-f43b67ee22af

jimniels:
I acknowledge this is _really_ nit picky, so feel free to disregard as it's definitely more of a polish thing than core functionality.

When I move a big table (I think it's happening with small tables too, but it's so fast that it's barely noticeable) there's a rendering thing where you see like a flash of content during the move.

Note how in this video each time I move the table down, in the upper-left corner of the table where the table header is, the table name flashes along with the cell cursor. It's pretty subtle, but if you move a big table sequentially a few times you'll notice it.

https://github.com/user-attachments/assets/2ed0fb96-dc3c-43e1-8254-38f64ddc50b6

If you look at it frame by frame in the video, this is what's happening:

CleanShot 2025-07-11 at 09 58 13@2x

CleanShot 2025-07-11 at 09 58 29@2x
CleanShot 2025-07-11 at 09 58 21@2x
CleanShot 2025-07-11 at 09 58 39@2x

Again, not a huge deal. If it seems like something you want to fix, go for it. Otherwise I'd much rather have this improvement than not. It's way better than what's in prod today!

AyushAgrawal-A2:
> (feels minor but worth noting)
> 
> There's a noticeable slowdown when copy/pasting content into a table when the table has sorted columns.
> 
> Slower than QA in this regard.
> 
> Edit: everything in the table seems to slow down when it's sorted



> Great improvement. SO much faster than before. It's noticeable that moving a sorted table is slightly slower than a non-sorted table. Non-sorted tables seem instantaneous, whereas sorted tables there's a very slight slowdown. Both are still WAY better than what's in prod today. Nice work @AyushAgrawal-A2 !

@luke-quadratic @jimniels 

perf should be  good now for sorted tables as well, I have enough changes that need user testing again

most formatting fns are redone, all formatting interactions need to be tested 
- inside  / outside of tables / combination
- copy paste formats for all table states (sort, hidden and show_ui), from table / into table

luke-quadratic:
@davidkircos final set of eyes that aren't mine before merging would be good

quadratic-core/src/controller/execution/execute_operation/execute_data_table.rs
```diff
@@ -1146,10 +1142,12 @@ impl GridController {
 
                     dt.insert_column_sorted(index as usize, column_header, values)?;
 
-                    if !format_update.is_default() {
-                        dt.formats
-                            .get_or_insert_default()
-                            .apply_updates(&format_update);
+                    if let Some(format_update) = format_update {
```
ddimaria, line 1145: You should be able to combined these conditionals now
```rust
if let Some(format_update) = format_update && !format_update.is_default()
```

quadratic-core/src/controller/user_actions/formats.rs
```diff
@@ -23,161 +22,115 @@ impl GridController {
...
-        // set table formats
-        for table_ref in table_ranges {
-            if let Some(table) = context.try_table(&table_ref.table_name) {
-                if let Some(range) =
-                    table_ref.convert_to_ref_range_bounds(true, context, false, true)
-                {
-                    add_table_ops(range, table, &mut ops);
-                }
-            }
+        dbgjs!(&format!(
```
davidfig, line 124: dbg

davidfig (approved):

quadratic-core/src/controller/operations/clipboard.rs
```diff
@@ -332,6 +517,8 @@ impl GridController {
                     .get(&source_pos)
                     .and_then(|data_table| data_table.cells_accessed(start_pos.sheet_id));
 
+                dbg!(&cells_accessed);
```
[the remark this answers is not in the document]
AyushAgrawal-A2: done

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
```diff
@@ -269,10 +283,30 @@ impl GridController {
...
-            let existing_offsets = sheet.offsets.clear_widths();
+        let Some(sheet) = self.try_sheet_mut(sheet_id) else {
+            return;
+        };
+
+        let existing_offsets = sheet.offsets.clear_widths();
+
+        let old_size = sheet.offsets.set_default_width(size);
+
+        if (cfg!(target_family = "wasm") || cfg!(test)) && !transaction.is_server() {
```
ddimaria, line 295: Consider moving to the wasm dir

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
```diff
@@ -286,16 +320,30 @@ impl GridController {
...
+    }
+
+    pub fn execute_default_row_size(
+        &mut self,
+        transaction: &mut PendingTransaction,
+        op: Operation,
+    ) {
+        unwrap_op!(let DefaultRowSize { sheet_id, size } = op);
+
+        let Some(sheet) = self.try_sheet_mut(sheet_id) else {
```
ddimaria, line 340: This feels like an anti-pattern (I know, it's everywhere).  Consider returning a result here since we have some mechanism for capturing the result in other exec files.

ddimaria (approved):
Rust code looks good, just left some questions/nits

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
```diff
@@ -269,10 +283,30 @@ impl GridController {
...
+            return;
+        };
+
+        let existing_offsets = sheet.offsets.clear_widths();
+
+        let old_size = sheet.offsets.set_default_width(size);
+
+        if (cfg!(target_family = "wasm") || cfg!(test)) && !transaction.is_server() {
+            if let Some(sheet) = self.try_sheet(sheet_id) {
```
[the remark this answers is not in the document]
AyushAgrawal-A2: rust complains if I use the previous one
```
cannot borrow `self.a1_context` as immutable because it is also borrowed as mutable
immutable borrow occurs here
```

quadratic-core/src/controller/execution/execute_operation/execute_offsets.rs
AyushAgrawal-A2: we want this to fail silently, this is being done differently in different fns, needs a comprehensive refactor to make it uniform

quadratic-core/src/controller/execution/execute_operation/mod.rs
```diff
@@ -205,6 +205,11 @@ impl GridController {
                 Operation::MoveRows { .. } => self.execute_move_rows(transaction, op),
             }
         }
+        #[cfg(feature = "show-operations")]
+        if cfg!(target_family = "wasm") {
```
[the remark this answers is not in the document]
AyushAgrawal-A2: removed

PR #3080, opened 2025-06-13
fix: memory leak / execessive rendering

AyushAgrawal-A2:


quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellsLabels.ts
```diff
@@ -334,12 +335,8 @@ export class CellsLabels {
           notVisibleDirtyHashes.push({ hash, distance: this.hashDistanceSquared(hash, bounds) });
         }
       } else {
-        if (dirty) {
-          if (hash.clientLoaded) hash.unloadClient();
-          notVisibleDirtyHashes.push({ hash, distance: this.hashDistanceSquared(hash, bounds) });
-        } else if (hash.loaded) {
-          hash.unload();
-        }
```
AyushAgrawal-A2, line 342: this part fixes the excessive memory use due to excessive rendering

ddimaria: Do you still need the `if (hash.clientLoaded)` check?

ddimaria (approved):

quadratic-client/src/app/web-workers/renderWebWorker/worker/cellsLabel/CellsLabels.ts
AyushAgrawal-A2: the one at line 341 is needed, this one is removed

luke-quadratic (approved):
Built a big sheet and didn't run into issues on this front. 

Give a shout if there's specific things I should have tested that I missed by just doing general use.

PR #3084, opened 2025-06-17
chore: add workos auth provider

### LINKED ISSUE #3037: Explore auth providers and implement SSO

Our current auth providers (Auth0, Ory) are fantastic. We are exploring options for auth implementation and adding SSO, as some organizations are now requesting SSO. 

## Providers with good SSO options 

https://github.com/ory/polis
https://workos.com/docs/sso/what-you-will-build

AyushAgrawal-A2:
embed: https://quadratic-embed-test.vercel.app/

---

Migration Steps ([Dual Write Strategy](https://workos.com/docs/migrate/other-services)):
1. Ship with auth0 as the auth provider, with logic in server that any new user gets created in workos as well
2. Export and back fill all old users from auth0 to workos
3. Switch to workos (change env to use workos)

---
env changes (QA + Prod in pulumi):
- [ ] quadratic-client-*
   - [ ] change `VITE_AUTH_TYPE` to `workos` - step 3
   - [ ] add `VITE_WORKOS_CLIENT_ID` - step 1
- [ ] quadratic-api-*
   - [ ] change `AUTH_TYPE` to `workos` - step 3
   - [ ] add `AUTH_CORS` - step 1
   - [ ] add `WORKOS_CLIENT_ID` - step 1
   - [ ] add `WORKOS_API_KEY` - step 1
- [ ] quadratic-shared-*
   - [ ] change `JWKS_URI`  - step 3


quadratic-api/src/middleware/user.ts
```diff
@@ -94,12 +94,13 @@ export const userOptionalMiddleware = async (req: Request, res: Response, next:
...
 
   if (auth && auth.sub) {
-    const { user } = await getOrCreateUser(auth.sub);
+    const { user, userCreated } = await getOrCreateUser(auth.sub);
     if (!user) {
       return res.status(500).json({ error: { message: 'Unable to get authenticated user' } });
     }
 
     (req as RequestWithUser).user = user;
+    (req as RequestWithUser).userCreated = userCreated === true;
```
jimniels, line 103: Was this the source of the bug? Or is it just for completion's sake because `userMiddleware` does it too?

AFAIK this isn't actually possible—or maybe I need to update my mental model?

If `getOrCreateUser` is being guarded here by `auth && auth.sub` how is it possible that `auth.sub` has a value (the auth service knows who it is) but they're not in our db yet (we don't know who it is yet)?

A user will never get sent over to authenticate from a "user is optional" middleware endpoint, which means they'll never come back from the auth service from a "user is optional" middleware endpoint — if that makes sense?

I think ultimately the question I'm asking is: How is it possible that `userCreated` here ever returns true?

AyushAgrawal-A2: >If getOrCreateUser is being guarded here by auth && auth.sub how is it possible that auth.sub has a value (the auth service knows who it is) but they're not in our db yet (we don't know who it is yet)?

this is always the case for a new user, auth service know and our db does not, it gets added here, hence the function `getOrCreateUser`

.github/workflows/preview-build-deploy-test.yml
```diff
@@ -533,7 +533,7 @@ jobs:
       pull-requests: write
       issues: write
     runs-on: blacksmith-2vcpu-ubuntu-2404
-    if: github.head_ref != 'qa' || github.base_ref != 'main'
+    if: (github.head_ref != 'qa' || github.base_ref != 'main') && false
```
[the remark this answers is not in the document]
AyushAgrawal-A2: intentional, as all e2e will fail for new auth, to be removed before merging

quadratic-api/src/middleware/user.ts
jimniels: > this is always the case for a new user

for routes that use `userMiddleware` yes, but not for the single route in our app (`/file/:uuid`) that uses the `userOptionalMiddleware`.

If somebody visits `file/:uuid` (and therefore the `/files/:uuid` routes is called in the API) one of two things will happen:

1. They'll already have an account and be logged in, in which case the auth service will say "this is user x" and then we'll look them up in our db (and we'll know who they are)
2. They won't have an account or be logged in, in which case we will show them the file (if it's publicly shared) or give a 404.

AFAIK there is no way possible someone will hit `/file/:uuid` and be redirected to the auth service for not being logged in. That's why it's the user _optional_ middleware. And if they'll never be sent to authenticate, they'll never redirected back to this route (and middleware) as a new user (the case the auth service knows who they are but our db does not).

By definition, the user is optional in this middleware, which means the front-end will never send them to auth service for this route, which means they'll never come back as having had an account created — right?

That's my conceptual understanding of this code and that's the way it is the way it is. So I'm trying to understand if you discovered something here I'm not aware of.

infra/aws-cloudformation/quadratic-preview.yml
```diff
@@ -128,6 +128,11 @@ Resources:
           AWS_S3_ACCESS_KEY_ID=$(aws ssm get-parameter --name "/quadratic-development/AWS_S3_ACCESS_KEY_ID" --with-decryption --query "Parameter.Value" --output text)
           AWS_S3_SECRET_ACCESS_KEY=$(aws ssm get-parameter --name "/quadratic-development/AWS_S3_SECRET_ACCESS_KEY" --with-decryption --query "Parameter.Value" --output text)
 
+          # workos
+          WORKOS_CLIENT_ID=$(aws ssm get-parameter --name "/quadratic-development/WORKOS_CLIENT_ID" --with-decryption --query "Parameter.Value" --output text)
+          WORKOS_API_KEY=$(aws ssm get-parameter --name "/quadratic-development/WORKOS_API_KEY" --with-decryption --query "Parameter.Value" --output text)
+          JWKS_URI=$(aws ssm get-parameter --name "/quadratic-development/WORKOS_JWKS_URI" --with-decryption --query "Parameter.Value" --output text)
```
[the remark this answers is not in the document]
AyushAgrawal-A2: intentional

quadratic-api/src/middleware/user.ts
jimniels: > Code wise adding to db if auth service knows the user makes sense in both middleware

That's what I'm saying, "if auth service knows the user" should never happen here. That would be an unexpected state of the application and I don't think we should code for it.

A more correct representation of what we expect to happen, I believe, is represented by this PR where we 1) try to find the user (and if we can't, return an error), otherwise 2) set the user on the request object and `userCreated` as false, because now there's no code path that might theoretically create a user.

https://github.com/quadratichq/quadratic/pull/3155

jimniels: Ayush explained more to me. Closing this thread and PR. TY @AyushAgrawal-A2

quadratic-client/src/auth/workos.ts
```diff
@@ -0,0 +1,131 @@
...
+        const state = redirectParams.state;
+        if (!!state && typeof state === 'object' && 'redirectTo' in state) {
+          const redirectTo = state.redirectTo;
+          if (typeof redirectTo === 'string' && !!redirectTo) {
+            window.location.assign(redirectTo);
+            return;
+          }
+        }
+      },
+      devMode: true,
```
ddimaria, line 35: This intentional?

quadratic-client/src/auth/workos.ts
```diff
@@ -0,0 +1,131 @@
...
+    });
+  }
+  return clientPromise;
+}
+
+export const workosClient: AuthClient = {
+  /**
+   * Return whether the user is authenticated and the session is valid.
+   */
+  async isAuthenticated() {
```
ddimaria, line 45: Adding a return type here would be consistent.

quadratic-client/src/auth/workos.ts
```diff
@@ -0,0 +1,131 @@
...
+    const client = await getClient();
```
ddimaria, line 46: Is this the same `getClient` from above?  If so, it doesn't appear to be `async`.

ddimaria (approved):
Some nits/questions, but looks great 💯

jimniels:
If I signup for an account (no social login) the avatars don't seem to appear correctly? The img `src` is just missing entirely.

CleanShot 2025-07-01 at 12 49 12@2x

Whereas if I signup for an account (no social login) in QA (using auth0) I see the fallback is a generic img src provided by auth0.

CleanShot 2025-07-01 at 12 52 37@2x

I can't remember the details of how our avatar works, but it seems like we had a fallback? Should we expect the workos one to be displaying some kind of fallback instead?

jimniels:
I think something is broken with starting a new file with a prompt?

I did the following:

- Signed up as a new user
- Filled out the onboarding survey
- Got redirected to a new file at this url: https://ayush-workos.quadratic-preview.com/file/2ee177ba-b38a-4767-9e53-1d41028c0b8b?prompt=Hi%2C+I%27m+a+new+user+to+Quadratic+and+I+plan+on+using+it+for+work.+I+work+in+Product%3B+show+me+what+you+can+do+by+creating+me+a+sample+dataset+for+that+field.+Once+finished+with+the+dataset%2C+create+a+chart+that+helps+explain+the+data.+Then+tell+me+what+else+is+possible+in+Quadratic.

But the app never took the `prompt` search param, populated the AI, and fired off a request.

In fact, trying it with a new file doesn't seem to work either, see this URL:

https://ayush-workos.quadratic-preview.com/files/create?prompt=hi+how+are+you

That creates a new file, but it doesn't trigger the prompt.

AyushAgrawal-A2:
> I think something is broken with starting a new file with a prompt?
> 
> I did the following:
> 
> * Signed up as a new user
> * Filled out the onboarding survey
> * Got redirected to a new file at this url: https://ayush-workos.quadratic-preview.com/file/2ee177ba-b38a-4767-9e53-1d41028c0b8b?prompt=Hi%2C+I%27m+a+new+user+to+Quadratic+and+I+plan+on+using+it+for+work.+I+work+in+Product%3B+show+me+what+you+can+do+by+creating+me+a+sample+dataset+for+that+field.+Once+finished+with+the+dataset%2C+create+a+chart+that+helps+explain+the+data.+Then+tell+me+what+else+is+possible+in+Quadratic.
> 
> But the app never took the `prompt` search param, populated the AI, and fired off a request.
> 
> In fact, trying it with a new file doesn't seem to work either, see this URL:
> 
> https://ayush-workos.quadratic-preview.com/files/create?prompt=hi+how+are+you
> 
> That creates a new file, but it doesn't trigger the prompt.

fixed, was a bug in event from recoil atom, not auth related

quadratic-api/src/auth/router/authRouter.ts
```diff
@@ -0,0 +1,26 @@
...
+import sendResetPassword from '../routes/sendResetPassword.POST';
+import signupWithPassword from '../routes/signupWithPassword.POST';
+import authenticateRouter from '../routes/user_management.authenticate.POST';
+import logoutRouter from '../routes/user_management.sessions.logout.POST';
+import verifyEmail from '../routes/verifyEmail.POST';
+
+const authRouter = express.Router();
+
+authRouter.use('/user_management', authenticateRouter);
+authRouter.use('/user_management', logoutRouter);
```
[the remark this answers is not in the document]
AyushAgrawal-A2:
```
authRouter.use('/user_management', authenticateRouter);
authRouter.use('/user_management', logoutRouter);
```
these two are not configurable, workos hits these endpoint of the apihost, rest can be changed

AyushAgrawal-A2:
> Not sure if we can change this (or if it will be when we ship to prod and this is just a preview branch thing?)
> 
> But when you sign in with a social, it tells you something like "Are you sure you want to let this app access your info?" And it shows the workOS logo and says "workOS demo"
> 
> Do we need to configure something to make it say "Quadratic" or is that even possible?

can need production version of workos, this does not work in dev workos

docker/client/config/default.conf
```diff
@@ -31,9 +28,7 @@ server {
 
         add_header Vary Accept-Encoding;
 
-        add_header Cache-Control "no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0" always;
-        add_header Pragma "no-cache" always;
-        expires 0;
+        add_header Cache-Control "public, max-age=30" always;
```
ddimaria, line 31: @AyushAgrawal-A2 what's the impact of these cache changes?

quadratic-api/.env.example
```diff
@@ -3,6 +3,7 @@ ENVIRONMENT=local
 DATABASE_URL=postgresql://postgres:postgres@0.0.0.0:5432/postgres
 
 CORS='*'
+AUTH_CORS=http://localhost:3000
```
ddimaria, line 6: No `https`?

quadratic-api/src/auth/providers/workos.ts
```diff
@@ -0,0 +1,367 @@
...
+  email,
+  res,
+}: {
+  email: string;
+  res: Response;
+}): Promise<{ pendingAuthenticationToken: string | undefined }> => {
+  return handleEmailVerificationRequiredError(async () => {
+    clearCookiesWorkos({ res });
+
+    await getWorkos().userManagement.createMagicAuth({
```
ddimaria, line 291: try/catch?

quadratic-client/src/routes/reset-password.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <FormItem>
+                <FormLabel>New Password</FormLabel>
+                <FormControl>
+                  <Input
+                    data-testid="reset-password-password"
+                    className="h-12 rounded-md border border-gray-300 px-3 text-base"
+                    autoCapitalize="off"
+                    type="password"
+                    autoFocus
+                    placeholder="New Password*"
```
ddimaria, line 76: Why the `*` here?

quadratic-client/src/shared/components/auth/SignupForm.tsx
```diff
@@ -0,0 +1,193 @@
...
+                <FormLabel>{'Email'}</FormLabel>
+                <FormControl>
+                  <Input
+                    data-testid="login-email"
+                    className="h-12 rounded-md border border-gray-300 px-3 text-base"
+                    autoCapitalize="off"
+                    type="text"
+                    autoComplete="username"
+                    autoFocus
+                    placeholder="Work email*"
```
ddimaria, line 72: Is `Work` intentional here?

jimniels:
This is super-edge case-y and, from what I can tell, is how it works in prod today. So maybe it's fine to leave as-is, but to me it is unexpected.

I'm seeing an issue where:

- Login to quadratic as the same user across two tabs (e.g. user 1)
- In one of the tabs, sign out as user 1 and sign in as user 2
- Switch to the other tab and navigate the dashboard. Notice that you haven't been logged out in that tab. You're still logged in, but the wrong user is showing.

https://github.com/user-attachments/assets/54cabdc4-3de5-44ac-9221-90fbfde05032

If you log out in one tab, then switch to another tab and navigate, it will log you out. That's 👍 

But logging out in one tab, then logging in as another user, then switching tabs and navigating is a bit unexpected. I think I would expect that it switches users and teams — or maybe even does a hard-reload.

But, as I said, this seems to be an issue at present in prod, so maybe fine to leave as-is unless you see an easy way forward with it.

jimniels (changes requested):
I tested the following flows for the auth pages in chrome/safari:

- Login
- Social logins (Google, Github, Apple, Microsoft, all end up as the same quadratic account if it’s the same email)
- Sign up
- Password reset
- Magic code (send & enter)
- Login + various redirect codes (e.g. `/files/create?example=...` properly cloned an example file for me after login)

I also tested these actions inside the app as they touch auth provider stuff:

- Logout via button
- Logout via URL
- Add/remove user from team

Everything seemed to work as expected. Remaining items from my point of view:

- [x] Left some more comments in this PR review, all minor and related to user-facing UI copy
- [x] [Per this comment](https://github.com/quadratichq/quadratic/pull/3084/files#r2263936097), Kebab-casing is implemented in the client for auth routes but not on the server. Did you plan to change it there too? Also did you want to version those endpoints? (Not sure of the current status of that work)
- [x] I don't think we need [the placeholders with the `*` at the end](https://github.com/quadratichq/quadratic/pull/3084#discussion_r2267936407).
- [x] [This bug](https://github.com/quadratichq/quadratic/pull/3084#issuecomment-3169337538) of hitting `/logout` directly still exists

quadratic-client/src/shared/components/auth/SendMagicAuthCode.tsx
```diff
@@ -0,0 +1,84 @@
...
+        console.error(error);
+        setErrorMessage('Invalid email');
+      }
+    },
+    [navigate]
+  );
+
+  return (
+    <>
+      <h1 className="text-2xl font-medium">Send Magic Code</h1>
```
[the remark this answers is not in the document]
AyushAgrawal-A2: done

quadratic-client/src/routes/verify-email.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+        setErrorMessage('Invalid or expired code');
+      }
+    },
+    [navigate]
+  );
+
+  return (
+    <>
+      <h1 className="text-2xl font-medium">Verify Email</h1>
```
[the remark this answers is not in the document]
AyushAgrawal-A2: done

quadratic-client/src/routes/reset-password.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+        setErrorMessage('Invalid token or password');
+      }
+    },
+    [navigate]
+  );
+
+  return (
+    <>
+      <h1 className="text-2xl font-medium">Reset Password</h1>
```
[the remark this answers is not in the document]
AyushAgrawal-A2: done

quadratic-client/src/routes/reset-password.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+      <Form {...form}>
+        <form onSubmit={form.handleSubmit(handleSubmitForm)} className="flex w-full flex-col gap-6">
+          <FormField
+            control={form.control}
+            name="password"
+            render={({ field }) => (
+              <FormItem>
+                <FormLabel>New Password</FormLabel>
```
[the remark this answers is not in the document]
AyushAgrawal-A2: done

quadratic-client/src/routes/reset-password.tsx
AyushAgrawal-A2: done

quadratic-client/src/shared/components/auth/SignupForm.tsx
AyushAgrawal-A2: removed

quadratic-api/src/auth/providers/workos.ts
AyushAgrawal-A2: there is try catch in the wrapped `handleEmailVerificationRequiredError` fn

quadratic-client/src/auth/auth.ts
```diff
@@ -15,27 +13,105 @@ export interface User {
...
+let cachedAuthClient: Promise<AuthClient> | AuthClient | null = null;
+const getAuthClient = async (): Promise<AuthClient> => {
+  if (cachedAuthClient) {
+    return await cachedAuthClient;
+  }
+
   switch (AUTH_TYPE) {
     case 'auth0':
-      return auth0Client;
+      cachedAuthClient = (await import('@/auth/auth0')).auth0Client;
```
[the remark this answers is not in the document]
AyushAgrawal-A2: we will now load only 1 auth sdk in client, currently we load both auth0 and ory in prod and everywhere, though we don't use ory

AyushAgrawal-A2:
> This is super-edge case-y and, from what I can tell, is how it works in prod today. So maybe it's fine to leave as-is, but to me it is unexpected.
> 
> I'm seeing an issue where:
> 
> * Login to quadratic as the same user across two tabs (e.g. user 1)
> * In one of the tabs, sign out as user 1 and sign in as user 2
> * Switch to the other tab and navigate the dashboard. Notice that you haven't been logged out in that tab. You're still logged in, but the wrong user is showing.
> 
>  CleanShot.2025-08-11.at.16.58.07.mp4 
> If you log out in one tab, then switch to another tab and navigate, it will log you out. That's 👍
> 
> But logging out in one tab, then logging in as another user, then switching tabs and navigating is a bit unexpected. I think I would expect that it switches users and teams — or maybe even does a hard-reload.
> 
> But, as I said, this seems to be an issue at present in prod, so maybe fine to leave as-is unless you see an easy way forward with it.

done

quadratic-api/src/auth/routes/resetPassword.POST.ts
```diff
@@ -0,0 +1,36 @@
...
+      const {
+        body: { token, password },
+      } = parseRequest(req, schema);
+
+      await resetPassword({ token, password, res });
+
+      return res.status(200).json({ message: 'Password reset successful' });
+    } catch {
+      clearCookies({ res });
+
```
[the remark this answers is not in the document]
AyushAgrawal-A2: done

jimniels (approved):
LGTM 👍 

Last thought: maybe add "customize workOS permission screen" to the deployment checklist on the PR (if we haven't already, so [we don't get this in prod](https://github.com/quadratichq/quadratic/pull/3084#issuecomment-3169351096))

luke-quadratic:
From a UI standpoint, when I log in I'd expect the form to not be editable and some loading state (spinner?) indicating I'm being logged in. Currently it's unclear the login is happening and the form is still editable.

davidkircos (approved):

PR #3096, opened 2025-06-18
fix: example files with prompt in query param

AyushAgrawal-A2:


luke-quadratic (approved):
Works great, tested: 
- file creation and prompt 
- file creation on its own 
- prompt creation on its own 
- private param

ddimaria (approved):

PR #3101, opened 2025-06-19
feat: cross domain file transfer using iframe

AyushAgrawal-A2:
Sister PR: https://github.com/quadratichq/quadratic-website/pull/338

for user testing start from adding file in the marketing website


quadratic-client/src/app/ai/iframeAiChatFiles/FilesFromIframe.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    switch (event.data.type) {
+      case 'iframe-indexeddb-ready':
+        this.getFiles();
+        break;
+      case 'get-files-response':
+        this.handleGetFiles(event.data);
+        break;
+    }
```
ddimaria, line 42: Do we care about other message types?

quadratic-client/src/app/ai/iframeAiChatFiles/IframeAiChatFiles.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  private sendMessage = (message: FromIframeMessages, origin: string, transferables?: Transferable[]) => {
+    if (transferables) {
+      parent.postMessage(message, origin, transferables);
+    } else {
+      parent.postMessage(message, origin);
+    }
+  };
+
+  private validateState = (methodName: string): { db: IDBDatabase } => {
```
ddimaria, line 82: This could be renamed...it just seems like it's retrieving the db, but maybe I'm missing something regarding state validation?

quadratic-client/src/app/helpers/files.ts
```diff
@@ -27,6 +6,16 @@ export function getExtension(name: string): string {
   return name.split('.').pop() || '';
 }
 
+export function getFileTypeFromName(name: string): 'csv' | 'excel' | 'parquet' | 'grid' | undefined {
```
ddimaria, line 9: I feel like there's something like this elsewhere in the codebase.

ddimaria (approved):
This is great 💯.  I left some non-blocking comments/questions.

quadratic-client/src/app/helpers/files.ts
AyushAgrawal-A2: there is another which checks file mimetype as well, this one just check file extension from file name

quadratic-client/src/app/gridGL/pixiApp/urlParams/UrlParamsUser.ts
```diff
@@ -65,41 +73,164 @@ export class UrlParamsUser {
...
+    const firstSheet = sheets.getFirst();
+    if (!firstSheet) {
+      throw new Error('Expected to find firstSheet in urlParams.loadAIAnalystPrompt');
+    }
+
+    // push excel files to the end of the array
+    importFiles.sort((a, b) => {
+      const extensionA = getExtension(a.name);
+      const extensionB = getExtension(b.name);
+      if (['xls', 'xlsx'].includes(extensionA)) return 1;
```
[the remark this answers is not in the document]
AyushAgrawal-A2: we allow only these 2 from marketing currently and we only try / test these in app

quadratic-client/src/app/ai/iframeAiChatFiles/IframeAiChatFiles.ts
AyushAgrawal-A2: this is same as 2 more indexeddb in app, following the same name and pattern

quadratic-client/src/routes/files.create.tsx
```diff
@@ -43,6 +43,8 @@ export const loader = async (loaderArgs: LoaderFunctionArgs) => {
     prompt: url.searchParams.get('prompt'),
     // Creating via this route is _always_ private unless explicitly stated
     private: url.searchParams.get('private') === 'false' ? false : true,
+    // File to be fetched from iframe indexeddb for this chatid from marketing site
+    chatId: url.searchParams.get('chatid') || null,
```
jimniels, line 47: minor: the mismatch in capitalization feels like a place for future confusion (e.g. it's called `chatId` most places in the code, but is `chatid` when referring to the param). Not sure if you care about changing this to either `chatId` in the URL so it's the same everywhere in the code base, or to `chat-id` since that's generally what I've used in the past for named params in the url—see screenshot example of example named params in the url today in the app

CleanShot 2025-06-24 at 10 07 13@2x

jimniels (approved):
LGTM 👍

PR #3108, opened 2025-06-20
Thinking animation for chunky text 

luke-quadratic:
## Description 

Temporary animation that becomes not necessary when thinking isn't streaming huge chunks 

### Current 

CleanShot 2025-06-20 at 07 38 01

### New behavior 

CleanShot 2025-06-20 at 07 39 16


luke-quadratic:
no progress happening in this PR, closing

PR #3162, opened 2025-07-02
Fix bug with moving table and highlighting table

davidfig:
## Relevant issue(s)
https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1751402689509509

## Description
When moving tables, the table would sometimes lose highlight.


AyushAgrawal-A2 (approved):

jimniels (changes requested):
Dragging a table around leaves behind row striping where the table used to be.

https://github.com/user-attachments/assets/4bc4bbd4-9e18-429e-b126-f1a51b746d72

jimniels (approved):
Nice, works as expected!

PR #3163, opened 2025-07-02
fix: put return type error on a single line

jimniels:
## Description
How it looks today:
CleanShot 2025-07-02 at 14 03 25@2x

How it should look:

CleanShot 2025-07-02 at 14 05 08@2x


AyushAgrawal-A2:
image

PR #3180, opened 2025-07-08
render the column rename only after update has completed

### LINKED ISSUE #3099: Scrolled table hides column headers when clicking
## Description and reproduction

As shown in video, column names are disappearing when being double clicked into

ONLY whenever the table is scrolled does this happen. So of course to reproduce use a table long enough to get into a scroll state and then attempt to change column name 

https://github.com/user-attachments/assets/ef270b1c-f84f-42c4-9813-78cb3f25188f

davidfig:
## Relevant issue(s)
Fixes #3099

## Description
Properly positioned column rename when hovering on first use.

### DIFF
--- a/quadratic-client/src/app/events/events.ts
+++ b/quadratic-client/src/app/events/events.ts
@@ -129,6 +129,10 @@ interface EventTypes {
 
   // use this only if you need to immediately get the viewport's value (ie, from React)
   viewportChangedReady: () => void;
+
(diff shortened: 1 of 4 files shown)


luke-quadratic (approved):
Tested specifically only that this bug no longer exists. Did not test surrounding behaviors (not sure if there were any relevant surrounding behaviors to test).

PR #3182, opened 2025-07-08
Fixes placement of table headers when generating thumbnail

davidfig:
When viewport is scrolled, the floating table headers may be misplaced during thumbnail generation.

Old:
CleanShot 2025-07-08 at 07 03 17@2x

New:
CleanShot 2025-07-08 at 07 25 12@2x


luke-quadratic (approved):
Tried a scrolled table and thumbnail looks correct. Don't know how to test otherwise.

PR #3186, opened 2025-07-08
fix: model router for tool calls

AyushAgrawal-A2:
## Relevant issue(s)
fixes: free users use `DEFAULT_MODEL_FREE` for everything, ignoring tool call models


ddimaria (approved):

luke-quadratic (approved):
Resolved, checked flash is getting called for chat name 
CleanShot 2025-07-08 at 18 16 05@2x

PR #3188, opened 2025-07-09
fix: send date context after static context for caching

AyushAgrawal-A2:


luke-quadratic (approved):
Functionally works as expected

PR #3195, opened 2025-07-09
Formatting bar improvements

davidfig:
- [x] add parts of toolbar to sub menu when too large for screen
- [x] highlight buttons when turned on for current selection
- [x] make number buttons toggle (eg, percentage)


luke-quadratic (approved):
This is really awesome. 

I tested: 
- Each control in the toolbar works as expected. 
- Rapidly resized and didn't run into issues. 

(Tested nothing other than the above ^)

jimniels:
>  add parts of toolbar to sub menu when too large for screen

I'm not seeing this at all?

CleanShot 2025-07-11 at 11 41 44@2x

jimniels (approved):
## Works as expected

**Toggle functionality in the toolbar works for the following items:**
- Currency
- Percent
- Bold
- Italic
- Underline
- Strikethrough
- Fill color
- Text color
- Text alignment
- Vertical align
- Text wrap

Image

Image

**Toggle functionality works for those in various combinations**

Image

**Toggle functionality works across states that aren't all the same**

## Comments

Text align toggles for explicitly set values, but not implicitly set ones. For example, set a value to right align. You see it. Have a number set? It right-aligns by default as a formatted number but doesn't set the toggle to being right aligned (which I kinda expect it would?)

https://github.com/user-attachments/assets/5f1ca9ac-aa2c-4160-836b-798351cecbfd

Not a big deal. Probably related to underlying types so feel free to ignore.

quadratic-core/src/controller/user_actions/formats.rs
```diff
@@ -992,4 +1013,126 @@ mod test {
...
+        let sheet = gc.sheet(sheet_id);
+        assert_eq!(
+            sheet
+                .formats
+                .try_format(Pos { x: 1, y: 1 })
+                .unwrap_or_default()
+                .numeric_format,
+            Some(crate::grid::NumericFormat {
+                kind: crate::grid::NumericFormatKind::Number,
+                symbol: Some("€".to_string())
```
AyushAgrawal-A2, line 1049: symbol should be None on toggle

AyushAgrawal-A2 (approved):

PR #3205, opened 2025-07-11
fix: arrow keys in new sheet

### LINKED ISSUE #3203: bug: keyboard becomes unresponsive to arrow keys when creating a new sheet
## Describe the bug
When you create a new sheet, the keyboard becomes unresponsive to arrow keys and you can't move the cell cursor around on the new sheet. Going back to the old sheet it still works.

Refreshing doesn't fix the issue.

If you have multiple sheets, delete all of them except the last one, then refresh the page, the cursor works again (in the remaining sheet).

## Expected behavior
Cursor should work in any sheet.

## Reproduction
1. Create a blank file
2. Create a (second) new sheet in the file
3. Try using the arrow keys

https://github.com/user-attachments/assets/04c2ebb2-ebdb-4fc2-b7a8-1b93b29feaa0

AyushAgrawal-A2:
## Relevant issue(s)
closes #3203

### DIFF
--- a/quadratic-core/src/controller/send_render.rs
+++ b/quadratic-core/src/controller/send_render.rs
@@ -280,6 +280,7 @@ impl GridController {
         }
 
         sheet.send_content_cache();
+        sheet.send_data_tables_cache();
     }
 
     /// Sends delete sheet to the client


luke-quadratic (approved):
Fixes bug it claims to fix 

Only tested that it fixes that specific bug.

PR #3206, opened 2025-07-11
Fix broken table headers

davidfig:
## Relevant issue(s)
Fixes https://quadratichq.slack.com/archives/C03VC0Q1MPA/p1751995576001699

## Description
Nasty bug that wasn't properly updating the table headers for tables.


luke-quadratic (approved):
Tested only for this bug and can no longer recreate.

PR #3210, opened 2025-07-11
feat: did you know tooltip (for model picker)

jimniels:
## Description

Add a "Did you know?" dialog to the model picker that shows when we suss out that a user hasn't used the model picker at all.

Also updates the language of the models from "Basic/Pro" to "Fast/Max"

CleanShot 2025-08-08 at 11 11 44@2x
CleanShot 2025-08-08 at 11 11 50@2x

### Todos

- [x] Create a "Did you know?" component that can be used for showing informational tooltips to people.
- [x] Create a column on the user for storing JSON data we'll use to keep track of user-specific information, like "have they seen dialog X?"
- [x] Create a "Did you know?" component for the model picker that shows up 1) when they've never clicked the model picker, 2) when they've chatted with the AI at least 5 times _in a single chat_, and 3) that they never see it again after that
- [x] Write tests for changes to API endpoints

## To Test

There are two cases to test here. After doing one, you'll need to log out and create a new account to test the other.

### 1. Engaging with the model picker

- Create a new account
- Open a new file
- Chat with the AI 1-2 times.
- Click on the model picker
- Chat with the AI more, up to 5 times _in a single chat_

Through this entire process, you should never see the "Did you know?" popover.

- Open a new file
- Chat with the AI 5 times

Ensure, again, that you never see the popover.

### 2. Only using the AI

- Create a new account
- Open a new file
- Chat with the AI 5 times _in a single chat_
- See that on the 5th time, the "Did you know?" popover appears
- Open a new file
- Chat with the AI 5 times _in a single chat_ again
- See that the "Did you know?" popover DOES NOT appear

You should only ever see it pop up once.


luke-quadratic (approved):
👍

PR #3212, opened 2025-07-12
chore: add azure openai

AyushAgrawal-A2:


luke-quadratic (approved):
Seems to work. Just tested a few prompts from Azure provider. 

Fastest model responses I've seen from any host - impressive.

quadratic-api/src/ai/helpers/openai.helper.ts
```diff
@@ -340,3 +344,15 @@ export function parseOpenAIResponse(
 
   return { responseMessage, usage };
 }
+
+export function isOpenAIModel(modelKey: AIModelKey): modelKey is OpenAIModelKey {
```
ddimaria, line 348: This code may be a duplicate from `model.helper.ts`?

AyushAgrawal-A2: removed

PR #3213, opened 2025-07-14
chore: add AIAnalystDebugChatInput

AyushAgrawal-A2:
## Relevant issue(s)
adds chat chat string input box to view / export a Analyst Chat

### DIFF
--- a/quadratic-api/src/ai/helpers/bedrock.helper.ts
+++ b/quadratic-api/src/ai/helpers/bedrock.helper.ts
@@ -33,6 +33,7 @@ import type {
   ParsedAIResponse,
   ToolResultContent,
 } from 'quadratic-shared/typesAndSchemasAI';
+import { v4 } from 'uuid';
 
 function convertContent(content: Content): ContentBlock[] {
   return content.map((content) => {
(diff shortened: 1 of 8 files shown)


quadratic-shared/ai/helpers/message.helper.ts
```diff
@@ -201,22 +203,33 @@ export const replaceOldGetToolCallResults = (messages: ChatMessage[]): ChatMessa
...
-                {
-                  type: 'text' as const,
-                  text: CLEAN_UP_MESSAGE,
-                },
-              ],
+              id: toolResult.id,
+              content:
+                toolResult.content.length > 1
+                  ? toolResult.content.filter((content) => !isContentImage(content))
+                  : toolResult.content,
```
ddimaria, line 231: Will this never be an image?

AyushAgrawal-A2: should not be, there is only one case where we add an image and we add another text content
also we cannot make content array empty, always need to leave 1

ddimaria: Could this ever be a zero sized array (all images)?
```ts
toolResult.content.filter((content) => !isContentImage(content))
```

PR #3219, opened 2025-07-16
fix: ssh connection host

AyushAgrawal-A2:
fixes ssh connection


quadratic-rust-shared/src/sql/mod.rs
```diff
@@ -204,7 +204,13 @@ pub trait UsesSsh {
     // Set the port to connect to
     fn set_port(&mut self, port: u16);
 
-    // The host to connect to
+    // The database host to connect to
+    fn host(&self) -> String;
```
ddimaria, line 208: Since all implementations are the same, you should have a default implementation here and remove the implementations in the mysql/mssq/postgres connection files.

ddimaria (approved):

PR #3226, opened 2025-07-16
fix: make sheet_name optional in ai tools

AyushAgrawal-A2:
## Relevant issue(s)
fixes ai code cell errors, when sheet_name is missing in ai args, defaults to current sheet

### DIFF
--- a/quadratic-client/src/app/ai/tools/aiToolsActions.ts
+++ b/quadratic-client/src/app/ai/tools/aiToolsActions.ts
@@ -220,7 +220,7 @@ export const aiToolsActions: AIToolActionsRecord = {
   [AITool.SetCellValues]: async (args) => {
     const { sheet_name, top_left_position, cell_values } = args;
     try {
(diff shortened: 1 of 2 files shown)


quadratic-client/src/app/ai/toolCards/SetCodeCellValue.tsx
```diff
@@ -33,7 +33,10 @@ export const SetCodeCellValue = memo(({ args, loading }: SetCodeCellValueProps)
 
         if (toolArgs.success) {
           try {
-            const selection = sheets.stringToSelection(toolArgs.data.code_cell_position, sheets.current);
+            const sheetId = toolArgs.data.sheet_name
+              ? (sheets.getSheetByName(toolArgs.data.sheet_name)?.id ?? sheets.current)
```
davidfig, line 37: I thought we were saving the sheet at the time of the query. Shouldn't we use that?

davidfig (approved):

PR #3235, opened 2025-07-17
fix: AI convert to table misreports actual state

### LINKED ISSUE #3232: AI convert to table misreports actual state
## Describe the bug

AI convert to table misreports actual state. 

This creates a lot of loops where AI has to figure out that what we told it was wrong was actually correct. 

## Reproduction

- Create some flattened data 
- Convert it to table 
- Note that the response after converting to table is that the action errored 

https://github.com/user-attachments/assets/fd4ada71-b959-4b7e-9f52-9646d5af69bb

AyushAgrawal-A2:
## Relevant issue(s)
closes #3232


luke-quadratic (approved):
Fixes error, ty!

PR #3248, opened 2025-07-18
Fix core error after getRenderCells

davidfig:
Fixed an expected error crashing the app.

### DIFF
--- a/quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
+++ b/quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
@@ -147,22 +147,16 @@ class Core {
     y: number;
     width: number;
     height: number;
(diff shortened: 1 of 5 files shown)


quadratic-client/src/app/web-workers/quadraticCore/worker/core.ts
```diff
@@ -157,7 +157,9 @@ class Core {
         );
         resolve(renderCells);
       } catch (e) {
-        this.handleCoreError('getRenderCells', e);
+        if (e instanceof Error && !e.message.includes('Sheet not found')) {
```
ddimaria, line 160: Relying on that text is a bit brittle.

AyushAgrawal-A2 (approved):

ddimaria (approved):

PR #3258, opened 2025-07-21
fix: deleted columns immediately adjacent to table disables functionality

AyushAgrawal-A2:
## Relevant issue(s)
closes #3204


luke-quadratic (approved):
works as expected 👍

PR #3263, opened 2025-07-21
fix: move table from top most edge copies table contents instead of moving

AyushAgrawal-A2:
## Relevant issue(s)
closes #3164


luke-quadratic (approved):
Fix works as expected 👍

PR #3266, opened 2025-07-22
fix: mixpanel core errors

AyushAgrawal-A2:


quadratic-client/src/app/ai/tools/aiToolsActions.ts
```diff
@@ -518,13 +532,22 @@ export const aiToolsActions: AIToolActionsRecord = {
...
-      await quadraticCore.setFormats(sheetId, args.selection, formatUpdates);
-      return [
-        {
-          type: 'text',
-          text: `Executed set formats tool on ${args.selection} for ${describeFormatUpdates(formatUpdates, args)} successfully.`,
-        },
-      ];
+      const response = await quadraticCore.setFormats(sheetId, args.selection, formatUpdates);
+      if (response?.result) {
+        return [
```
ddimaria, line 537: I like the explicitness of returning the array, but there's a lot of boilerplate for these text types.  Maybe a helper function would tidy things up?
```ts
const textResponse(text: string) {
  ({ type: 'text', text })
}
```

quadratic-client/src/app/ui/components/FixSpillError.tsx
```diff
@@ -35,6 +35,7 @@ export const FixSpillError = ({ codeCell, evaluationResult, onClick }: FixSpillE
           reverse: false,
         })
         .then((pos) => {
+          if (!pos) return;
```
ddimaria, line 38: I wonder why pos would not be defined here?

quadratic-client/src/app/ai/tools/aiToolsActions.ts
AyushAgrawal-A2: refactoring this in another PR

Issue #3281, opened 2025-07-22
BigQuery Dataset hard-coded for Query Selected Table functionality


ddimaria:
## Describe the bug
When selecting the `Query Selected Table` button in the UI, the test Dataset was hard-coded rather than being pulled from the connection config.

## Expected behavior
Selecting the `Query Selected Table` button should show the correct Dataset in the query.

## Reproduction
1. Create BigQuery connection
2. With a cell highlighted, open connections and select BigQuery connection
3. For any table, click on the `Query Selected Table` button
4. `all_native_data_types` will appear as the dataset rather than the correct dataset

PR #3282, opened 2025-07-22
Fix hard-coded biqquery dataset in Query Selected Table

ddimaria:
## Relevant issue(s)
Fixes https://github.com/quadratichq/quadratic/issues/3281

## Description
When selecting the Query Selected Table button in the UI, the test Dataset was hard-coded rather than being pulled from the connection config. With this PR, selecting the Query Selected Table button should show the correct Dataset in the query.

## (Optional) Testing considerations
1. Create BigQuery connection
1. With a cell highlighted, open connections and select BigQuery connection
1. For any table, click on the Query Selected Table button
1. The correct dataset should display in the window


luke-quadratic (approved):
Fix works

PR #3290, opened 2025-07-23
Ai insert and delete columns and rows

davidfig:
- [x] insert columns
- [x] insert rows
- [x] delete columns
- [x] delete rows


luke-quadratic (approved):
Approving purely on functionality, will review prompts etc deeper when all AI PRs are together.

AyushAgrawal-A2 (approved):

PR #3296, opened 2025-07-24
fix: when deleting many table rows, header UI persists at bottom of table

AyushAgrawal-A2:
## Relevant issue(s)
closes #3291


luke-quadratic (approved):
Tested: deleted a large amount of rows and was unable to recreate

PR #3308, opened 2025-07-26
AI prefer table names

davidfig:
Add prompting to prefer table name A1 notation instead of range selection. This makes q.cells and makes formatting more robust when the table changes sizes.


luke-quadratic (changes requested):
Few issues with this: 
- Worried about saying "Preferred" for table references without giving the clear "Preferred" counter for when it's not tables. Can easily see AI hallucinating here and going oh tables are preferred I should always use tables. 
- You describe the usage of #DATA. This doesn't work in Python. This instruction will likely cause Python generation to try to use this instruction sometimes and fail.

PR #3309, opened 2025-07-27
AI Tool call for validations

davidfig:
- [x] all ai validation tools
- [x] getValidation
- [x] removeValidation
- [x] add e2e tests for all validations, including infinite ranges
- [x] fix bug with dropdown icons not properly positioned for infinite ranges
- [x] fix bug with cell ribbons not moving when resizing cell


AyushAgrawal-A2 (approved):

luke-quadratic (changes requested):
Tool calls always getting added with errors/invalid since there's no entry choice 
CleanShot 2025-08-04 at 14 35 01@2x

luke-quadratic (approved):
👍

PR #3327, opened 2025-08-03
fix: use dexie for IndexedDb

AyushAgrawal-A2:
## Relevant issue(s)
fixes app crashing due to IndexedDb error (received in Mixpanel)

image


davidfig (approved):
Did we test to ensure offline works from old code -> new code?

AyushAgrawal-A2:
> Did we test to ensure offline works from old code -> new code?

yup, works

PR #3334, opened 2025-08-05
fix: mixpanel identity tracking

jimniels:
## Description
Implement the simplified identity management for mixpanel.

Docs: https://docs.mixpanel.com/docs/tracking-methods/id-management/identifying-users-simplified

### Code changes

- [`analytics.ts`](https://github.com/quadratichq/quadratic/pull/3334/files#diff-ecdf3e02b933e63af3ced1f53e6a517deea21573f2f326ef3d5c309caeb33408) - Moved current analytics event initialization code out of this file to its own file (see `analyticsEvents.ts`)
- [`analyticsEvents.ts`](https://github.com/quadratichq/quadratic/pull/3334/files#diff-7da1aad3d0f1251e7eed2299d1b0860767c9f4793a546bd30445271880cc4288) - This is where all centralized logic for event tracking lives. We centralize all integration with event analytics SDKs here so we can use multiple ones if we want (analytics sdks are initialized when the module is loaded).
- `**/*.ts|tsx` - Change any files with `mixpanel.track()` to use the new `trackEvent()` function

### Mixpanel integration changes

The following changes were made to how we work with the mixpanel SDK based on [their recommendations for simplified ID management](https://docs.mixpanel.com/docs/tracking-methods/id-management/identifying-users-simplified)

- [x] "Call `identify(user_id)` when a user signs up or logs in"
  - I think we want to keep the current call to `.identify` where it is, because [their best practices](https://docs.mixpanel.com/docs/tracking-methods/id-management/identifying-users-simplified#best-practices) state (emphasis mine): "Call `.identify` upon a registration/login **or when an app is re-opened in a logged-in state**". 
  - If a user closes a tab to Quadratic, then re-opens a tab and they're logged in (has token in local storage), they won't go through the `login-result` page and therefore mixpanel would never be initialized. `.identify` needs to happen with the initialization of the app.
- [x] "Send at least one event after the `.identify()` call"
  - We're sending `[mixpanel].identify` as just a dummy event
- [x] "Call `.reset()` when a user logs out
  - Happens in [`logout.tsx`](https://github.com/quadratichq/quadratic/pull/3334/files#diff-5002edbc54bd3e41a8c5c5f2d59adca3e3db729a8c5613632ef687c0056a9fb4)
- [x] Send `[Auth].signup` event when a new user is created (from the client)
  - Happens in [`login-result.tsx`](https://github.com/quadratichq/quadratic/pull/3334/files#diff-3e7509f0c12c14e79ef9d0a842c58779162e82ecf89abaf8e8546f89c7490258)

On Deployment
- [ ] Remove the `[Auth0].signUp` event from the auth0 server setup (@davidkircos)
- [ ] Setup new project in mixpanel and change `VITE_MIXPANEL_ANALYTICS_KEY` key (@davidkircos)
- [ ] Check mixpanel on the marketing site. Update project ID on marketing site @davidkircos

### DIFF
--- a/quadratic-client/src/routes/logout.tsx
+++ b/quadratic-client/src/routes/logout.tsx
@@ -1,4 +1,5 @@
 import { authClient } from '@/auth/auth';
+import mixpanel from 'mixpanel-browser';
 import { redirect } from 'react-router';
 
 // If you visit `/logout` directly in your browser, we'll log you out.
@@ -11,5 +12,6 @@ export const action = logout;
 async function logout() {
   localStorage.clear();
   await authClient.logout();
+  mixpanel.reset();
   return redirect('/');
 }
--- a/quadratic-client/src/routes/logout.tsx
+++ b/quadratic-client/src/routes/logout.tsx
@@ -1,5 +1,5 @@
 import { authClient } from '@/auth/auth';
-import mixpanel from 'mixpanel-browser';
+import { resetEventAnalytics } from '@/shared/utils/analyticsEvents';
 import { redirect } from 'react-router';
 
 // If you visit `/logout` directly in your browser, we'll log you out.
@@ -12,6 +12,6 @@ export const action = logout;
 async function logout() {
   localStorage.clear();
   await authClient.logout();
-  mixpanel.reset();
+  resetEventAnalytics();
   return redirect('/');
 }
--- a/quadratic-client/src/shared/utils/analytics.ts
+++ b/quadratic-client/src/shared/utils/analytics.ts
@@ -1,8 +1,8 @@
 import { debugFlag } from '@/app/debugFlags/debugFlags';
 import type { User as AuthUser } from '@/auth/auth';
+import { identifyEventAnalyticsUser } from '@/shared/utils/analyticsEvents';
 import * as amplitude from '@amplitude/analytics-browser';
 import { setUser } from '@sentry/react';
-import mixpanel from 'mixpanel-browser';
 
 // Quadratic only shares analytics on the QuadraticHQ.com hosted version where the environment variables are set.
 
@@ -12,7 +12,7 @@ export function googleAnalyticsAvailable(): boolean {
   return import.meta.env.VITE_GOOGLE_ANALYTICS_GTAG && import.meta.env.VITE_GOOGLE_ANALYTICS_GTAG !== 'none';
 }
 
-function getUtmDataFromCookie(): {
+export function getUtmDataFromCookie(): {
   utm_source: string | undefined;
   utm_medium: string | undefined;
   utm_campaign: string | undefined;
@@ -40,7 +40,7 @@ function getUtmDataFromCookie(): {
 export function initializeAnalytics(user: User) {
   loadGoogleAnalytics(user);
   initAmplitudeAnalytics(user);
-  initMixpanelAnalytics(user);
+  if (user) identifyEventAnalyticsUser(user);
   configureSentry(user);
 }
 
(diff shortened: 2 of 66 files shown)


quadratic-client/src/shared/utils/analytics.ts
```diff
@@ -119,21 +119,31 @@ export function initMixpanelAnalytics(user: User) {
...
+  // https://docs.mixpanel.com/docs/tracking-methods/id-management/identifying-users-simplified#best-practices
+  if (!user?.sub) return;
+
+  // Identify the user
+  mixpanel.identify(user.sub);
+
+  // Globally register stuff we want to send with every event
   mixpanel.register({
-    email: user?.email,
-    distinct_id: user?.sub,
```
jimniels, line 124: We're removing `distinct_id` here (and on line 131 below) because this should be taken care of by Mixpanel's system and sdk.

quadratic-client/src/shared/utils/analytics.ts
```diff
@@ -12,7 +12,7 @@ export function googleAnalyticsAvailable(): boolean {
   return import.meta.env.VITE_GOOGLE_ANALYTICS_GTAG && import.meta.env.VITE_GOOGLE_ANALYTICS_GTAG !== 'none';
 }
 
-function getUtmDataFromCookie(): {
+export function getUtmDataFromCookie(): {
```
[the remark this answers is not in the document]
AyushAgrawal-A2: also add a try catch to `JSON.parse` in this fn

quadratic-client/src/routes/logout.tsx
```diff
@@ -11,5 +12,6 @@ export const action = logout;
 async function logout() {
   localStorage.clear();
   await authClient.logout();
+  mixpanel.reset();
```
AyushAgrawal-A2, line 15: check for `isMixpanelEnabled` here ?

AyushAgrawal-A2: or this also can be an export from `analyticsEvents.ts` where `isMixpanelEnabled` checked internally

AyushAgrawal-A2 (changes requested):

quadratic-client/src/shared/utils/analyticsEvents.ts
```diff
@@ -0,0 +1,98 @@
...
+// https://docs.mixpanel.com/docs/tracking-methods/id-management/identifying-users-simplified#best-practices
+function startMixpanel(user: User) {
+  const utmData = getUtmDataFromCookie();
+
+  // Identify the user
+  mixpanel.identify(user.sub);
+
+  // Globally register stuff we want to send with every event
+  mixpanel.register({
+    email: user.email,
```
davidkircos, line 81: I don't think we need to send the user's email here. We put it in the profile below. 

Question: can we register their plan status here? Do you have that information
`{'plan':'free'}` `{'plan':'pro'}`

davidkircos: If we register it here, Mixpanel will send it with all other events in the session as well

quadratic-client/src/routes/logout.tsx
davidkircos: Wrap in a function in analytics.ts so we are ideally only importing `mixpanel` from that one file.

jimniels: ah yes, good catch. will export a fn from our centralized analytics file and use that here

jimniels: [Fixed in this commit](https://github.com/quadratichq/quadratic/pull/3334/commits/c8c023974b9847c3581f9e8c91cf632852042210)

quadratic-client/src/shared/utils/analytics.ts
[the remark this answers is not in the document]
jimniels: [Done](https://github.com/quadratichq/quadratic/pull/3334/commits/38359f192e5e53b03eb0674f24313648c8ec1ede). Moved into new file and wrapped JSON.parse in try/catch

AyushAgrawal-A2 (approved):

davidkircos (approved):

PR #3340, opened 2025-08-07
feat: potential connections

jimniels:
## Description

- [x] Add unsupported connections to the list of functioning connections.
- [x] When a user clicks on an unsupported connection, send a mixpanel event denoting they clicked on it
- [x] Allow users to submit additional feedback about an unsupported connection. This message gets logged as feedback using our existing feedback mechanism (logged in db + sent to slack)

CleanShot 2025-08-13 at 12 50 17@2x

CleanShot 2025-08-13 at 12 50 24@2x

## To Test

- Go to connections and see that there are new connections in the list
- Click on any and submit feedback, see that it submits
- Message will only get logged to slack if we setup the `SLACK_FEEDBACK_URL` in the preview env

Goes with https://github.com/quadratichq/quadratic-website/pull/355/files


luke-quadratic (approved):
👍

AyushAgrawal-A2 (approved):

PR #3353, opened 2025-08-09
Alert users in members page that adding members while on Pro costs per seat 

luke-quadratic:
Alert users in members page that adding members while on Pro costs per seat. 

CleanShot 2025-08-08 at 21 28 01@2x

## Testing considerations 

Notification should not show for free teams. Should show for paid teams.


jimniels (approved):
I moved this to the side as a kind of addendum to the UI itself.

Image

PR #3367, opened 2025-08-14
fix: minor ai chat bugs

AyushAgrawal-A2:
1. make backup model non sticky
2. unique message insertion fail due to debug chat id not updating chat_id
3. handle bad tool calls from dropped ai responses


luke-quadratic (approved):
👍

Issue #3307, opened 2025-07-25
Unsubscribe flow


luke-quadratic:
## Description 

Upon clicking unsubscribe, pop a modal to collect voluntary information on why user is leaving to help improve the product.

jimniels:
Resolved in #3381

PR #3381, opened 2025-08-22
feat: try to retain users who are unsubscribing

jimniels:
## Relevant issue(s)
 #3307 

## Description

tk

Todos:

- [ ] Write API tests

## To test

- [ ] tk


davidkircos (approved):
UI UX looks great to me! 

When redirecting the user to stripe, I think we should set the success / return URLs to the team settings page 

Image

AyushAgrawal-A2 (approved):

PR #3384, opened 2025-08-25
Simplifies ScrollBars so it can be used in other views

davidfig:
This should have the same behavior as qa.


quadratic-client/src/app/gridGL/HTMLGrid/scrollBars/ScrollBars.tsx
```diff
@@ -52,30 +90,37 @@ export const ScrollBars = () => {
...
-    [scrollBarsHandler]
+    [horizontalStart]
   );
 
   const pointerMoveHorizontal = useCallback(
     (e: PointerEvent) => {
       if (state === 'horizontal' && down !== undefined && start !== undefined) {
         const delta = e.clientX - down.x;
-        const actualDelta = scrollBarsHandler.adjustHorizontal(delta);
+        let actualDelta: number;
```
ddimaria, line 103: Less branching approach:
```ts
let actualDelta = delta;

if (delta !== 0) {
  ...
}
```

PR #3385, opened 2025-08-25
fix: backup model in config

AyushAgrawal-A2:


luke-quadratic (approved):
good as long as this is ok 

https://github.com/quadratichq/quadratic/pull/3385/files#r2298610315

PR #3389, opened 2025-08-26
fix: tool card text length

AyushAgrawal-A2:
ref: https://github.com/quadratichq/quadratic/pull/3375


davidfig (approved):

luke-quadratic (approved):
ran through most to test, text looks nice

PR #3419, opened 2025-09-06
fix(workos): invites and connection schema

AyushAgrawal-A2:
fixes: 
1. https://quadratichq.slack.com/archives/C06Q34HGV8A/p1757105331707949
2. https://quadratichq.slack.com/archives/C06Q34HGV8A/p1757108711002269


quadratic-connection/src/auth.rs
```diff
@@ -23,7 +23,7 @@ use crate::{
 /// We need our own implementation of this because we need to impl on it.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Claims {
-    pub sub: String,
+    pub email: String,
```
ddimaria, line 26: Will this cause a problem with existing JWTs?

quadratic-rust-shared/src/auth/jwt.rs
```diff
@@ -96,12 +96,14 @@ pub mod tests {
         let result = authorize::<Claims>(&jwks, TOKEN, false, false);
         assert!(result.is_ok());
 
-        // invalid audience causes error
-        let result = authorize::<Claims>(&jwks, TOKEN, true, false);
-        assert_eq!(
-            result.unwrap_err(),
-            SharedError::Auth(Auth::Jwt("InvalidAudience".into()))
-        );
+        // no audience in workos claims
```
ddimaria, line 99: Delete this code?

ddimaria (approved):
Looks good!

quadratic-connection/src/auth.rs
AyushAgrawal-A2: nope, all existing jwts have emails as well, we shipped that change weeks ago, auth0 jwt expiry is 24 hrs, so all valid jwts will have emails also

PR #3443, opened 2025-09-15
[WIP] feat: narrow AI context with sheet data & remove right chat

jimniels:
## Update

Closed this in favor of #3473 

That said, there's still some useful code in here we may resurrect around right hand code chat patterns for deprecation:
CleanShot 2025-09-24 at 14 33 09@2x

As well as GoTo dropdown filters

CleanShot 2025-09-24 at 14 32 59@2x

## Description

> [!IMPORTANT]  
> This is a prototype and will be broken in places. It's purely meant to give a feel for a certain direction.

Prototype that allows people to narrow the context of their conversation in the left chat by specifying data in the sheet as context. A few features:

1. Allow users to narrow context from the AI prompt
2. Allow users to right-click things on the sheet and narrow context from a dropdown item
3. Remove right chat and replace with pointers to new chat

When the code editor is open, behind-the-scenes, we should instruct the AI that that particular code cell is open and visible to the user and so it should focus on that.

### Outstanding questions

- What does it mean to actually select something (table, code, etc.) and how does that change instructions to the AI?
    - How do we handle all the different kinds of context menus and what the "chat" does in response to those?
        - Cells with spills
        - Cells inside of tables or code tabls, etc.
        - Do we want these interactions to pre-fill the prompt with text? Or do we want some kind of pill-shaped context attached to the chat?
- Do we want to be more specific than just the range?
    - "I want to focus on cells `A1:D100`"
    - "I want to focus on the table `Table1`"
    - "I want to focus on the code table `Python1`"
    - "I want to focus on the chart `Javascript1`"

### Todos

- [ ] UI for making a "selection" as context in the prompt
- [ ] UI to right-click, "add to context" from sheet
- [ ] Add control to "add to context" / "chat with AI" from the code editor
- [ ] Remove "AI assistant" from code editor

## To test
tk

## QA Wolf tests 

Any?


jimniels:
closing in favor of #3473

PR #3481, opened 2025-09-27
Fix various bugs in Find

davidfig:
- [x] Fix bug where search would change sheets improperly (from Cole: for example if you do a search on one tab, then close it and do another search on another tab, it’ll take you back to the first tab) + e2e
- [x] going backwards through the search results doesn't always move the viewport + e2e
- [x] search should refresh after any transaction (but keep its current cursor location unless the results change) + e2e
- [x] fix bug that prevented the fill color from showing up on search results
- [x] fix screenshots for ^ in e2e tests


quadratic-client/src/app/ui/components/Search.tsx
```diff
@@ -76,14 +79,48 @@ export function Search() {
     [searchOptions]
   );
 
+  useEffect(() => {
+    const updateSearch = async () => {
+      if (!showSearch || !inputEl?.value) return;
+      const found = await quadraticCore.search(inputEl?.value, searchOptions);
+      if (found) {
+        setResults(found);
+        const currentLocal = current > found.length - 1 ? 0 : current;
```
ddimaria, line 88: What does `currentLocal` mean?

ddimaria (approved):

